Issue #3, opened 2017-05-16
Broken links - references to parts of OAS specification?


MikeRalphson:
Found by [mdv](https://github.com/mermade/mdv), which runs as part of the OAS CI task.

````js
{ source: 'README.md',
  missingAnchors:
   [ { name: 'componentsObject', defined: 0, seen: 1 },
     { name: 'dataTypeFormat', defined: 0, seen: 1 },
     { name: 'referenceObject', defined: 0, seen: 1 } ] }
````

Happy to raise a PR to add a CI step. I think all of the above need sections adding to the spec or links created to a version of the OAS spec?

fmvilas:
Awesome, didn't know about mdv. A PR would be much appreciated.

MikeRalphson:
`componentsObject` looks like it should have its own definition in the AsyncAPI spec as it has a different structure to the OAS object. Should the other two be links into OAS? If so, to which version/branch/tag?

fmvilas:
IMO everything should be in the AsyncAPI document and not referencing OAS externally beacause the two specifications can evolve in a different way/pace. So I would just copy the missing pieces and paste them here. And in case it's different (like components object) we should redact its definition.

What do you think?

fmvilas:
@MikeRalphson it's working perfectly: https://travis-ci.org/asyncapi/asyncapi

Thanks man 👏

PR #6, opened 2017-05-17
Add Travis CI with mdv step

MikeRalphson:
References #3 - the build [fails currently](https://travis-ci.org/MikeRalphson/asyncapi/builds/233126839) due to these broken links.


fmvilas:
Ok so now that links are fixed let's see if this doesn't fail. I'm gonna setup Travis today. Thanks @MikeRalphson.

PR #14, opened 2017-05-23
Typo: should be 'baseTopic' instead of ' host'

ignasi35:


fmvilas:
Hey @ignasi35! Thanks for spotting that and taking the time to fix it. LGTM, merging.

PR #17, opened 2017-05-23
Use a Github friendly Apache 2 License

ignasi35:
It has the same legal effects but GH will recognise it and use more user friendly UI for the repo.


fmvilas:
Sweet, didn't know that. Thanks very much!

PR #20, opened 2017-05-28
Small typo on section 'schemes'

repejota:
The YAML example was 'host'  when it should be called 'schemes'.


fmvilas:
Eagle eye! Thanks mate!

PR #27, opened 2017-07-24
schema; allow $ref in message object

MikeRalphson:
Allows samples to be validated


fmvilas:
Thank you very much for spotting it mate.

PR #29, opened 2017-07-28
Add servers object and remove host and schemes properties

fmvilas:
This removes the host and schemes properties. For a better way to declare servers, protocols, etc. we followed the same approach as OpenAPI 3.0.0, using a `servers` object in the top level of the document. As discussed in https://github.com/asyncapi/asyncapi/issues/28.

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,14 +307,14 @@ A single server would be described as:
 
 ```json
 {
-  "url": "//development.gigantic-server.com",
+  "url": "development.gigantic-server.com",
   "description": "Development server",
   "scheme": "mqtts"
 }
 ```
 
 ```yaml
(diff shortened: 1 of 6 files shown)


fmvilas:
@MikeRalphson would you mind to review this?

MikeRalphson:
@fmvilas All looks good to me. The one question I have is over the leading `//` in the server urls. Is that a standard way of referring to the endpoint of an async API, or should they be dropped? Showing my ignorance of the subject here!

fmvilas:
Well actually is not needed. It's fine to drop it or to use, i.e. `mqtt://api.company.com`. I was actually following your comment on the issue hehe. I'll drop it to avoid misunderstandings. Thanks man.

PR #33, opened 2017-08-01
Eliminate red-highlight of [...] in json example

MikeRalphson:
Uses the fact that YAML is a superset of JSON, but treats ... as a string literal.


fmvilas:
Thanks!

PR #36, opened 2017-08-21
Mark URL field as format 'uri'

ctron:
To me it looks like as is if the `termsOfService` field is missing the URI format.


fmvilas:
Agree. Good catch! Thanks for the PR!

ctron:
Thanks for the quick merge!

PR #37, opened 2017-09-15
Add security schemes and requirements

fmvilas:
It adds support for:

1. Defining security schemes inside the Components object.
2. Specify security requirements to authorize the connection.

#### Security schemes

As a result of an investigation about ways to protect your API using MQTT, AMQP and WebSockets, I came up with the following list:

1. User/Password: That's the most common when the API is private.
2. API Key (over User/Password): It feels like a hack to me but I've seen it in many places. It consists in sending an API token as the user or as the password when connecting to a system that expects a User/Password. Usually, the other field is left empty. For instance:
```yaml
user: AF89AFDKLK23423NADSFDS89AS79
password: (empty)
```
3. X.509 Certificate: That's very common for internal and/or partnership APIs.
4. End-to-end Encryption: Either symmetric or asymmetric. **I'm not sure this should be an auth scheme but another way to protect your API. Comments welcome :blush:**
5. HTTP: It's especially useful for WebSockets although suitable for other types of APIs. That's exactly the same as Open API's `http` scheme. It can be bearer or non-bearer (i.e. basic).
6. HTTP API Token: It's especially useful for WebSockets although suitable for other types of APIs. That's exactly the same as Open API's `apiKey` scheme. You can specify where the token is placed: header, query param or cookie.

#### Missing stuff

Support for security requirements object inside operations is still pending. It shouldn't be hard to add now, so feel free to make a PR :wink:.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,8 @@ It means [processes](#definitionsProcess) can subscribe to `event.user.signup` t
 		- [Reference Object](#referenceObject)
 		- [Schema Object](#schemaObject)
 		- [XML Object](#xmlObject)
+    - [Security Scheme Object](#securitySchemeObject)
+    - [Security Requirement Object](#securityRequirementObject)
 	- [Specification Extensions](#specificationExtensions)
 
 <!-- /TOC -->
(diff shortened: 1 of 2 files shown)


fmvilas:
@MikeRalphson @bpedro @ctron your comments are welcome 😉

Issue #34, opened 2017-08-17
IANA service name for secure MQTT is not MQTTS


fmvilas:
IANA service name for secure MQTT is `secure-mqtt` and not `mqtts`.

https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=secure-mqtt

This:
```yaml
servers:
  - url: api.company.com
    scheme: mqtts
```

should become this:

```yaml
servers:
  - url: api.company.com
    scheme: secure-mqtt
```

PR #45, opened 2017-11-12
Rename mqtts to secure-mqtt

fmvilas:
It now uses [IANA name for secure MQTT](https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=secure-mqtt) (`secure-mqtt`) instead of `mqtts`.

The change is backward compatible. The schema still supports `mqtts` but it will be removed on the next major version.


fmvilas:
It solves #34.

PR #70, opened 2018-03-30
Fix asyncapi version

### LINKED ISSUE #69: Version 1.1
Is this a new version of the spec, or a 1.1 revision of the 1.0.0 spec?

I.e. is a topic parameter valid in an asyncapi document which uses `asyncapi: 1.0.0` ?

https://github.com/asyncapi/asyncapi/blob/3a45dad21f4536c1a87c2ca0aea1e651e95456de/test/docs/sample.yml#L1

fmvilas:
It fixes #69.

### DIFF
--- a/schema/asyncapi.json
+++ b/schema/asyncapi.json
@@ -1,5 +1,5 @@
 {
-  "title": "AsyncAPI 1.0 schema.",
+  "title": "AsyncAPI 1.1.0 schema.",
   "id": "http://asyncapi.hitchhq.com/v1/schema.json#",
   "$schema": "http://json-schema.org/draft-04/schema#",
   "type": "object",
--- a/test/docs/anyof.yml
+++ b/test/docs/anyof.yml
@@ -1,4 +1,4 @@
-asyncapi: '1.0.0'
+asyncapi: '1.1.0'
 info:
   title: AnyOf example
   version: '1.0.0'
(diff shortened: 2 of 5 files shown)


fmvilas:
https://github.com/asyncapi/asyncapi/blob/3a45dad21f4536c1a87c2ca0aea1e651e95456de/schema/asyncapi.json#L21

@MikeRalphson I wonder if this line should be removed or not. Version 1.1.0 should be compatible with 1.0.0 documents, however, there's no simple way to check if this "version 1.0.0 document" is using features from 1.1.0, as you reported in #69.

What's your opinion here?

MikeRalphson:
The draft OpenAPI 3.0.x schema "solves" this in the following way:

```yaml
"pattern": "^3\\.0\\.\\d(-.+)?$"
```

But I assume separate 3.0.x and 3.1.x schemas would be produced, and in the absence of a 3.1 version yet, we don't know how that would be handled.

As you say, if you allow the 1.1 schema to validate 1.0 documents there is no way to flip a switch to disable the new features.

I would defer the decision to the tool which is doing the validating, it is that tool's job to use the correct schema. Thus leaving the line as is, or converting to a pattern would be the right thing to do.

fmvilas:
Agree. I think it's tooling responsibility to check for it. In this case, I think we should decouple the Node.js package version from the spec version. And probably create another one (or another major version) that will offer "a list" of version schemas so that tooling can choose.

MikeRalphson:
> I think we should decouple the Node.js package version from the spec version

Yes, otherwise you hit the problem of not being able to depend on multiple versions of the same package to pull in the schemas.

fmvilas:
Exactly. Cool! Thanks for the help!

PR #73, opened 2018-04-06
Re-usable parameters, fixes #71

### LINKED ISSUE #71: Allow re-usable `parameters` reffed to `components`
See https://github.com/asyncapi/asyncapi/blob/0684308ae117a2cf2b710613989aa65ea261a1f6/test/docs/sample.yml#L34-L45

where the necessary `streetlightId` parameter is only defined for one topic, but it would be useful to be able to `$ref` a re-usable *Parameter Object*

If approved will target a PR at ... v1.2.0 ?

MikeRalphson:
* Parameter object expanded to include `$ref` property, as per schema etc
* Components object expanded to include `parameters` property
* Schema updated
* Sample doc updated
* Version bumped to 1.2.0 speculatively
* Minor fix to an example to ensure they all parse ok (mdv)
* "Open API Intiative" and "Open API Specification" -> "OpenAPI" as per recent change

Test passes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,9 +2,9 @@
 
 #### Disclaimer
 
-Part of this content has been taken from the great work done by the folks at the [Open API Initiative](https://openapis.org). Mainly because **it's a great work** and we want to keep as much compatibility as possible with the [Open API Specification](https://github.com/OAI/OpenAPI-Specification).
+Part of this content has been taken from the great work done by the folks at the [OpenAPI Initiative](https://openapis.org). Mainly because **it's a great work** and we want to keep as much compatibility as possible with the [OpenAPI Specification](https://github.com/OAI/OpenAPI-Specification).
 
-#### Version 1.1.0
+#### Version 1.2.0
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](http://www.ietf.org/rfc/rfc2119.txt).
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -482,7 +482,7 @@ Field Name | Type | Description
...
 <a name="topicItemObjectSubscribe"></a>subscribe | [Message Object](#messageObject) &#124; Map[`"oneOf"`, [[Message Object](#messageObject)]] | A definition of the message a SUBSCRIBE operation will receive on this topic. `oneOf` is allowed here to specify multiple messages, however, **a message MUST be valid only against one of the referenced message objects.**
 <a name="topicItemObjectPublish"></a>publish | [Message Object](#messageObject) &#124; Map[`"oneOf"`, [[Message Object](#messageObject)]] | A definition of the message a PUBLISH operation will receive on this topic. `oneOf` is allowed here to specify multiple messages, however, **a message MUST be valid only against one of the referenced message objects.**
-<a name="topicItemObjectParameters"></a>parameters | [[Parameter Object](#parameterObject)] | A list of the parameters included in the topic name, if using [topic templating](#definitionsTopicTemplating).
+<a name="topicItemObjectParameters"></a>parameters | [[Parameter Object](#parameterObject) &#124; [Reference Object](#referenceObject)] | A list of the parameters included in the topic name, if using [topic templating](#definitionsTopicTemplating).
```
fmvilas: I think the type here is not correct. Shouldn't it be like the following?

```markdown
[[Parameter Object](#parameterObject)] &#124; [[Reference Object](#referenceObject)]
```

So it renders as:

[[Parameter Object](#parameterObject)] &#124; [[Reference Object](#referenceObject)]

instead of:

[[Parameter Object](#parameterObject) &#124; [Reference Object](#referenceObject)]

schema/asyncapi.json
```diff
@@ -19,7 +19,8 @@
       "type": "string",
       "enum": [
         "1.0.0",
-        "1.1.0"
+        "1.1.0",
+	"1.2.0"
```
fmvilas: Can you fix indentation here?

README.md
fmvilas: Oh you're right. Forget about it then.

fmvilas:
Merged. Thanks mate!

Issue #93, opened 2018-10-04
No OAuth support in specification


SensibleWood:
Hey, just wondering on the rationale for not including OAuth support in the Security Schemes?

Whilst this may not make sense in many protocols in the WebSocket case presenting an access token which is granted an appropriate scope when connecting over HTTPS (and expected an upgrade to WSS) seems to make sense to me.

Describing this schematically would be really useful (especially for the work I'm doing ATM, which is exactly this scenario).

fmvilas:
You're right. WebSockets support is relatively new, and I wanted to keep it simple for the first release, but that's definitely something that needs to be there in future releases.

SensibleWood:
Happy to help with it! I'll work it up and let you know when its ready for review

fmvilas:
Awesome Chris! Let me know if you need help along the way. I'll be happy to help!

PR #95, opened 2018-10-12
Changes to address issue #93 on OAuth support

SensibleWood:
Summary of changes:

* Pulled in OAuth objects and samples from OpenAPI specification
* Also pulled in OpenID Connect discovery (as this is very useful, although there's some questions on scope that need to be address IMO - for another time...)
* Updated JSON schema with new objects
* Added relevant objects to test doc
* Blew out the list of supported security standards into a bullet point list - to aid readability
* Added package-lock.json


schema/asyncapi.json
```diff
@@ -1,6 +1,6 @@
 {
-  "title": "AsyncAPI 1.2.0 schema.",
-  "id": "http://asyncapi.hitchhq.com/v1/schema.json#",
+  "title": "AsyncAPI 2.0.0 schema.",
+  "id": "http://asyncapi.hitchhq.com/v2/schema.json#",
```
fmvilas: Let's not raise the version here. I'll do it when releasing a new version. BTW, is there any breaking change in this PR?

SensibleWood: This should be non-breaking as I've only added new objects Actually thought of another useful change to the schema on the OAuth flow objects so I've bring that in as well when I correct the version number

SensibleWood: Changes added as request - please see updates in pull request

README.md
```diff
@@ -1922,10 +1932,101 @@ scheme: bearer
...
+```
 
+#### <a name="oauthFlowsObject"></a>OAuth Flows Object
 
+Allows configuration of the supported OAuth Flows.
+
+##### Fixed Fields
+Field Name | Type | Description
+---|:---:|---
+<a name="oauthFlowsImplicit"></a>implicit| [OAuth Flow Object](#oauthFlowObject) | Configuration for the OAuth Implicit flow
```
mattheworiordan: A meaningless comment I know, but should you not be consistent with spacing i.e. 
```
<a name="oauthFlowsImplicit"></a>implicit | [OAuth Flow Object](#oauthFlowObject) | Configuration for the OAuth Implicit flow 
```
if you want to be consistent with the table heading columns and most other columns (apart from `---|:---:|---`)

SensibleWood: 👍 actually a cut-and-paste from the OpenAPI spec - but I'll check for it in future

PR #104, opened 2019-01-23
Add support for coap scheme

MikeRalphson:
Adds support for the Constrained Application Protocol [CoAP](https://en.wikipedia.org/wiki/Constrained_Application_Protocol) as a `scheme`.

Also add missing `jms` value from `scheme` description.

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,7 +298,7 @@ An object representing a Server.
 Field Name | Type | Description
 ---|:---:|---
 <a name="serverObjectUrl"></a>url | `string` | **REQUIRED**. A URL to the target host.  This URL supports Server Variables and MAY be relative, to indicate that the host location is relative to the location where the AsyncAPI document is being served. Variable substitutions will be made when a variable is named in `{`brackets`}`.
-<a name="serverObjectScheme"></a>scheme | `string` | **REQUIRED**. The scheme this URL supports for connection. The value MUST be one of the following: `kafka`, `kafka-secure`, `amqp`, `amqps`, `mqtt`, `secure-mqtt`, `ws`, `wss`, `stomp`, `stomps`.
+<a name="serverObjectScheme"></a>scheme | `string` | **REQUIRED**. The scheme this URL supports for connection. The value MUST be one of the following: `kafka`, `kafka-secure`, `amqp`, `amqps`, `mqtt`, `secure-mqtt`, `ws`, `wss`, `stomp`, `stomps`, `jms` or `coap`.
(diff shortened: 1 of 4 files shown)


fmvilas:
But, how do you plan to add support to CoAP in AsyncAPI? AFAIK, it has the same semantics as HTTP, i.e., GET, PUT, POST, etc. We only have publish and subscribe. Am I missing something?

fmvilas:
The underlying protocol is UDP, but my guess is that people would want to describe the operations (GET, POST, PUT, etc.) in the AsyncAPI file, don't you think? The same way OpenAPI describes HTTP and not TCP, even though it's over TCP. Could you provide an example AsyncAPI document of how you see both, AsyncAPI and CoAP, playing together? It's just for me to be able to see it 🤔

MikeRalphson:
OK, will do. The library I'm using abstracts away the HTTP semantics, so I will see if it has a lower layer to play with. I'll report back.

MikeRalphson:
Agreed this is a better fit for OpenAPI. Will resubmit just the missing `jms` entry in the `scheme` description.

beriberikix:
👋 new to AsyncAPI but a user of CoAP. Agree that the request-response parts of CoAP can mostly be captured via OAS but as Mike mentioned the "group communication" feels inline with the spirit of AsyncAPI. The initial CoAP RFC 7252 didn't go into much detail on how to use multicast but [RFC7390](https://tools.ietf.org/html/rfc7390) does and is what most people are implementing.

I need to dig deeper into how to translate the RFC into an AsyncAPI example but I've reaching out to the CoAP WG for feedback.

P.S. is there a shorthand for AsyncAPI like OAS? 😃

derberg:
@beriberikix thanks for looking into it

I saw someone one day on twitter wrote AAS but that looks/sounds weird 😄

pdgendt:
Just bumped into this (old) PR, any chance the current (2.6.0) spec would be able to describe CoAP resources, OAS seems to be a lesser fit with regards to the group communication an also observable resources.

jonaslagoni:
This might be [perfect for v3 of AsyncAPI](https://v3.asyncapi.com/blog/release-notes-3.0.0#requestreply) as we now support request/reply pattern i.e. enable http protocols better.

@pdgendt @beriberikix if you want this, I would suggest to create an issue for it, and then if you are up for it, [champion](https://github.com/asyncapi/spec/blob/master/CONTRIBUTING.md#rfc-contribution-champions) the change in maybe v3.1? 🤔

PR #124, opened 2019-02-07
Normalize topics as URI path templates

### LINKED ISSUE #109: [FEATURE REQUEST] Normalize topics using URL format
### Is your feature request related to a problem? Please describe.
Different brokers have different topic separators. AQMP uses dots, MQTT uses slashes, and Kafka doesn't have separators. Since we support topic parameters, we have to standardize what's the character that separates topic chunks. Also, things are confusing when you have to concat `baseTopic` and a topic. So far, we've been using a dot, but it makes it harder to parse topics. By having topics as URL paths, we make it easier for tooling to parse them, and could even get rid of the `events` section of the spec, because it will be the same as putting the operations under the `/`. It will also align things with OpenAPI paths.

### Can't it be tackled using specification extensions?
No.

### Describe the solution you'd like
At a glance, the solution would look like the following document:

```yaml
topics:
  /user/updated:
    ...
  user/updated: # Different from the one above (notice the absence of slash in the beginning of the topic)
    ...  
  user/{id}/updated: # Example with a topic parameter
    ...
```

New topic format would comply with [URI templates RFC 6570](https://tools.ietf.org/html/rfc6570).

It's now the duty of a specific protocol mapping to define how this URL translates to the underlying protocol's topic naming convention.

##### AMQP
`user/updated` ➡️ `user.updated`
`/user/updated` ➡️ `.user.updated`
`user/{id}/updated` ➡️ `user.*.updated`
`/user/{id}/updated` ➡️ `.user.*.updated`

> Not sure a topic starting with a period is possible in AMQP. In any case, that's AMQP protocol mapping duty to define it.

##### Kafka
`user/updated` ➡️ `user/updated`
`/user/updated` ➡️ `/user/updated`
`user/{id}/updated` ➡️ (not supported)
`/user/{id}/updated` ➡️ (not supported)

### Describe alternatives you've considered
I've considered concatenating `baseTopic` and topics without any separator in the middle. I also considered making the separator configurable.

fmvilas:
### What?
Fixes #109 

##### Summary of changes:
* Base topic is now part of the server object and has been removed from the root object. They MUST conform to [RFC 3986 URI path](https://tools.ietf.org/html/rfc3986#section-3.3). Defaults to an empty string.
* Topics MUST now conform to [RFC 6570](https://tools.ietf.org/html/rfc6570).
* Topic resolution is now based on URI resolving algorithms. So, for instance:
  - `my/base/topic` + `my/event` = `my/base/topic/my/event`
  - `my/base/topic` + `/my/event` = `/my/event`
* Upgrades JSON schema version to Draft-07 (latest).

### Why?
See #109


MikeRalphson:
To clarify, this PR changes the JSON Schema draft to 07 **used by the AsyncAPI schema instances**? It does not update the use of JSON Schema Draft 07 in the `schemaObjects` used within AsyncAPI documents?

fmvilas:
Correct. This is an update for our schema. Schema objects are still OpenAPI flavor based on draft-04.

rmelian:
@fmvilas my only comment is what I told you over slack: To create a separate example file for 2.0 and not to override 1.2

fmvilas:
@rmelian added changes as per your comments. Thanks!

versions/next/asyncapi.md
```diff
@@ -304,6 +279,7 @@ Field Name | Type | Description
 <a name="serverObjectSchemeVersion"></a>schemeVersion | `string` | The version of the scheme. For instance: AMQP `0.9.1`, Kafka `1.0.0`, etc.
 <a name="serverObjectDescription"></a>description | `string` | An optional string describing the host designated by the URL. [CommonMark syntax](http://spec.commonmark.org/) MAY be used for rich text representation.
 <a name="serverObjectVariables"></a>variables | Map[`string`, [Server Variable Object](#serverVariableObject)] | A map between a variable name and its value.  The value is used for substitution in the server's URL template.
+<a name="serverObjectBaseTopic"></a>baseTopic | `string` | A string describing the base topic. MUST be in the form of a [RFC 3986 URI path](https://tools.ietf.org/html/rfc3986#section-3.3). URI templates are not allowed here. Defaults to empty string.
```
MikeRalphson: I wonder what purpose a specific `baseTopic` property has now? Can't the use either provide a base-topic by simply appending a constant value to the `url` property, or provide a templated variable with the usual `default` and `enum` properties?

fmvilas: Unlike HTTP (or OpenAPI basePath), baseTopic is not part of the URL. You can't append the baseTopic to the server URL because it will never connect to the server. However, I think it's a good idea to have it inside the server object because different brokers might have different prepended base topics.

Example:
```
URL: mqtt://broker.company.com
Subscribe: user/created
```

is not the same as

```
URL: mqtt://broker.company.com/user
Subscribe: created
```

It will complain there's no MQTT server listening at `mqtt://broker.company.com/user`.

MikeRalphson: OK, understood.

MikeRalphson: Should we say explicitly that the `url` property cannot contain a `path` or `fragment-id`, or maybe more strictly that it may only contain a `scheme` and a `host` if that is the case?

fmvilas: Sorry, my previous example was not clear enough. It doesn't mean you can't have paths or fragment identifiers. A broker URL might look like `amqp://broker.company.com/vhost/1?token=mysecrettoken`, and it's totally valid. Check out [AMQP URI scheme](https://www.rabbitmq.com/uri-spec.html) (scroll down to `2. The "amqp" URI scheme`).

And that's not just AMQP specific. You can stablish a WebSockets connection in a URI similar to the one above. All protocols allow you to pass a fully qualified URI. Not to mention that the server URI might just be the URI of a proxy in the middle.

To your point, fragment identifiers might not be valid since, as per my understanding, they're just for client-side handling, mime-type dependent, and should not produce any impact server-side.

fmvilas:

PR #125, opened 2019-02-07
Fix wording and link for meeting agendas

fmvilas:
### What?
As @derberg correctly pointed out [here](https://github.com/asyncapi/asyncapi/pull/123/files#r254436884), I'm changing the words and the link for past and future meeting agendas.

### Why?
It was really confusing.


derberg:

Issue #122, opened 2019-02-06
Create a team/contributors document


fmvilas:
As a contributor, I'd like to get recognition for my contributions and appear as a project contributor/maintainer.

derberg:
do you have something specific in mind
I bumped into this interesting solution https://all-contributors.js.org/ 
example: https://github.com/jaredpalmer/formik#contributors

fmvilas:
That looks great! Would you like to pick this issue?

derberg:
definitely 💃

derberg:
So far I integrated all-contributors only with asyncapi-react as I have rights only there.

- you can talk to bot any time in any context, issue or PR. Example issue: https://github.com/asyncapi/asyncapi-react/issues/36
- bot replies he hears you and creates a pull request with change in readme file to add new contributor. Example https://github.com/asyncapi/asyncapi-react/pull/37 
- not only code contrib is valuable, this is why it supports different emojis https://github.com/all-contributors/all-contributors#emoji-key
- readme gets a badge about contributors and a section with nice overview of contributors

derberg:
summary:

- advantage
  - automation with bot, flexibility
  - nice outcome
  - emojis do not have to be used
- disadvantage 
  - bot creates branches in main repo, so if you have strict rule that all work comes in through forks, this might be a problem for ya
  - bot shows as one of contributors https://github.com/asyncapi/asyncapi-react/graphs/contributors, but maybe this is also advantage as he is the only one making those additions and by this approach it is more clean, the contributors section is clear thanks to it

derberg:
@fmvilas let me know whay you think, remember I'm out next week so if I need to join SIG meeting to talk about that, I'll do that but week after

fmvilas:
I just made available the @all-contributors app to all the repos in the organization.

fmvilas:
LOL. Ok let's use it!

fmvilas:
@all-contributors please add @derberg for doc, ideas and review

fmvilas:
@all-contributors please add @MikeRalphson for question, doc, code, ideas, infra, review and test.

fmvilas:
@all-contributors please add @rmelian for question, bug, code, doc, ideas, maintenance, review, tool and test.

fmvilas:
@all-contributors please add @MikeRalphson for question, doc, code, ideas, infra, review and test.

fmvilas:
[...]

PR #131, opened 2019-02-08
Run tests with mdv and apply mdv only to versions directory

fmvilas:
### What?
I'm making `npm test` to run "tests" and mdv. Also, I'm limiting mdv to check for versions directory only.

### Why?
So it doesn't bother on README with all-contributors bot.


fmvilas:
I'm merging this one because I want to continue working on #122.

PR #139, opened 2019-02-11
Automate testing against all examples

fmvilas:
### What?
I just improved "tests" a little bit. It now runs tests for every document in the examples directory. You can also select which versions you want to run.

### Example

Run tests for specific versions:
```bash
npm test -- --versions 1.2.0,next
```

### Why?
It's very convenient while working on the spec.

### Screenshots
screen shot 2019-02-11 at 11 50 06

screen shot 2019-02-11 at 11 51 08

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,3 +3,4 @@ node_js:
   - "6"
 script:
   - npm run test
+  - npm run mdv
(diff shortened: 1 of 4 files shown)


fmvilas:

PR #141, opened 2019-02-12
Add scope document

### LINKED ISSUE #119: [FEATURE REQUEST]
As a user, I'd like to understand what's the vision and scope of AsyncAPI. I should be clear what is part of the scope and what i
[vision-document-template-v1.pdf](https://github.com/asyncapi/asyncapi/files/2835480/vision-document-template-v1.pdf)
s not, and where the project is heading to.

@jhigginbotham did some great questions that must be answered in this document:

* What is the overall scope?
* Is it just to capture events and streams available?
* Is it a higher-level view of all interaction styles, e.g. RPC, REST, message streams, events, file drops/FTP
* What is the end goal/stopping point for the spec - drive definition capture, doc generation, other?
* How does this fit with OAS?
* How does this fit with ALPS?
* Do we support any/all of the following technologies? e.g. gRPC (has bi-dir support), Websockets, SSE, Webhooks, Kafka, and Kinesis streams
* What formats do we support? e.g. JSON, XML, Avro, ProtoBuf, Binary

References:

* https://wiki.base22.com/btg/template-vision-and-scope-document-20480031.html
* http://trishmarie.com/vision.pdf
* See attached documents

fmvilas:
### What?
Fixes #119.

### Why?
[Let's make AsyncAPI a welcoming community](https://github.com/asyncapi/asyncapi/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%22Make+AsyncAPI+a+welcoming+community%22)

### DIFF
--- a/SCOPE.md
+++ b/SCOPE.md
@@ -1,0 +1,32 @@
+# AsyncAPI Scope
+
+AsyncAPI is a specification for defining message-driven APIs. We're on a mission to standardize message-based communication and increase interoperability of the different types of systems out there.
+
+## Scope
+
+The following is a list of features the AsyncAPI specification and its official tooling should offer:
+
+- Provide enough information to connect and consume message-driven APIs. An AsyncAPI document should provide all your software needs to connect to a given API and start publishing and/or consuming messages from it.
+- Offer a framework for validating message payloads and headers.
+- Allow you to strictly define your message-driven API in a machine-readable format.
+- Provide mechanisms for mapping the information contained in an AsyncAPI document to the major messaging protocols.
+- Protocol-specific features, such as queues, exchanges, policies, partitions, etc; always via protocol mappers.
+- Provide a standard way to create specification extensions and promote their discoverability. 
+- Offer mechanisms to define message schemas in different formats. Namely: JSON Schema, OpenAPI schema (_slightly different than JSON Schema Draft 04_), Protobuf, Avro, etc.
+- Provide parsers for all the major programming languages, so it's easier for others to create tooling.
+- Bring human-readable documentation generators in React, HTML, and Markdown formats.
+- Offer code generators for all the major programming languages.
+
+## Out of scope
+
+- Solutions for REST, RPC, gRPC, GraphQL.
+- HTTP APIs, except HTTP streaming.
+- Everything that's provided by OpenAPI.
+
+## Compatiblity with OpenAPI
+
+AsyncAPI was born out of the OpenAPI specification. The idea was to provide a similar format to describe message-driven APIs. Therefore, you'll find many similarities between both.
+
+There's especially one of the components that's exactly the same between both: schemas. We made it on purpose so you could re-use OpenAPI schemas in your AsyncAPI documents and viceversa.
+
+Although AsyncAPI is not part of the OpenAPI Initiative (OAI), we regularly work together in order to make both of them as compatible as possible.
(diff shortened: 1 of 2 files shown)


SCOPE.md
```diff
@@ -0,0 +1,32 @@
+# AsyncAPI Scope
+
+AsyncAPI is a specification for defining message-driven APIs. We're on a mission to standardize message-based communication and increase interoperability of the different types of systems out there.
```
jhigginbotham: out there -> available

SCOPE.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Scope
+
+The following is a list of features the AsyncAPI specification and its official tooling should offer:
+
+- Provide enough information to connect and consume message-driven APIs. An AsyncAPI document should provide all your software needs to connect to a given API and start publishing and/or consuming messages from it.
```
jhigginbotham: "all your software needs" - as a specification, it isn't providing software. It is providing a description of the existing async APIs available for consumption, right?

SCOPE.md
```diff
@@ -0,0 +1,32 @@
...
+- Offer a framework for validating message payloads and headers.
+- Allow you to strictly define your message-driven API in a machine-readable format.
+- Provide mechanisms for mapping the information contained in an AsyncAPI document to the major messaging protocols.
```
jhigginbotham: mechanisms, or a definition language to describe protocols available for publishing and consuming messages?

SCOPE.md
```diff
@@ -0,0 +1,32 @@
...
+- Protocol-specific features, such as queues, exchanges, policies, partitions, etc; always via protocol mappers.
+- Provide a standard way to create specification extensions and promote their discoverability. 
+- Offer mechanisms to define message schemas in different formats. Namely: JSON Schema, OpenAPI schema (_slightly different than JSON Schema Draft 04_), Protobuf, Avro, etc.
+- Provide parsers for all the major programming languages, so it's easier for others to create tooling.
```
jhigginbotham: Perhaps separate the definition spec from the tooling, as not to create confusion like what we saw with Swagger vs. OAS?!

SCOPE.md
fmvilas: Maybe like this?

An AsyncAPI document should provide **everything** your software needs to connect to a given API and start publishing and/or consuming messages from it.

It doesn't say it provides software. It provides all the information your software will need to connect and consume (maybe should be like this?). In any case, AsyncAPI also provides tooling for that purpose so it offers software, not just the spec.

SCOPE.md
fmvilas: Both, a definition language and tools to process these definitions.

SCOPE.md
fmvilas: I'd like to keep it part of the same thing. The problem between Swagger and OAS is that one is part of a company and the other part of a foundation. I want to keep both together, always. Most people don't care about specs, but they do about the tooling. Ultimately is what you use. That's one of the reasons it's being hard for the OAS folks to make people say OpenAPI and not Swagger.

SCOPE.md
MikeRalphson: To clarify is "official tooling" both that which is originated within, or donated to, the AsyncAPI github organisation and no others?

fmvilas: So far yes. The ones the AsyncAPI is responsible for maintaining.

SCOPE.md
jhigginbotham: Fair enough. May I suggest then to rework the scope and such and call out the spec + tooling as two separate offerings as part of the initiative? The bullets indicate this but aren't grouped, so it becomes confusing. I could see readers asking "is it a spec? Is it a solution? Is it tools?" Simply reorganizing and clarifying the intent will go far, I think.

fmvilas: Agree. It should be clear.

PR #149, opened 2019-02-20
Move security requirements to server object

### LINKED ISSUE #48: Move security requirements to server objects
Many async APIs out there have proved to support different protocols for the same API. For instance, they might support using WebSockets and MQTT, HTTP streaming and WebSockets, etc.

The way you authenticate to these APIs depend very much on the protocol, so I think it would make sense to attach the security requirement to the server you're connecting to.

**Example:**
```yaml
servers:
  # This server supports connecting using user-password or X509 certificate
  - url: async.api.com:1883
    scheme: mqtt
    securityRequirements:
      - X509Certificate
      - userPass
  # This server only supports connecting using user-password
  - url: async.api.com
    scheme: amqp
    securityRequirements:
      - userPass

components:
  securitySchemes:
    userPass:
      type: userPassword
    X509Certificate:
      type: X509
```

Right now the security requirements are at the root level, making it impossible to guess which server supports which security scheme.

fmvilas:
### What?
Fixes #48 

### Why?
It makes more sense to have the security requirement as part of the server object since auth mechanisms might change depending on the server.

### DIFF
--- a/examples/next/gitter-streaming.yml
+++ b/examples/next/gitter-streaming.yml
@@ -15,9 +15,8 @@ servers:
         enum:
           - chatMessages
           - events
-
-security:
-  - httpBearerToken: []
+    security:
+      - httpBearerToken: []
 
 stream:
   framing:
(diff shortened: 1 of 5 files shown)


MikeRalphson:
I think over at OAS, one reason not to do this was that it was seen that differing security meant the signature of the API was different, whereas servers was more about alternative or additional environments. However, I do see the likelihood of dev / staging / live for example having different security requirements and OAS !== AsyncAPI.

PR #150, opened 2019-02-20
Remove mqtts from the list of supported schemes

### LINKED ISSUE #46: Remove support for mqtts
[In this PR](https://github.com/asyncapi/asyncapi/pull/45), support for `mqtts` was deprecated in favor of `secure-mqtt`. In v2 we should totally remove it from the schema file.

fmvilas:
### What?
Fixes #46 

Removes mqtts in favor of secure-mqtt, which is the right IANA id for the protocol.

### Why?
See #46 for more details.

### DIFF
--- a/versions/next/schema.json
+++ b/versions/next/schema.json
@@ -199,7 +199,6 @@
             "amqp",
             "amqps",
             "mqtt",
-            "mqtts",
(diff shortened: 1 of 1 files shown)


fmvilas:
Not needed since #151 got merged.

Issue #77, opened 2018-06-24
attribute to identify function to call


chris3081:
So the idea is to have an attribute that can identify what attribute to call when a message is pulled from the queue. 

```
components:
  messages:
     HelloWorld:
       summary: Hello World
       payload:
         type: object
         properties:
           arga:
             type: string
           argb:
             type: string
       x-operation-id: "helloworld.say"
```

Where Helloworld is a python file and say is a method in it

A sample function for calling the operation would look like this
```
    def callmethod(self, message, **kwargs):
        try:
            if 'x-osperation-id' in self._config['components']['messages'][message].keys():
                methodlookup = self._config['components']['messages'][message]['x-operation-id']
                library, func = methodlookup.rsplit('.', 1)
                i = importlib.import_module(library)
                method_to_call = getattr(i, func)
                return method_to_call(**kwargs)
        except Exception as e:
            raise e
        return None
```

fmvilas:
Thanks for raising the issue @chris3081.

I guess the debate here is if we should standardize `operationId`. My bet is that we should do it in order to increase compatibility with OpenAPI 3. However, it's not that simple in AsyncAPI because there's no such thing as an "operation object". The `publish`/`subscribe` operations are, in fact, message objects. So adding `operationId` as an attribute of "message" would couple messages to operations, 1:1, and that's an architecture decision the spec should not be imposing.

That said, we can rework the spec in the next major version to have operation objects separate from messages. Something like the following:

```yaml
publish:
  operationId: sayHello
  summary: Says hello.
  description: Publish a message to say hello.
  message:
   summary: A hello message
   payload:
     type: object
     properties:
       arga:
         type: string
       argb:
         type: string
```

This would be more similar to OpenAPI 3. But, just to make it clear, **this should be done in a next major version as it's a breaking change**. In the meantime, your solution looks good enough to bypass the limitation.

aerfio:
@fmvilas are you sure the assignment is right? 😛

fmvilas:
Aren't you going to work on this? :trollface: Sorry lad, fixing it now.

aerfio:
No problem 😃

Issue #54, opened 2018-01-16
Allow for payload media types to be documented


wout3r:
Right now, AsyncAPI is missing a way to document the payloads' media type (e.g., JSON, XML, ...). 
Can the specification be enriched to make this possible?

The approach for documenting media types could be similar to the one proposed in the OpenAPI specifications (v2 & v3):

1. The AsyncAPI Object could be extended with a 'defaultPayloadMediaType' property that allows to specify the default payload media type (e.g., with "application/json" as default value).

2. The type of the Message.payload property could be changed to "Map[string (media type), Schema Object] | Schema Object",  which allows the default payload media type to be overridden at the payload-level.

This would allow API specifications such as:
```
asyncapi: '1.0.0'
...
defaultPayloadMediaType: text/xml

topics:
  X:
    publish:
      payload:
        type: object
        properties:
          ...

  J:
    publish:
      payload:
        application/json:
          type: object
          properties:
            ...

  O:
    subscribe:
      payload:
        application/x-java-serialized-object:
          type: object
          properties:
            ...
```

Thanks in advance for considering this feature request!

fmvilas:
That's a great idea. Thanks for the feedback! I'm marking it for version 2.0.0.

marchon:
Not to be repetitive (I know when all you have is a hammer - everything can look like a nail - but I have solved a great many problems using the built in header capabilities of HTTP, and have always wanted to extend them to message processing. - I hope that you find this feedback both constructive, well-intentioned, and helpful. ) 

The RFC for HTTP [RFC 7231] includes a specification for "Content Negotiation" between the request agent and the response agent. Encapsulating a method of communication of preferred request and response metadata could solve a lot of not yet anticipated future problems. 


> 5.3.  Content Negotiation
> 
>    The following request header fields are sent by a user agent to
>    engage in proactive negotiation of the response content, as defined
>    in Section 3.4.1.  The preferences sent in these fields apply to any
>    content in the response, including representations of the target
>    resource, representations of error or processing status, and
>    potentially even the miscellaneous text strings that might appear
>    within the protocol.
> 
>    +-------------------+---------------+
>    | Header Field Name | Defined in... |
>    +-------------------+---------------+
>    | Accept            | Section 5.3.2 |
>    | Accept-Charset    | Section 5.3.3 |
>    | Accept-Encoding   | Section 5.3.4 |
>    | Accept-Language   | Section 5.3.5 |
>    +-------------------+---------------+
> 
> 
[...]

PR #158, opened 2019-02-23
Separate operation from message and change schema format definition mechanism

fmvilas:
### What?
Prepares the spec for #77 and #54.

It also changes how schema format is defined. Changed from this:

```yaml
channels:
  event/{streetlightId}/lighting/measured:
    subscribe:
      payload:
        'application/schema+json;version=draft-07':
          $ref: "#/components/schemas/lightMeasuredPayload"
```

to this:

```yaml
channels:
  event/{streetlightId}/lighting/measured:
    subscribe:
      message:
        schemaFormat: 'application/schema+json;version=draft-07'
        payload:
          $ref: "#/components/schemas/lightMeasuredPayload"
```

### Why?

We need to accommodate the spec to do things like the following:

```yaml
channels:
  event/{streetlightId}/lighting/measured:
    subscribe:
      operationId: receiveLightMeasure
      contentType: 'application/json'
      message:
        schemaFormat: 'application/schema+json;version=draft-07'
        payload:
          $ref: "#/components/schemas/lightMeasuredPayload"
```

Notice that we'll have 2 new properties: `operationId` and `contentType`. Also, and since we could only have a single schema format per payload, it doesn't make sense to create another nesting level for it, but instead we specify it as the `schemaFormat` property. This also allows us to default to AsyncAPI 1.2.0/OpenAPI 3.0.0 schema format when it's not present, making it easier to migrate.

### But what happened with the schema format for headers?
Investigating a bit more, I realized that neither Avro nor Protobuf really offer a solution for headers, so it should be enough to provide our own schema format for headers. No options there.

### DIFF
--- a/examples/next/anyof.yml
+++ b/examples/next/anyof.yml
@@ -6,16 +6,16 @@ info:
 channels:
   test:
     publish:
-      $ref: '#/components/messages/testMessages'
+      message:
+        $ref: '#/components/messages/testMessages'
 
 components:
   messages:
     testMessages:
       payload:
-        'application/schema+json;version=draft-07':
-          anyOf: # anyOf in payload schema
-            - $ref: "#/components/schemas/objectWithKey"
-            - $ref: "#/components/schemas/objectWithKey2"
+        anyOf: # anyOf in payload schema
+          - $ref: "#/components/schemas/objectWithKey"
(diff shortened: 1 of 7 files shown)


rmelian:
this LGTM

fmvilas:

PR #167, opened 2019-03-05
Issue #78 addition of Correlation ID to specification

### LINKED ISSUE #78: Correlation Ids
Would it be a good idea for AsyncAPI to support built-in correlation ids to enable distributed tracing ? 

I would see this as a (optional?) parameter correlationId under #/definitions/operation. The value of this parameter must be the name of one of properties defined in the message schema for the operation.

Automated tracing tools could be implemented with no knowledge of the internal implementation behind the APIs (these tools can listen to publish topics and aggregate on the values of correlationId). Clients of subscribe topics would be aware of the correlation semantics of a particular property in the message they must send to the subscriber of the topic.

SensibleWood:
Fixes #78 

As follows:

* Added Correlation ID object
* Added runtime expression support
* Added Correlation ID Map to Components Object
* Added Correlation ID to Message Object
* Updated schema and added example to test

A couple of bits of food for thought:

* Is the ABNF definition too restrictive?
* Is it right to match the ABNF definition in schema?
* **Would welcome suggestions as to whether more references are needed anywhere else in the specification as it stands i.e. defining a Correlation ID at a higher level than Message**

### DIFF
--- a/examples/next/correlation-id.yml
+++ b/examples/next/correlation-id.yml
@@ -1,0 +1,106 @@
+asyncapi: '2.0.0'
+id: 'urn:com:smartylighting:streetlights:server'
+info:
+  title: Correlation ID Example
+  version: '1.0.0'
+  description: A cut of the Streetlights API to test Correlation ID
+  license:
+    name: Apache 2.0
+    url: https://www.apache.org/licenses/LICENSE-2.0
+
+servers:
+  - url: api.streetlights.smartylighting.com:{port}
+    scheme: mqtt
+    description: Test broker
+    baseChannel: smartylighting/streetlights/1/0
+    variables:
+      port:
+        description: Secure connection (TLS) is available through port 8883.
+        default: '1883'
+        enum:
+          - '1883'
+          - '8883'
+    security:
+      - apiKey: []
+      - supportedOauthFlows:
+        - streetlights:on
+        - streetlights:off
+        - streetlights:dim
+      - openIdConnectWellKnown: []
+
+defaultContentType: application/json
+
+channels:
+  event/{streetlightId}/lighting/measured:
+    parameters:
+      - $ref: '#/components/parameters/streetlightId'
+    subscribe:
+      summary: Receive information about environmental lighting conditions of a particular streetlight.
+      operationId: receiveLightMeasurement
+      message:
+        $ref: '#/components/messages/lightMeasured'
+
+  action/{streetlightId}/dim:
+    parameters:
+      - $ref: '#/components/parameters/streetlightId'
+    publish:
+      operationId: dimLight
+      message:
+        $ref: '#/components/messages/dimLight'
+
+components:
+  messages:
+    lightMeasured:
+      name: lightMeasured
+      title: Light measured
+      summary: Inform about environmental lighting conditions for a particular streetlight.
+      correlationId:
+        location: "$message.header.correlationId"
+      contentType: application/json
+      payload:
+        $ref: "#/components/schemas/lightMeasuredPayload"
+    dimLight:
+      name: dimLight
+      title: Dim light
+      summary: Command a particular streetlight to dim the lights.
+      correlationId:
+        $ref: "#/components/correlationIds/sentAtCorrelator"
+      payload:
+        $ref: "#/components/schemas/dimLightPayload"
+
+  schemas:
+    lightMeasuredPayload:
+      type: object
+      properties:
+        lumens:
+          type: integer
+          minimum: 0
+          description: Light intensity measured in lumens.
+        sentAt:
+          $ref: "#/components/schemas/sentAt"
+    dimLightPayload:
+      type: object
+      properties:
+        percentage:
+          type: integer
+          description: Percentage to which the light should be dimmed to.
+          minimum: 0
+          maximum: 100
+        sentAt:
+          $ref: "#/components/schemas/sentAt"
+    sentAt:
+      type: string
+      format: date-time
+      description: Date and time when the message was sent.
+
+  parameters:
+    streetlightId:
+      name: streetlightId
+      description: The ID of the streetlight.
+      schema:
+        type: string
+
+  correlationIds:
+    sentAtCorrelator:
+      description: Data from message payload used as correlation ID
+      location: $message.payload.sentAt
--- a/examples/next/correlation-id.yml
+++ b/examples/next/correlation-id.yml
@@ -55,7 +55,7 @@ components:
       title: Light measured
       summary: Inform about environmental lighting conditions for a particular streetlight.
       correlationId:
-        location: "$message.header#/MQMD/CorrelationId"
+        location: "$message.header#/MQMD/CorrelId"
       contentType: application/json
       payload:
         $ref: "#/components/schemas/lightMeasuredPayload"
--- a/versions/next/asyncapi.md
+++ b/versions/next/asyncapi.md
@@ -2280,7 +2280,7 @@ The table below provides examples of runtime expressions and examples of their u
 
 Source Location | Example expression  | Notes
 ---|:---|:---|
-Message Header Property | `$message.header#/MQMD/CorrelationId` | Correlation ID is set using the `CorrelationID` value from the `MQMD` header.
+Message Header Property | `$message.header#/MQMD/CorrelId` | Correlation ID is set using the `CorrelId` value from the `MQMD` header.
 Message Payload Property | `$message.payload#/messageId` | Correlation ID is set using the `messageId` value from the message payload.
 
 Runtime expressions preserve the type of the referenced value.
--- a/versions/next/asyncapi.md
+++ b/versions/next/asyncapi.md
@@ -814,6 +814,7 @@ Field Name | Type | Description
 
 <a name="messageObjectHeaders"></a>headers | [Schema Wrapper Object](#schemaObject) | Definition of the message headers. It MAY or MAY NOT define the protocol headers.
 <a name="messageObjectPayload"></a>payload | `any` | Definition of the message payload. It can be of any type but defaults to [Schema object](#schemaObject).
+<a name="messageObjectCorrelationId"></a>correlationId | [Correlation ID Object](#correlationIdObject) \| [Reference Object](#referenceObject) | Definition of the correlation ID used for message tracing or matching.
 <a name="messageObjectSchemaFormat"></a>schemaFormat | `string` | A string containing the name of the schema format/language used to define the message payload. If omitted, implementations should parse the payload as a [Schema object](#schemaObject).
 <a name="messageObjectContentType"></a>contentType | `string` | The content type to use when encoding/decoding a message's payload. The value MUST be a specific media type (e.g. `application/json`). When omitted, the value MUST be the one specified on the [defaultContentType](#defaultContentTypeString) field.
 <a name="messageObjectName"></a>name | `string` | A machine-friendly name for the message.
@@ -860,6 +861,11 @@ This object can be extended with [Specification Extensions](#specificationExtens
         "$ref": "#/components/schemas/signup"
       }
     }
+  },
+  "correlationId": {
+    "description": "Default Correlation ID",
+    "type": "string",
+    "location": "$message.header.correlationId"
   }
 }
 ```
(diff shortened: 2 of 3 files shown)


fmvilas:
Added "Fixes #78" so the issue is closed when the PR gets merged.

versions/next/asyncapi.md
```diff
@@ -2208,6 +2231,68 @@ petstore_auth:
...
+type: string
+location: $message.header.correlationId
+```
+
+### <a name="runtimeExpression"></a>Runtime Expression
+
+A runtime expression allows values to be defined based on information that will be available within the message.
+This mechanism is used by [Correlation ID Object](#correlationIdObject).
+
+The runtime expression is defined by the following [ABNF](https://tools.ietf.org/html/rfc5234) syntax
```
fmvilas:
```suggestion
The runtime expression is defined by the following [ABNF](https://tools.ietf.org/html/rfc5234) syntax:
```

versions/next/asyncapi.md
```diff
@@ -2208,6 +2231,68 @@ petstore_auth:
...
+
+```
+      expression = ( "$message" )
+      source = ( header-reference | payload-reference )  
+      header-reference = "header." token
+      payload-reference = "payload." name
+      fragment = a JSON Pointer [RFC 6901](https://tools.ietf.org/html/rfc6901)
```
fmvilas: What's fragment doing here?

versions/next/asyncapi.md
```diff
@@ -2208,6 +2231,68 @@ petstore_auth:
...
+
+The runtime expression is defined by the following [ABNF](https://tools.ietf.org/html/rfc5234) syntax
+
+```
+      expression = ( "$message" )
+      source = ( header-reference | payload-reference )  
+      header-reference = "header." token
+      payload-reference = "payload." name
+      fragment = a JSON Pointer [RFC 6901](https://tools.ietf.org/html/rfc6901) 
+      name = *( char )
```
fmvilas: If `name` allows for "zero" chars, we could end up having something like `$message.payload.`.

What about this?
```suggestion
      name = 1*( char )
```

versions/next/asyncapi.md
```diff
@@ -2208,6 +2231,68 @@ petstore_auth:
...
+      char = as per RFC [7159](https://tools.ietf.org/html/rfc7159#section-7)
+      token = as per RFC [7230](https://tools.ietf.org/html/rfc7230#section-3.2.6)
+```
+
+The `name` identifier is case-sensitive, whereas `token` is not.
```
fmvilas: Why?

versions/next/asyncapi.md
```diff
@@ -2208,6 +2231,68 @@ petstore_auth:
...
+
+### <a name="runtimeExpression"></a>Runtime Expression
+
+A runtime expression allows values to be defined based on information that will be available within the message.
+This mechanism is used by [Correlation ID Object](#correlationIdObject).
+
+The runtime expression is defined by the following [ABNF](https://tools.ietf.org/html/rfc5234) syntax
+
+```
+      expression = ( "$message" )
```
fmvilas: I think there's something missing here:

```suggestion
      expression = ( "$message" source )
```

versions/next/asyncapi.md
```diff
@@ -2208,6 +2231,68 @@ petstore_auth:
...
+
+The `name` identifier is case-sensitive, whereas `token` is not. 
+
+The table below provides examples of runtime expressions and examples of their use in a value:
+
+##### <a name="runtimeExpressionExamples"></a>Examples
+
+Source Location | Example expression  | Notes
+---|:---|:---|
+Message Header Property | `$message.header.correlationId` | Headers are analogous to message properties
```
fmvilas:
```suggestion
Message Header Property | `$message.header.correlationId` | Headers are analogous to message headers.
```

versions/next/asyncapi.md
```diff
@@ -2208,6 +2231,68 @@ petstore_auth:
...
+Message Payload Property | `$message.payload.meta.messageId` | Correlation ID is set using a property from the `meta` section of a message
```
fmvilas:
```suggestion
Message Payload Property | `$message.payload.meta.messageId` | Correlation ID is set using a property from the `meta` section of a message.
```

versions/next/asyncapi.md
```diff
@@ -2208,6 +2231,68 @@ petstore_auth:
...
+
+Runtime expressions preserve the type of the referenced value.
```
fmvilas: I think this answers my suggestion about getting rid of `type` in correlation Ids, right?

versions/next/asyncapi.md
```diff
@@ -2208,6 +2231,68 @@ petstore_auth:
...
+Expressions can be embedded into string values by surrounding the expression with `{}` curly braces.
```
fmvilas: Do you have an example where this should be possible?

versions/next/asyncapi.md
SensibleWood: What I was getting here is that some messaging providers call headers properties, so giving some guidance for the reader if they understand the meaning of properties not properties

examples/next/correlation-id.yml
```diff
@@ -0,0 +1,106 @@
...
+        $ref: '#/components/messages/dimLight'
+
+components:
+  messages:
+    lightMeasured:
+      name: lightMeasured
+      title: Light measured
+      summary: Inform about environmental lighting conditions for a particular streetlight.
+      correlationId:
+        location: "$message.header.correlationId"
```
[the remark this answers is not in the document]
SensibleWood: `.header.` is the way this is specified in OpenAPI:

image

Personally I think we should be in sync.

examples/next/correlation-id.yml
```diff
@@ -0,0 +1,106 @@
...
+    streetlightId:
+      name: streetlightId
+      description: The ID of the streetlight.
+      schema:
+        type: string
+
+  correlationIds:
+    sentAtCorrelator:
+      description: Data from message payload used as correlation ID
+      location: $message.payload.sentAt
```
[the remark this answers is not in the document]
SensibleWood: `sentAt` is an attribute of the messages in the example:

```  dimLightPayload:
      type: object
      properties:
        percentage:
          type: integer
          description: Percentage to which the light should be dimmed to.
          minimum: 0
          maximum: 100
        sentAt:
          $ref: "#/components/schemas/sentAt"```

The idea of the example is to generalise the reference i.e. using the ABNF you don't need to have it specifically wedded to a schema. However, if it's confusing then I guess it needs to change. Will sort out a different approach.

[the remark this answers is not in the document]
Comment (fmvilas) [reply examples/next/correlation-id.yml]: Oh! I thought you forgot it here but now I see where it's being applied. In this case, it's fine. Thanks for clarifying!

File: examples/next/correlation-id.yml (hunk 10 above)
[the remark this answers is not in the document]
Comment (fmvilas) [reply examples/next/correlation-id.yml]: You're right, we should be in sync. I didn't know OpenAPI was using `header`. Thanks!

File: versions/next/asyncapi.md (hunk 6 above)
Comment (fmvilas) [reply versions/next/asyncapi.md]: Yeah, but that should be explained in the message headers section, don't you think so? This note seems confusing to me because we never called them `message properties` (on purpose) and suddenly it appears here.

File: examples/next/correlation-id.yml (hunk 11 above)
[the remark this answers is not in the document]
Comment (SensibleWood) [reply examples/next/correlation-id.yml]: OK no worries - will leave as is then

File: versions/next/asyncapi.md (hunk 6 above)
Comment (SensibleWood) [reply versions/next/asyncapi.md]: Sure sounds good - will add something appropriate and then change the wording here

File: versions/next/asyncapi.md [hunk 12]
```diff
@@ -2208,6 +2231,68 @@ petstore_auth:
...
+For specifying and computing the location of a Correlation ID a [runtime expression](#runtimeExpression) is used.
+
+##### Fixed Fields
+
+Field Name | Type | Description
+---|:---|---
+description | `string` | A optional description of the identifier. [CommonMark syntax](http://spec.commonmark.org/) can be used for rich text representation.
+type | `integer` \| `string` \| `number` | An optional type specification.
+location | {expression} | **REQUIRED.** A runtime expression that specifies the location of the correlation ID.
+
```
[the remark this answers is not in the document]
SensibleWood: OK sounds good - will add

versions/next/asyncapi.md
```diff
@@ -2208,6 +2231,68 @@ petstore_auth:
...
+An object that specifies an identifier at design time that can used for message tracing and correlation.
+
+For specifying and computing the location of a Correlation ID a [runtime expression](#runtimeExpression) is used.
+
+##### Fixed Fields
+
+Field Name | Type | Description
+---|:---|---
+description | `string` | A optional description of the identifier. [CommonMark syntax](http://spec.commonmark.org/) can be used for rich text representation.
+type | `integer` \| `string` \| `number` | An optional type specification.
```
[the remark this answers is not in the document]
SensibleWood: I was in two minds whether we need `type` or not - the design of this feature has moved on - so I'm going to remove it as it seems overkill.

versions/next/asyncapi.md
SensibleWood: Fragment is the last piece of the ABNF notation that points - in JSON Path syntax - to the specific property. So it's the `.correlationId` in `$message.header.correlationId`

versions/next/asyncapi.md
SensibleWood: Works for me

versions/next/asyncapi.md
SensibleWood: Yes it does - going to remove and update PR

versions/next/asyncapi.md
SensibleWood: There's a missing piece from the port from OpenAPI I need to add back in - will update the PR with changes

versions/next/asyncapi.md
SensibleWood: On review I need to remove the references to `token` - will change and update PR

versions/next/asyncapi.md
SensibleWood: Where you use a custom correlation ID? This would be down to API designers - if they want to enforce a correlation ID and cannot add a message header property then this offers the requisite features to use something in the payload. Alternatively, they may be retrofitting an API specification to a messaging platform and already use a custom correlation ID in the message payload.

Is this what you were getting at?

SensibleWood: You mean the curly braces? Will add an example

fmvilas: I'm more worried about where you can use runtime expressions. Specification-wise it's easy to say we support them, but tooling-wise it's not that simple. I want to make sure we only support things where we really need them.

SensibleWood: On reflection we don't support this at the moment. In OpenAPI it's a construct that relates to [Callback Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#callbackObject). I can't see a good fit so will remove the guidance.

SensibleWood: > I'm more worried about where you can use runtime expressions. Specification-wise it's easy to say we support them, but tooling-wise it's not that simple. I want to make sure we only support things where we really need them.

I think the runtime expression support is a really good investment for the future. If you look at the way Callback Object in OpenAPI can now be engineered it adds a great deal of flexibility for the API designer. Specifically, when we tackle RPC in earnest a runtime expression to specify replyTo queues and the like based on attributes of the message will give API designers a real boost IMHO.

fmvilas: Yeah, I get it. But I still can't see the value outside the correlation ID feature or a possible RPC support in the future. My concern is just that this sentence really looked like you can embed runtime expressions anywhere you have a string. Anyway, I think your previous comment clarified it :)

versions/next/asyncapi.md
SensibleWood: Have changed the examples and hopefully made them more useful

versions/next/asyncapi.md
SensibleWood: 👍

SensibleWood:
Made changes as requested. Please note that the ABNF syntax was incorrect and I've changed as follows:

* Removed `name`, `char` and `token`
* Both `header` and `payload` are now reliant on `fragment`

I think this fits better for headers as they can be nested e.g. if you look at WebSphere MQ they support (for example) CorrelationId inside the header MQMD - so using a JSON Pointer to provide a link to value seems like the right approach.

versions/next/asyncapi.md
fmvilas: Indeed. Thanks!

Issue #112, opened 2019-02-03
[FEATURE REQUEST] Add a way to differentiate headers from application headers


fmvilas:
**Is your feature request related to a problem? Please describe.**
When mapping message header definitions to protocols such as JMS or AMQP 1.0, it's important to differentiate headers from application headers. AMQP defines them as properties and applicationProperties respectively. JMS defines them as headers and properties respectively.

Therefore, we should have a way to label a header as an `applicationHeader`.

**Can't it be tackled using specification extensions?**
Yes, but this way we standardize the behavior.

**Describe the solution you'd like**
Don't have a solution yet. Will make a proposal soon.

**Describe alternatives you've considered**
Making it a specification extension.

fmvilas:
Probably the best option is to remove headers from the spec and leave it for the protocol mappers. In any case, the protocol mappers must support this feature.

SensibleWood:
Working on this issue

SensibleWood:
Merged #169 into master so closing this guy down (not sure why it didn't auto close, anyhoo...)

Issue #42, opened 2017-10-26
Consideration: Protocol specific message objects


gmr:
AMQP, STOMP, MQTT, and others all have variations in the message structure.

I wanted to suggest that you consider the idea of protocol specific message objects *or* protocol specific message object properties, that embrace the individuality and uniqueness of each protocol's message specification.

In addition, the variations between AMQP 0.9 and 1.0 are enough to warrant protocol version disambiguation.

Have you previously put any thought into this? As someone who's a fairly invested in AMQP 0.9, I'd love to see more detail in that area than what the spec currently provides.

fmvilas:
That sounds like an excellent idea. Probably, we could accomplish it by extending messages' fields. Example:

```yaml
lightMeasured:
  summary: Inform about environmental lighting conditions for a particular streetlight.
  payload:
    type: object
    properties:
      lumens:
        type: integer
        minimum: 0
        description: Light intensity measured in lumens.
  amqp:
    version: "0.9.1"
    type: object
    properties:
      somethingUniqueForAMQP: something
```

There are many (and better) ways to do that but I'm leaving this here as a PoC. Is this what you mean?

fmvilas:
@gmr ?

fmvilas:
This will be addressed in v2 as it's a breaking change.

gmr:
Sorry, was heads down @ work. Looks great.

fmvilas:
Cool. I think a better solution can be thought so feel free to make a proposal if you want. Thanks!

fmvilas:
This issue sounds like a candidate for many [specification extensions](https://github.com/asyncapi/asyncapi/issues/138) (one for each protocol).

PR #169, opened 2019-03-07
Issue #112 change scope of Message Object headers to application headers only

SensibleWood:
Summary of changes:

* Changed the `headers` property from a Schema Object to a Map of Schema or Reference Objects
* Changed wording on `headers` property to constrain it to applications headers i.e. it does not include protocol headers, which will be defined by #42 
* Updated examples of headers throughout
* Changed JSON Schema and used `patternProperties` to define the map and constraints on types.
* Provided an example for testing.

### DIFF
--- a/examples/next/application-headers.yml
+++ b/examples/next/application-headers.yml
@@ -1,9 +1,9 @@
 asyncapi: '2.0.0'
 id: 'urn:com:smartylighting:streetlights:server'
 info:
-  title: Correlation ID Example
+  title: Application Headers example
   version: '1.0.0'
-  description: A cut of the Streetlights API to test Correlation ID
+  description: A cut of the Streetlights API to test application header changes supporting #112
   license:
     name: Apache 2.0
     url: https://www.apache.org/licenses/LICENSE-2.0
--- a/examples/next/application-headers.yml
+++ b/examples/next/application-headers.yml
@@ -51,8 +51,8 @@ components:
               minLength: 24
               maxLength: 24
               format: binary
-        replyTo:
-          $ref: "#/components/schemas/replyTo"
+        applicationInstanceId:
+          $ref: "#/components/schemas/applicationInstanceId"
       payload:
         $ref: "#/components/schemas/lightMeasuredPayload"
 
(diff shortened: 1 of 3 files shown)


examples/next/application-headers.yml
```diff
@@ -0,0 +1,82 @@
...
+      headers:
+        MQMD:
+          type: object
+          properties:
+            CorrelId:
+              type: string
+              minLength: 24
+              maxLength: 24
+              format: binary
+        replyTo:
```
fmvilas: Isn't replyTo a protocol header? Also, the only server defined above is MQTT and you're defining WebSphere headers. It may deserve a different example document.

versions/next/asyncapi.md
```diff
@@ -618,14 +618,11 @@ This object can be extended with [Specification Extensions](#specificationExtens
...
     "headers": {
-      "type": "object",
-      "properties": {
-        "qos": {
-          "$ref": "#/components/schemas/MQTTQoSHeader"
-        },
-        "retainFlag": {
-          "$ref": "#/components/schemas/MQTTRetainHeader"
-        }
+      "qos": {
```
fmvilas: QoS and retainFlag are MQTT headers. We should get rid of them here.

versions/next/schema.json
```diff
@@ -742,7 +742,16 @@
               "type": "string"
             },
             "headers": {
-              "$ref": "#/definitions/schema"
+              "type": "object",
+              "additionalProperties": false,
+              "patternProperties": {
```
fmvilas: You can just use `additionalProperties: { oneOf...` instead of patternProperties.

versions/next/asyncapi.md
SensibleWood: Sure - will change and resubmit

examples/next/application-headers.yml
SensibleWood: Not universally but as this is an example will change and come up with a better example

versions/next/schema.json
SensibleWood: OK I'll change

SensibleWood:
Changes based on feedback submitted

Issue #108, opened 2019-02-03
[FEATURE REQUEST] RAML-like traits


fmvilas:
## Is your feature request related to a problem? Please describe.
I want to be able to better reuse parts of an AsyncAPI document. As an example, if all my messages have 4 headers in common, I'd like to be able to group them using a name and express it in each message.

## Can't it be tackled using specification extensions?
No.

## Describe the solution you'd like
The idea is to implement something similar to what RAML did with traits. Traits are pieces of information that will be merged into the referencing object. Here are some rules:

1. Traits will be merged using the [JSON Merge Patch](https://tools.ietf.org/html/rfc7386) algorithm.
2. The object that uses traits MUST be valid before applying them.
3. The resulting object after applying the traits MUST be valid too.
4. Traits can only be applied to [operation](https://github.com/asyncapi/asyncapi/blob/master/versions/next/asyncapi.md#operationObject) and [message](https://github.com/asyncapi/asyncapi/blob/master/versions/next/asyncapi.md#messageObject) objects.
5. Message traits MUST NOT contain the `payload` property. Operation traits MUST NOT contain the `message` property.
6. Traits MUST be applied in the order they're defined.
7. Traits MUST NOT contain traits.
8. It MUST support variables, allowing for simple variable replacement. Variables MUST be enclosed between `{{` and `}}`.
9. Where used, the items in the array of traits MUST be either a trait definition or a 2-elements array where the first element is the trait definition and the second is a list of variable values. All variables MUST be fulfilled.

#### Example

```yaml
channels:
  my/channel:
    publish:
      traits:
        - [{$ref: '#/components/traits/docs'}, {headerId: 'my-channel-publish'}]
      message:
        traits:
          - $ref: '#/components/traits/jsonOverProto3'
          - $ref: '#/components/traits/kafka'
components:
  traits:
    docs:
      externalDocs:
        url: https://mycompany.com/docs#{{headerId}}
    jsonOverProto3:
      schemaFormat: 'application/vnd.google.protobuf;version=3'
      contentType: application/json
    kafka:
      protocolInfo:
        kafka:
          compression: snappy
```

MikeRalphson:
https://gist.github.com/MikeRalphson/7ee0624fc1df70c0475717d89a5b6e4b

fmvilas:
This is exactly how I was imagining it 👍

fmvilas:
Just updated the description with the proposal.

fmvilas:
[...]

Issue #107, opened 2019-02-03
[FEATURE REQUEST] RAML-like overlays


fmvilas:
**Is your feature request related to a problem? Please describe.**
I want to be able to add and/or override parts of the content of an AsyncAPI document, for instance, being able to provide the same documentation in different languages.

**Can't it be tackled using specification extensions?**
No.

**Describe the solution you'd like**
Don't have a solution yet. Will make a proposal soon.

**Describe alternatives you've considered**
Using specification extensions, but it's overly complicated.

MikeRalphson:
See also https://github.com/OAI/OpenAPI-Specification/issues/1722

mattbishop:
OpenAPI has an overlay spec they are working on here: https://github.com/OAI/OpenAPI-Specification/blob/master/proposals/004_Overlays.md

fmvilas:
Thanks, Matt. @darrelmiller, is it going to finally make it to 3.1.0?

fmvilas:
I think that's already in v3, right? We changed the merging algorithm.

PR #170, opened 2019-03-10
Add support for Protocol Info objects

### LINKED ISSUE #41: AMQP Message Exchange
I noticed there is no construct for an exchange in AMQP. It would be nice to have that explicitly called out somehow.

### LINKED ISSUE #42: Consideration: Protocol specific message objects

### LINKED ISSUE #164: [FEATURE REQUEST] Message-classes/Topic-classes
**Is your feature request related to a problem? Please describe.**
I believe it would be useful to allow protocol specific settings for topics or messages, much like the Kubernetes spec does for persistent volumes in the form of Storage-classes.

In this case I would like to request a feature for publishing messages directly to a queue for the AMQP protocol. As well as the possibility to add a default queue and exchange. Allowing this would cover many desirable use-cases.

**Can't it be tackled using specification extensions?**
I believe that if you support a protocol you should support all its features. So in this case it would not suffice.

**Describe the solution you'd like**
I would suggest something similar to storage-classes in Kubernetes. Think of it like a message-class/topic-class. Adding a layer of abstraction so the topic-class can be reused.

Below I've added an example for topic-classes something similar could be added for messages if desirable. In it I've added a `topic-classes` attribute to the components property. It contains a `default-class` which defines a `protocol` which to me is akin to a `provisioner` in Kubernetes storage-classes. The class states it is the default-class for all topics however, topics can override it by specifying their preferred class.
The `default-class` specifies a `queue` which applies to both subscribing and publishing, unless the queue is overridden in the `subscribe` or `publish` properties of the class. If the properties do not specify a queue the default is used.

```
asyncapi: '1.3.0'
info:
  title: Example API
  version: '1.0.0'
  description: |
    Example API.
  license:
    name: Apache 2.0
    url: 'https://www.apache.org/licenses/LICENSE-2.0'

servers:
- url: some.url.net
  scheme: amqp
  description: Example broker
  variables:
    port:
      description: Secure connection (TLS) is available through port 8883.
      default: '1883'
      enum:
      - '1883'
      - '8883'

topics:
  example-topic:
    class: 
      $ref: '#/components/topic-classes/default-class'
    publish:
      $ref: '#/components/messages/example'
  example-message-class-topic:
    publish:
      $ref: '#/components/messages/message-class-example'

components:
  messages:
    example:
      summary: An example message.
      payload:
        $ref: "#/components/schemas/example"
    message-class-example:
      summary: An example message.
      class: '#/components/message-classes/rpc-class'
        $ref: 
      payload:
        $ref: "#/components/schemas/example"
  schemas:
    example:
      type: object
      properties:
        example:
          type: sting
          description: An example string
  topic-classes:
    default-class: 
      scheme: amqp
      default: true
      queue:
        name: example-queue
      publish: 
        direct-to-queue: true
        routing-key: routed-to-something
#      exchange: <- Defining exchange specifics would also be possible, just not when publishing directly to a queue.
#        name: overriding-exchange 
#        type: fanout, direct, topic
      subscribe:
        binding-key: routed-to-something-else
        queue:
          name: overriding-queue
  message-classes:
    rpc-class:
      scheme: ampq
      delivery_mode: 2
      content_type: application/json
      reply_to: rpc-queue
```

**Describe alternatives you've considered**
I've considered adding extensions which tools could read. Though I feel it would not be a proper solution due to the reasons I've stated above. If a specification supports a protocol it should support the protocols features.

**Additional context**
Some benefits of topic-classes:

1. Adding topic-classes/message-classes opens up the specification to many more protocol-specific use-cases.
2. Topic-classes would not be a breaking change, it would simply add functionality as new properties. 
3. I believe this proposal would tie into https://github.com/asyncapi/asyncapi/issues/42 as it has a few commonalities.
4. Topic-classes/message-classes can be reused across multiple topics/messages.
5. Message-classes could be used to implement features like [RPC in AMQP](https://www.rabbitmq.com/tutorials/tutorial-six-javascript.html). I believe this is similar to the feature-request in https://github.com/asyncapi/asyncapi/issues/94 not certain though.

Let me know what you think.

fmvilas:
### What?
Fixes #41
Fixes #42
Fixes #164

### Proposal
ProtocolInfo objects are opaque objects that may be used by tools to perform/add protocol-specific operations/features. Although the object is opaque from a specification point of view, we'll maintain a catalog of protocolInfo objects, so they are standardized across tools but don't compromise the specification. That doesn't stop anyone to use custom protocolInfo objects, at the expense of creating the tooling to support them.

ProtocolInfo objects can be attached to:
* Channels: Describes protocol-specific information about the channel, e.g., queue name, exchange name or partition number.
* Operations: Describes protocol-specific information about the operation. Example: `noAck` (AMQP).
* Messages: Describes protocol-specific information about the message. Examples: protocol headers/properties like `QoS` (MQTT), `delivery_mode` (AMQP), message `key` (Kafka), etc.

Reusing protocolInfo objects will be possible through traits (#108) and overlays (#107).

### Missing things

- [x] ~Add to Markdown version of the specification.~
- [x] ~Add to JSON Schema version of the specification.~


fmvilas:
@RobertDiebels @gmr ☝️ early draft of the new feature to support protocol-specific information.

examples/next/rpc-client.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: queue
+        schema:
+          type: string
+          pattern: '^amq\\.gen\\-.+$'
+    protocolObjects:
+      amqp-0-9-1:
+        channelIsQueue: true
+        queue:
+          randomName: true
+          exclusive: true
```
fmvilas: Please, ignore the "quality" of the `amqp-0-9-1` object here. The definition of a specific protocol object must be done separately. This example is just to illustrate the feature.

examples/next/rpc-client.yml
```diff
@@ -0,0 +1,73 @@
...
+    subscribe:
+      operationId: receiveSumResult
+      correlationId:
+        location: $message.header#/correlation_id
+      protocolObjects:
+        amqp-0-9-1:
+          noAck: true
+          properties:
+            correlation_id:
+              type: string
```
fmvilas: Please, ignore the "quality" of the `amqp-0-9-1` object here. The definition of a specific protocol object must be done separately. This example is just to illustrate the feature.

RobertDiebels:
> 
> 
> @RobertDiebels @gmr ☝️ early draft of the new feature to support protocol-specific information.

+1 for the traits compatibility. Re-using protocol-info seems super useful.

fmvilas:
BTW, there's already a proposal for traits: https://github.com/asyncapi/asyncapi/issues/108.

PR #173, opened 2019-03-11
Add multiLevel field to parameters

### LINKED ISSUE #171: [FEATURE REQUEST] Annotate a channel parameter as multi-level
## Is your feature request related to a problem? Please describe.
I'd like to specify that a channel parameter must span across multiple levels. Like `#` in AMQP and MQTT or `>` in NATS. In case the underlying protocol doesn't support it, it MUST be ignored.

## Can't it be tackled using specification extensions?
Yes, but looks interesting enough to make it part of the spec.

## Describe the solution you'd like
Add a property called `multiLevel` to the [Parameter Object](https://github.com/asyncapi/asyncapi/blob/master/versions/next/asyncapi.md#parameterObject). Its type MUST be boolean and the default value is `false`.

## Additional context
https://www.hivemq.com/blog/mqtt-essentials-part-5-mqtt-topics-best-practices/#wildcards

https://nats.io/documentation/internals/nats-protocol/

fmvilas:
### What?
Fixes #171

### DIFF
--- a/versions/next/asyncapi.md
+++ b/versions/next/asyncapi.md
@@ -672,6 +672,7 @@ Field Name | Type | Description
 ---|:---:|---
 <a name="parameterObjectName"></a>name | `string` | The name of the parameter.
 <a name="parameterObjectDescription"></a>description | `string` | A verbose explanation of the parameter. [CommonMark syntax](http://spec.commonmark.org/) can be used for rich text representation.
+<a name="parameterObjectMultiLevel"></a>multiLevel | `boolean` | Whether a parameter spans across multiple levels or not. Defaults to `false`.
 <a name="parameterObjectSchema"></a>schema | [Schema Object](#schemaObject) | Definition of the parameter.
 
 This object can be extended with [Specification Extensions](#specificationExtensions).
--- a/versions/next/schema.json
+++ b/versions/next/schema.json
@@ -563,11 +563,16 @@
       "properties": {
         "description": {
           "type": "string",
-          "description": "A brief description of the parameter. This could contain examples of use.  GitHub Flavored Markdown is allowed."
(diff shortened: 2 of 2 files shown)


versions/next/asyncapi.md
```diff
@@ -672,6 +672,7 @@ Field Name | Type | Description
 ---|:---:|---
 <a name="parameterObjectName"></a>name | `string` | The name of the parameter.
 <a name="parameterObjectDescription"></a>description | `string` | A verbose explanation of the parameter. [CommonMark syntax](http://spec.commonmark.org/) can be used for rich text representation.
+<a name="parameterObjectMultiLevel"></a>multiLevel | `boolean` | Whether a parameter spans across multiple levels or not. Defaults to `false`.
```
MikeRalphson: This appears to be the only relevant use of the word level(s) in the spec. I.e. it is undefined. Can we be clearer as to a) what kind of parameters can be defined as `multiLevel` - i.e. is it only parameters that act as wildcards and b) what a 'level' is in this context?

fmvilas: You're right, it should be clearer. To answer your points:

A) Yes, only parameters can act as wildcards, and this property is just for wildcard parameters.

B) Check out this article for context: https://www.hivemq.com/blog/mqtt-essentials-part-5-mqtt-topics-best-practices/. We should come up with a good explanation here.

Thanks!

fmvilas:
Actually, regarding A, all parameters are wildcards.

versions/next/asyncapi.md
fmvilas: Actually, regarding A, all parameters are wildcards.

fmvilas: @MikeRalphson can you have a look at the PR fixing this? https://github.com/asyncapi/asyncapi/pull/187. I'm not happy with my explanation, to be honest 😅

PR #187, opened 2019-03-26
Remove parameter multiLevel flag

fmvilas:
### What?
Explains what a level is.

### Why?
As per @MikeRalphson comment: https://github.com/asyncapi/asyncapi/pull/173#discussion_r268209183

### DIFF
--- a/versions/next/asyncapi.md
+++ b/versions/next/asyncapi.md
@@ -757,7 +757,7 @@ Field Name | Type | Description
 ---|:---:|---
 <a name="parameterObjectName"></a>name | `string` | The name of the parameter.
 <a name="parameterObjectDescription"></a>description | `string` | A verbose explanation of the parameter. [CommonMark syntax](http://spec.commonmark.org/) can be used for rich text representation.
-<a name="parameterObjectMultiLevel"></a>multiLevel | `boolean` | Whether a parameter spans across multiple levels or not. Defaults to `false`.
+<a name="parameterObjectMultiLevel"></a>multiLevel | `boolean` | Whether a parameter spans across multiple levels or not. Levels are the pieces of text delimited by slashes (`/`). Defaults to `false`.
 <a name="parameterObjectSchema"></a>schema | [Schema Object](#schemaObject) | Definition of the parameter.
 
 This object can be extended with [Specification Extensions](#specificationExtensions).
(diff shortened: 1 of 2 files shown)


versions/next/asyncapi.md
```diff
@@ -757,7 +757,7 @@ Field Name | Type | Description
 ---|:---:|---
 <a name="parameterObjectName"></a>name | `string` | The name of the parameter.
 <a name="parameterObjectDescription"></a>description | `string` | A verbose explanation of the parameter. [CommonMark syntax](http://spec.commonmark.org/) can be used for rich text representation.
-<a name="parameterObjectMultiLevel"></a>multiLevel | `boolean` | Whether a parameter spans across multiple levels or not. Defaults to `false`.
+<a name="parameterObjectMultiLevel"></a>multiLevel | `boolean` | Whether a parameter spans across multiple levels or not. Levels are the pieces of text delimited by slashes (`/`). Defaults to `false`.
```
MikeRalphson: While this is obviously better, I still feel it's a little thin on the possible restrictions and effects. Say a parameter was a simple `string` with an `enum` of `userId` / `orgId`, if we tagged that with `multiLevel: true` is that an error as the allowable values of the parameter include neither a slash nor a wildcard character?

fmvilas: Mmmm now I see how this is creating confusion. The fact that it's multi-level doesn't mean that it must contain a wildcard character. Actually, this flag (`multiLevel`) is, precisely, "the wildcard". This is just an indicator for code generators and SDKs that you might want to use, e.g., a `#` wildcard. But now I'm thinking if it would be better suited as an extension or as a `protocolInfo` object.

Example as it is right now:

```yaml
channels:
  /flights/{destination}:
    parameters:
      - name: destination
        examples:
          - 'es/bcn'
          - 'uk/lgw'
        multiLevel: true
```

Example as a protocolInfo object:

```yaml
channels:
  /flights/{destination}:
    parameters:
      - name: destination
        examples:
          - 'es/bcn'
          - 'uk/lgw'
        protocolInfo:
          mqtt:
            wildcard: '#'
```

I see as a group of parameters, for instance:

```
         |---- destination ----|
/flights/{country}/{airportCode}
```

fmvilas: But that makes you lose information. Mmmmm I'm not convinced of what I did here. I think I'll better leave it out for now until we have a better reason to add it. Thanks for pointing out mate!

Issue #192, opened 2019-03-31
Review and expand introduction copy for each object section in v2 spec


jhigginbotham:
Example 1: Operation object lacks clear description/definition of its purpose, why it exists. Current description: "Operation Object - Describes a publish or a subscribe operation."

Example 2: Servers Object section description is lacking. Currently "Server Object - An object representing a Server.", however, it should discuss that this allows capturing runtime details about one or more brokers for each protocol, with support for variable substitution in generating the host based on environment, tenant name, etc. 

Recommend revisiting the introduction paragraph of each section the spec, expanding those lacking clarity to better inform the reader about the purpose of the section relative to the overall definition.

PR #202, opened 2019-04-19
Fixed issue 193 - broken links in the TOC of the spec.

### LINKED ISSUE #193: v2 spec TOC link points to the wrong location
Servers Object link in table of contents (under Schema heading) points to the wrong location. It should point to the start of the Servers section. Anchor for TOC link points to `#A2SServers` instead of `#serverObject`. 

Recommend revisiting all TOC entries to verify links point to the proper locations

damaru-inc:
Fixes #193


versions/next/asyncapi.md
```diff
@@ -47,7 +47,7 @@ Means that the [application](#definitionsApplication) is a [consumer](#definitio
       - [Info Object](#infoObject)
       - [Contact Object](#contactObject)
       - [License Object](#licenseObject)
-      - [Servers Object](#A2SServers)
+      - [Servers Object](#serverObject)
```
fmvilas: This link is not correct. It should link to the "servers object" but it's missing. We should create the Servers Object section in the spec.

damaru-inc: Oh, I see. But wouldn't the "servers object" just be an array of server objects?

fmvilas: Yes, that's why it was linked to `#A2SServers`. Ideally and since every object has its own "space" in the spec, this should too. It would be a matter of adding another section saying it's just an array of Server Objects.

What do you think?

fmvilas:
By the way, if you type "Fixes #NN" on the description of the pull request, Github will automatically close the issue when the PR gets merged. See an example: https://github.com/asyncapi/asyncapi.github.io/pull/15.

versions/next/asyncapi.md
MichaelDavisSolace: Okay, makes sense. I can put that in. I realize that the task to put in the Servers Object is in #192 , I could do that first.

damaru-inc: I got back to working on this. I wrote a tiny section called Servers Object and said it was just an array of Server Objects. I'm having trouble seeing the value in this. The AsyncAPI object has a field named servers which is an array of Server Objects, rather than a Servers Object. Creating that extra bit of doc seems to me to just be adding noise. Maybe I'm misunderstanding something!

fmvilas: Well, it's actually just to be consistent. Think this is a specification, not a documentation page or article. Sometimes you have to write obvious things to make them explicit.

damaru-inc: Okay, I'll carry on then.

fmvilas:
What's the state of this PR, @damaru-inc? Do you think it's done?

MichaelDavisSolace:
I thought it was done when I did that last commit, but that was a while ago. I'd better see if there are other more recent commits that interfere. I'll pull/merge and review it again.

MichaelDavisSolace:
I merged from master, checked everything, I believe this PR is ready.

fmvilas:
Awesome! Thanks for this PR, Michael!

PR #242, opened 2019-07-24
Remove baseChannel and clarify channel naming constraints

### LINKED ISSUE #215: [FEATURE REQUEST] Add Parameter "in" property from OpenAPI
Following the Specification for the parameter object.
_"It SHOULD be present only when using channels with expressions (as defined by RFC 6570 section 2.2)."_

The RFC 6570 section 2.2 defines query-style as &name=value pairs within a literal query component.

So Channels following URI pattern and query-styles being defined the parameters should accept **"in": path** and **"in": query**.

for that the schema for Parameter Object should accept the "in" property, going close towards the OpenAPI specification.

### LINKED ISSUE #237: [2.0.0 REVIEW] Clarification on Channel naming
The [Channels](https://www.asyncapi.com/docs/specifications/2.0.0-rc1/#a-name-channelsobject-a-channels-object) object mentioned that Channel keys must start with a "/" or be relative to the server's `baseChannel` - This may be a bit ambiguous in situations where the `baseChannel` is not provided (defined as defaulting to "empty string"). I would thus assume the key is relative to "/" - effectively, treating it as an absolute path. Perhaps this can be clarified in this description block?

Additionally, a total nit-pick... Consider changing "base channel" to "baseChannel," perhaps? It makes searching of the page easier to find all references.

fmvilas:
Fixes #237.
Fixes #215.

### DIFF
--- a/versions/next/asyncapi.md
+++ b/versions/next/asyncapi.md
@@ -314,7 +314,6 @@ Field Name | Type | Description
 <a name="serverObjectProtocolVersion"></a>protocolVersion | `string` | The version of the protocol used for connection. For instance: AMQP `0.9.1`, HTTP `2.0`, Kafka `1.0.0`, etc.
 <a name="serverObjectDescription"></a>description | `string` | An optional string describing the host designated by the URL. [CommonMark syntax](http://spec.commonmark.org/) MAY be used for rich text representation.
 <a name="serverObjectVariables"></a>variables | Map[`string`, [Server Variable Object](#serverVariableObject)] | A map between a variable name and its value.  The value is used for substitution in the server's URL template.
-<a name="serverObjectBaseChannel"></a>baseChannel | `string` | A string describing the base channel. MUST be in the form of a [RFC 3986 URI path](https://tools.ietf.org/html/rfc3986#section-3.3). URI templates are not allowed here. Defaults to empty string.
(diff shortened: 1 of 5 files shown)


versions/next/asyncapi.md
```diff
@@ -479,14 +476,15 @@ defaultContentType: application/json
 
 #### <a name="channelsObject"></a>Channels Object
 
-Holds the relative paths to the individual channel and their operations.
-Channels starting with a slash (`/`) MUST be treated as absolute paths and therefore skipping the [`server's base channel`](#serverObjectBaseChannel). Otherwise, the final channel MUST be resolved using the [`server's base channel`](#serverObjectBaseChannel) and the current path.
+Holds the relative paths to the individual channel and their operations. Channel paths are relative to servers.
+
+Channels are also known as "topics", "routing keys", "event types" or "paths".
```
WaleedAshraf: good to have this explanation 👍

WaleedAshraf:
Where do update the actual schema definition w.r.t to these changes?

WaleedAshraf:
I think we need to remove this also: https://github.com/asyncapi/asyncapi/blob/master/versions/next/schema.json#L193

fmvilas:
You're right. It's missing schema changes. Thanks for pointing out.

WaleedAshraf:
LGTM 👍

PR #243, opened 2019-07-24
Simplify traits

fmvilas:
* Simplify traits by not allowing variable replacement.
* Separate traits in `messageTraits` and `operationTraits` in components as they are different object types.


fmvilas:

PR #245, opened 2019-07-25
Use maps instead of arrays when possible

fmvilas:
Changes:

* Makes parameters a map instead of an array. This way we always make sure a name is provided.
* Makes servers a map instead of an array. It gives us the advantage of assigning names to each server, e.g. production, development, staging, etc.

The reason behind this change is to facilitate the #107 and #108 features. These features make use of JSON Merge Patch and arrays are particularly "problematic" when trying to merge 2 or more. In the case of JSON Merge Patch, 2 arrays can't be merged and instead the second replaces the first one. If we remove unnecessary arrays from the spec, it will be possible to merge these objects too.

### DIFF
--- a/examples/next/application-headers.yml
+++ b/examples/next/application-headers.yml
@@ -8,7 +8,8 @@ info:
     url: https://www.apache.org/licenses/LICENSE-2.0
 
 servers:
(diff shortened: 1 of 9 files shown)


fmvilas:

PR #256, opened 2019-08-08
Make Schema a superset of JSON Schema Draft 07

### LINKED ISSUE #212: [2.0.0 REVIEW] Use JSON Schema Draft 07 as the default schema
After playing a lot these days with AsyncAPI 2.0.0-rc1, I realized most of the pain comes from having the OpenAPI schema as the default. We want to keep supporting OpenAPI schemas but there's no need to make it the default choice.

There's only one thing that worries me and it's the lack of a deprecation property in Draft 07. However, @philsturgeon is working on it and will probably make it for Draft 08. As per `discriminator` and `externalDocs`, they can just be extensions in the form of `x-discriminator` and `x-external-docs`. Same for `deprecated` in the meantime, it can be done using `x-deprecated`.

@handrews do you have a list of changes (or potential changes) for Draft 08? I've seen [this](https://github.com/json-schema-org/json-schema-spec/milestone/6) but I'd like to know if upgrading the AsyncAPI default schema from Draft 07 to Draft 08 will be a breaking change. Or put it another way, are there breaking changes in Draft 08 compared to Draft 07?

fmvilas:
Fixes #212 

@philsturgeon would you mind having a look at this PR? I took yours and changed/added some stuff.

Clarifications:
* XML Object and `xml` property have been removed. Future versions of AsyncAPI will support XSD, which is better suited for XML definitions.
* We're not moving to Draft-08 yet because of the lack of tooling and, IIRC, Draft-08 is still not released. However, we're doing steps in this direction by adding `deprecated` to our superset. Once Draft-08 is out and tooling implemented it, we can get rid of this property.
* We still support OpenAPI schemas on message payloads. Just indicate `schemaFormat` is `application/vnd.oai.openapi;version=3.0.0`. Actually, [once OpenAPI gets rid of their current Schema object](https://github.com/OAI/OpenAPI-Specification/pull/1977) and evolves into a Draft-08 superset, we will be able to align again.

### DIFF
--- a/versions/next/asyncapi.md
+++ b/versions/next/asyncapi.md
@@ -59,7 +59,6 @@ Means that the [application](#definitionsApplication) is a [consumer](#definitio
       - [Components Object](#componentsObject)
       - [Reference Object](#referenceObject)
       - [Schema Object](#schemaObject)
-      - [XML Object](#xmlObject)
       - [Security Scheme Object](#securitySchemeObject)
       - [Parameter Object](#parameterObject)
       - [Protocol Info Object](#protocolInfoObject)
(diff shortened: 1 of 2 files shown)


versions/next/asyncapi.md
```diff
@@ -1405,40 +1406,39 @@ The following properties are taken directly from the JSON Schema definition and
...
 The following properties are taken from the JSON Schema definition but their definitions were adjusted to the AsyncAPI Specification.
 
-- type - Value MUST be a string. Multiple types via an array are not supported.
+- properties - Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema.
+- patternProperties - Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema.
+- additionalProperties - Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema.
+- additionalItems - Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema.
+- items - Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema.
+- propertyNames - Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema.
+- contains - Inline or referenced schema MUST be of a [Schema Object](#schemaObject) and not a standard JSON Schema.
```
philsturgeon: should `not` be on here?

WaleedAshraf:
the payload can still be XML serializable?

 https://github.com/asyncapi/asyncapi/pull/256/files#diff-c579311543db4dde2bd2e2ca895ffed1R76

fmvilas:
Oh yeah! The payload can be serialized to anything. We're just removing a feature (originally from OpenAPI schemas) that allows you to define XML using JSON Schema-like syntax. I'm doing this on purpose to keep it simple. If people request this feature in the future we can consider adding it. Adding is easy and quick but removing it later would be a mess and would require a major version bump.

It's on the roadmap for the next minor version to add support for non-JSON schema languages: XSD, Protobuf, etc. XSD is what most people using XML use for schema contracts.

fmvilas:
Also, people can still use OpenAPI schemas if they need it. What's being changed here is the **default** schema format.

versions/next/asyncapi.md
fmvilas: It's below. It was already there, that's why it's not showing the change.

fmvilas:

fmvilas:
Thanks, @philsturgeon, and @WaleedAshraf for reviewing it.

PR #263, opened 2019-09-09
Fix operation verbs semantics and definitions

fmvilas:
* Changes the meaning of the publish/subscribe verbs as it was in v1.
* Removes the definition of message broker and replaces any reference to it with the word "server" so we don't confuse people using AsyncAPI for non-broker-based topologies.

### DIFF
--- a/examples/next/streetlights.yml
+++ b/examples/next/streetlights.yml
@@ -43,8 +43,8 @@ channels:
     parameters:
       streetlightId:
         $ref: '#/components/parameters/streetlightId'
-    subscribe:
-      summary: Receive information about environmental lighting conditions of a particular streetlight.
+    publish:
+      summary: Inform about environmental lighting conditions for a particular streetlight.
       operationId: receiveLightMeasurement
       traits:
         - $ref: '#/components/operationTraits/kafka'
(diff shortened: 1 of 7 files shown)


examples/next/streetlights.yml
```diff
@@ -55,7 +55,7 @@ channels:
     parameters:
       streetlightId:
         $ref: '#/components/parameters/streetlightId'
-    publish:
+    subscribe:
```
treeder: These are `commands` to turn on, turn off and dim the lights ya?  In messages below this it says "Command a particular streetlight to turn the lights on or off." So I Believe these should be `publish`, not `subscribe`, since the client would be publishing those commands to the service. Maybe these were already the new way (or the 1.0 way?).

fmvilas: There's a mistake in the description of the message. It's because, in the beginning, this example was representing a specific streetlight API. Right now it's representing a system that can deal with streetlights. Like a control panel or something similar.

Thanks for pointing out. I'll make sure every description gets nailed in this document.

fmvilas: Just realized it's correct. This example is describing a "control panel" so `subscribe` is correct.

fmvilas: See the description of the API:

The Smartylighting Streetlights API allows you to remotely manage the city lights.

### Check out its awesome features:

  * Turn a specific streetlight on/off 🌃
  * Dim a specific streetlight 😎
  * Receive real-time information about environmental lighting conditions 📈

PR #268, opened 2019-09-12
added type boolean for additionalProperties in schema

WaleedAshraf:
closes #267

PR #269, opened 2019-09-17
add a test for the fix

maquessime:
not sure if the strategy is one `.yml` per test case

Depends on #268

### DIFF
--- a/examples/2.0.0/anyof.yml
+++ b/examples/2.0.0/anyof.yml
@@ -23,6 +23,7 @@ components:
       properties:
         key:
           type: string
+          additionalProperties: false
     objectWithKey2:
       type: object
       properties:
--- a/versions/2.0.0/schema.json
+++ b/versions/2.0.0/schema.json
@@ -366,7 +366,15 @@
           },
           "properties": {
             "additionalProperties": {
-              "$ref": "#/definitions/schema"
+              "anyOf": [
+                {
+                  "$ref": "#/definitions/schema"
+                },
+                {
+                  "type": "boolean"
+                }
+              ],
+              "default": {}
             },
             "items": {
               "anyOf": [


WaleedAshraf:
Nice.
So, if we base this change on current master, CI would fail. Right?

About adding one .yaml per test.
I think yes, we should have a test suie with different yaml files and cases. But that's a separate task.

maquessime:
> So, if we base this change on current master, CI would fail. Right?
correct - that's why I did it against your fix branch 😄

examples/2.0.0/anyof.yml
```diff
@@ -23,6 +23,7 @@ components:
       properties:
         key:
           type: string
+          additionalProperties: false
```
fmvilas:
```suggestion
      additionalProperties: false
```

fmvilas:
Oh, just saw your reply. Forget my comment then. You were pointing to his branch. Let me point to master so we don't have branches that depend on other branches.

WaleedAshraf:
let me re-run tests, just to confirm.
it should fail on the master.

maquessime:
to double check I did

- `git revert 4032d954766689693d6d5fd34f6fed975170e5d7`
- `npm run test`
- got 
```
Error details:

[ { keyword: 'type',
    dataPath:
     '.components.schemas[\'objectWithKey\'].properties[\'key\'].additionalProperties',
    schemaPath: '#/allOf/1/type',
    params: { type: 'object' },
    message: 'should be object' } ]
```
- `git cherry-pick 4032d954766689693d6d5fd34f6fed975170e5d7`
- tests were passing

WaleedAshraf:
Eh, you have created a branch from mine. Makes sense why they are still passing.

WaleedAshraf:
So, now we only need to merge this. I can close my PR.

maquessime:
Yeah sorry for the confusion!

PR #278, opened 2019-10-06
Add support for the Mercure protocol

dunglas:
Hi,

This PR adds support for Mercure, a protocol for async and real-time HTTP APIs.

More info about the protocol:
* https://datatracker.ietf.org/doc/draft-dunglas-mercure/
* https://mercure.rocks

### DIFF
--- a/examples/2.0.0/mercure.yml
+++ b/examples/2.0.0/mercure.yml
@@ -1,0 +1,51 @@
+asyncapi: '2.0.0'
+info:
+  title: Mercure Hub Example
+  version: '1.0.0'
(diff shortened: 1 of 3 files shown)


fmvilas:
Hey @dunglas! Great to see you're adding the Mercure protocol! 🎉 

I find missing the definition of the binding though, you have to create one here: https://github.com/asyncapi/bindings. Once it's done, we can merge this PR.

Thanks!

dunglas:
And here is the binding description (nothing specific is needed 🎆): https://github.com/asyncapi/bindings/pull/17

I also added an example describing a Mercure hub with AsyncAPI!

PR #288, opened 2019-10-29
Remove free-form from bindings

### LINKED ISSUE #283: Definition of bindings needs clarification
In AsyncAPI 2.0, the server/channel/operation/message bindings can be either "fixed" or custom via `x-` extension. The "fixed" set of bindings are all defined in [asyncapi/bindings](https://github.com/asyncapi/bindings). That part is clear. 

The part that is confusing in the spec is that the description of each type of binding starts with `A free-form map...`. However, if it's "free-form", it cannot by definition be "fixed". It also clashes with some of the "fixed" bindings described in  [asyncapi/bindings](https://github.com/asyncapi/bindings) which disallow the use of certain types of bindings. E.g. for `amqp`, `server` bindings are [disallowed](https://github.com/asyncapi/bindings/tree/master/amqp#server-binding-object) altogether.

fmvilas:
Fixes #283


fmvilas:
FYI @jstoiko

PR #289, opened 2019-10-29
Make Avro and OpenAPI Schemas support recommended instead of mandatory

### LINKED ISSUE #287: Avro is not be mendatory
Currently the [`Schema format table`](https://github.com/asyncapi/asyncapi/blob/master/versions/2.0.0/asyncapi.md#schema-formats-table) section, says that `The following table contains a set of values that every implementation MUST support`. The "MUST" seems to indicate that all formats and values below are mandatory to support. However, I don't think this was what was intended in version 2.0 as only JSON Schema should be mandatory.

fmvilas:
Fixes #287


fmvilas:
FYI @jstoiko

derberg:

Issue #286, opened 2019-10-22
Add example polymorphism that overrides use of schema name as discriminator


gibson042:
[Composition and Inheritance (Polymorphism)](https://www.asyncapi.com/docs/specifications/2.0.0/#schemaComposition) documents the following:
> There are are two ways to define the value of a discriminator for an inheriting instance.
> - Use the schema’s name.
> - Override the schema’s name by overriding the property with a new value. If exists, this takes precedence over the schema’s name.

However, no example is provided for the second option, and it is not quite clear to me how it is intended to work. My best guess is the following, in which the property referenced by `discriminator` in the base schema is constrained in the inheriting schema with `const` (or perhaps with a functionally equivalent single-item `enum`?):
```yaml
schemas:
  Pet:
    type: object
    discriminator: petType
    properties:
      name:
        type: string
      petType:
        type: string
    required:
    - name
    - petType

  Cat:  ## applies to instances with `petType: "cat"`
    description: A representation of a cat
    allOf:
    - $ref: '#/components/schemas/Pet'
    - type: object
      properties:
        petType:
          const: "cat"
        huntingSkill:
          type: string
          description: The measured skill for hunting
          enum:
          - clueless
          - lazy
          - adventurous
          - aggressive
      required:
      - huntingSkill

  Dog:  ## applies to instances with `petType: "dog"`
    description: A representation of a dog
    allOf:
    - $ref: '#/components/schemas/Pet'
    - type: object
      properties:
        petType:
          const: "dog"
        packSize:
          type: integer
          format: int32
          description: the size of the pack the dog is from
          minimum: 0
      required:
      - packSize
```

Something like the above (or a correction, if I have misunderstood) should be added to the documentation.

fmvilas:
Hey @gibson042, sorry for the delay responding. Yes, that's how I understand it too. Adding an example would be great.

PR #293, opened 2019-11-14
Add example polymorphism that overrides use of schema name as discriminator value

gibson042:
Fixes gh-286


gibson042:
> Thanks for taking the time, Richard. I hope you don't have a stick insect as a pet :stuck_out_tongue_winking_eye:

No, but [people definitely do](https://en.wikipedia.org/wiki/Phasmatodea#In_human_culture). It was just surprisingly difficult to come up with a pet whose common name (discriminator field value) would differ from an OpenAPI schema name that made sense alongside Cat and Dog, and I didn't want to completely rewrite the example.

PR #298, opened 2019-12-09
Initial cut at aws-sns publish example

karlwbrown:
Draft attempt at a aws-sns publish example.

Some more work needs to be done, including answering a few questions:
1) How do we manage bindings of topicArn - which is a combination of "server"-type information, and "channel" type information. For now it is hardcoded, and thus somewhat duplicative
2) What should the bindings specify? All (standard) parameters/headers that are passed on a publication request?

### DIFF
--- a/examples/2.0.0/aws-sns-publish.yml
+++ b/examples/2.0.0/aws-sns-publish.yml
@@ -1,0 +1,46 @@
+asyncapi: 2.0.0
+id: 'https://github.com/asyncapi/asyncapi/blob/master/examples/2.0.0/aws-sns-publish.yml'
+info:
+  title: Simple AWS SNS Publish specification
+  description: |
+    Publishes a hello-world message to a particular queue
+  version: '0.1.0'
+servers:
+  test:
+    # This URL doesn't really describe a test server, but it does describe the
+    # top-level URL used to post a message to a given AWS region
+    url: https://sns.us-east-1.amazonaws.com/
+    description: Test environment for message publishing
+    protocol: sns
+    protocolVersion: "2010-03-31" #Can we reuse the "version" from down below?
+    bindings:
+      # Details on topicARNs: https://docs.aws.amazon.com/sns/latest/dg/sns-tutorial-create-topic.html
+      # This is a prefix, not a complete topicARN, but the prefix identifies
+      # the "server" environment (which is really just the account where this topic lives)
+      topicArnPrefix: "arn:aws:sns:us-east-1:123456789012:"
+channels:
+  helloWorld:
+    description: Topic used for hello world messages
+    bindings:
+      sns:
+        # This should be appended to the servers' topicARNPrefix to get the complete topicARN
+        # Can we do this automatically using JSON-Schema trickery somehow?
+        topicArnTopicName: "helloWorld"
+    publish:
+      message:
+        # The payload is equivalent to the AWS "Message" field
+        # Publish documentation: https://docs.aws.amazon.com/sns/latest/api/API_Publish.html
+        payload:
+          type: string
+          pattern: '^hello .+$'
+      bindings:
+        sns:
+          Action: Publish
+          # This is a concatenation of topicArnPrefix + topicArnTopicName from servers/channels above
+          TopicArn: "arn:aws:sns:us-east-1:123456789012:helloWorld"
+          # Details here: https://docs.aws.amazon.com/sns/latest/dg/sns-message-attributes.html
+          MessageAttributes:
+            # This is up to the application - any key/value mapping allowed here
+            messageType: HELLO_WORLD
+          # This is a default for all SNS messages
+          Version: "2010-03-31"
--- a/examples/2.0.0/aws-sns-publish.yml
+++ b/examples/2.0.0/aws-sns-publish.yml
@@ -1,46 +1,80 @@
 asyncapi: 2.0.0
 id: 'https://github.com/asyncapi/asyncapi/blob/master/examples/2.0.0/aws-sns-publish.yml'
 info:
-  title: Simple AWS SNS Publish specification
+  title: 'Simple AWS SNS Publish specification'
   description: |
-    Publishes a hello-world message to a particular queue
+    Specification for a service that publishes various "hello world"
+    type messages to the helloWorld SNS topic
   version: '0.1.0'
 servers:
-  test:
-    # This URL doesn't really describe a test server, but it does describe the
-    # top-level URL used to post a message to a given AWS region
-    url: https://sns.us-east-1.amazonaws.com/
-    description: Test environment for message publishing
-    protocol: sns
-    protocolVersion: "2010-03-31" #Can we reuse the "version" from down below?
+  development:
+    # This URL doesn't really describe a development server, but it does describe the
+    # top-level URL used to subscribe to messages from a given AWS region
+    url: 'https://sns.us-east-1.amazonaws.com/'
+    description: 'Development/test account for message publishing'
+    protocol: 'sns'
+    protocolVersion: '2010-03-31'
     bindings:
-      # Details on topicARNs: https://docs.aws.amazon.com/sns/latest/dg/sns-tutorial-create-topic.html
-      # This is a prefix, not a complete topicARN, but the prefix identifies
-      # the "server" environment (which is really just the account where this topic lives)
-      topicArnPrefix: "arn:aws:sns:us-east-1:123456789012:"
+      # AccountID for this environment
+      awsAccountID: "123456789012"
+  test:
+    url: 'https://sns.us-east-1.amazonaws.com/'
+    description: 'Development/test account for message publishing'
+    protocol: 'sns'
+    protocolVersion: '2010-03-31'
+    bindings:
+      awsAccountID: '123456789012'
+  production:
+    url: 'https://sns.us-east-1.amazonaws.com/'
+    description: 'Production account for message publishing'
+    protocol: 'sns'
+    protocolVersion: '2010-03-31'
+    bindings:
+      awsAccountID: '567891011121'
 channels:
   helloWorld:
-    description: Topic used for hello world messages
+    description: 'Topic used for hello world and similar messages'
     bindings:
-      sns:
-        # This should be appended to the servers' topicARNPrefix to get the complete topicARN
-        # Can we do this automatically using JSON-Schema trickery somehow?
-        topicArnTopicName: "helloWorld"
-    publish:
+      $ref: '#/components/channelBindings/sns/helloWorld'
+    subscribe:
       message:
-        # The payload is equivalent to the AWS "Message" field
-        # Publish documentation: https://docs.aws.amazon.com/sns/latest/api/API_Publish.html
-        payload:
-          type: string
-          pattern: '^hello .+$'
+        oneOf:
+          - $ref: '#/components/messages/helloWorldMessage'
+          - $ref: '#/components/messages/goodNightWorldMessage'
+components:
+  channelBindings:
+    sns:
+      helloWorld:
+        development:
+          topicArn: 'arn:aws:sns:us-east-1:123456789012:helloWorld-dev'
+        test:
+          topicArn: 'arn:aws:sns:us-east-1:123456789012:helloWorld'
+        production:
+          topicArn: 'arn:aws:sns:us-east-1:567891011121:helloWorld'
+  messages:
+    helloWorldMessage:
+      name: 'helloWorld'
+      title: 'Hello World message'
+      summary: 'Message sent to say Hello to the World'
+      payload:
+        type: string
+        const: 'Hello World!'
       bindings:
         sns:
-          Action: Publish
-          # This is a concatenation of topicArnPrefix + topicArnTopicName from servers/channels above
-          TopicArn: "arn:aws:sns:us-east-1:123456789012:helloWorld"
           # Details here: https://docs.aws.amazon.com/sns/latest/dg/sns-message-attributes.html
           MessageAttributes:
             # This is up to the application - any key/value mapping allowed here
-            messageType: HELLO_WORLD
-          # This is a default for all SNS messages
-          Version: "2010-03-31"
+            messageType: 'HELLO_WORLD'
+    goodNightWorldMessage:
+      name: 'goodNightWorld'
+      title: 'Goodnight World message'
+      summary: 'Message sent to say goodnight to the World'
+      payload:
+        type: string
+        const: 'Goodnight World!'
+      bindings:
+        sns:
+          # Details here: https://docs.aws.amazon.com/sns/latest/dg/sns-message-attributes.html
+          MessageAttributes:
+            # This is up to the application - any key/value mapping allowed here
+            messageType: 'GOODNIGHT_WORLD'


examples/2.0.0/aws-sns-publish.yml
```diff
@@ -0,0 +1,46 @@
...
+        # The payload is equivalent to the AWS "Message" field
+        # Publish documentation: https://docs.aws.amazon.com/sns/latest/api/API_Publish.html
+        payload:
+          type: string
+          pattern: '^hello .+$'
+      bindings:
+        sns:
+          Action: Publish
+          # This is a concatenation of topicArnPrefix + topicArnTopicName from servers/channels above
+          TopicArn: "arn:aws:sns:us-east-1:123456789012:helloWorld"
```
jonaslagoni: I think the whole `TopicArn` is redundant, since the template/client should use the `topicArnPrefix` together with this  `TopicArn` variable. Therefore I suggest removing `arn:aws:sns:us-east-1:123456789012:` to it only contain `helloWorld`.

examples/2.0.0/aws-sns-publish.yml
```diff
@@ -0,0 +1,46 @@
...
+    publish:
+      message:
+        # The payload is equivalent to the AWS "Message" field
+        # Publish documentation: https://docs.aws.amazon.com/sns/latest/api/API_Publish.html
+        payload:
+          type: string
+          pattern: '^hello .+$'
+      bindings:
+        sns:
+          Action: Publish
```
jonaslagoni: Does SNS use any other actions then Publish and subscribe? else I find this redundant since the binding is already associated with the standard `publish? operation.

jonaslagoni:
Another comment, even though the example is great, we need to specify the binding for SNS more so then giving an example :smile: Maybe its better to start specifying the bindings [here](https://github.com/asyncapi/bindings)?

karlwbrown:
Yes, SNS allows lots of actions; however I don’t know if we want to
enumerate all of them.

https://docs.aws.amazon.com/IAM/latest/UserGuide/list_amazonsns.html

Normally when using SNS, you would use an SDK which will format the request
properly, and would deal with putting ‘Action=Publish’ in the request.

So I guess I’m not sure what we should put in the bindings - I agree that
it is redundant to specify the ‘Publish’ action here, just not sure if it’s
useful to clarify that is the *only* action being covered by this interface
specification.

On Dec 12, 2019, at 07:28, Jonas Lagoni <notifications@github.com> wrote:

﻿

*@jonaslagoni* commented on this pull request.
------------------------------

In examples/2.0.0/aws-sns-publish.yml
<https://github.com/asyncapi/asyncapi/pull/298#discussion_r357118409>:

> +    description: Topic used for hello world messages
+    bindings:
+      sns:
+        # This should be appended to the servers' topicARNPrefix to
get the complete topicARN
+        # Can we do this automatically using JSON-Schema trickery somehow?
+        topicArnTopicName: "helloWorld"
+    publish:
+      message:
+        # The payload is equivalent to the AWS "Message" field
+        # Publish documentation:
https://docs.aws.amazon.com/sns/latest/api/API_Publish.html
+        payload:
+          type: string
+          pattern: '^hello .+$'
+      bindings:
+        sns:
+          Action: Publish

Does SNS use any other actions then Publish and subscribe? else I find this
redundant since the binding is already associated with the standard
`publish? operation.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/asyncapi/asyncapi/pull/298?email_source=notifications&email_token=ACN2DR4UZH4HS4R2YIUASX3QYIVAVA5CNFSM4JYOIZY2YY3PNVWWK3TUL52HS4DFWFIHK3DMKJSXC5LFON2FEZLWNFSXPKTDN5WW2ZLOORPWSZGOCO62GMY#pullrequestreview-331195187>,
or unsubscribe

jonaslagoni:
I see, yes then the Action variable is justified :+1:

examples/2.0.0/aws-sns-publish.yml
jonaslagoni: Justified since [SNS contains multiple](https://docs.aws.amazon.com/IAM/latest/UserGuide/list_amazonsns.html ) `Actions`.

karlwbrown:
Sure - There are many ways of doing it.
There are pros and cons of different approaches.

The reason I split the topic ARN into two parts was because the TopicARN
differs per ‘server’ (really it’s not a server but an AWS account), and per
topic, so I figured it made sense to link those parts to servers and
channels. (Although to complicate matters, the topicName even for a topic
with the same semantic meaning across environments may be different (for
example, it could have another string of digits appended))

However the only ‘real’ parameter that gets passed to Amazon is the
TopicARN, which is provided at the message level.

Another approach would be to define message bindings which would have the
full TopicARN for each channel, and have an array or map of them for each
environment (eg - for the channel known as ‘helloWorld’, X is the actual
TopicARN for test environments whereas Y is the actual TopicARN for the
production environment.)

Example:

```
MessageBindings:
  sns:
    environments:
      - test:
          Channels:
            - helloWorld:
                TopicArn: "arn:aws:sns:us-east-1:123456789012:helloWorld"
            - goodbyeWorld:
                 TopicArn: "arn:aws:sns:us-east-1:123456789012:goodbyeWorld"
       - production:
            Channels:
              - helloWorld:
                   TopicArn:
"arn:aws:sns:us-east-1:123differentaccount:helloWorldProd"
              - goodbyeWorld:
                    TopicArn:
"arn:aws:sns:us-east-1:123differentaccount:goodbyeWorldProd”
```

Also we could define the above as channel bindings, and then import them by
reference to message bindings? There are so many ways to slice this!

Does JSON schema allow for concetenation of strings? (So for example saying
that one string should be a combination of two others?)

I guess my main question is, how will it be used? Do we want to specify as
bindings *all* of the possible headers that could be passed in a publish
request to SNS? If so we need to put topicARN, MessageAttributes, and even
‘Message’ (which Asyncapi defines as ‘payload’).

Or, are the bindings more intended to be informative - eg explain your a
consumer what they might expect the subscriber to have done, but not
worrying so much about the particulars of the headers - I don’t think
Asyncapi is intended to replace an actual spec (for example, MQTT headers
are something like fixed-width binary format which JSON schema doesn’t
allow to define/validate in that way)

On Dec 12, 2019, at 07:27, Jonas Lagoni <notifications@github.com> wrote:

﻿

*@jonaslagoni* commented on this pull request.
------------------------------

In examples/2.0.0/aws-sns-publish.yml
<https://github.com/asyncapi/asyncapi/pull/298#discussion_r357117812>:

> +      sns:
+        # This should be appended to the servers' topicARNPrefix to
get the complete topicARN
+        # Can we do this automatically using JSON-Schema trickery somehow?
+        topicArnTopicName: "helloWorld"
+    publish:
+      message:
+        # The payload is equivalent to the AWS "Message" field
+        # Publish documentation:
https://docs.aws.amazon.com/sns/latest/api/API_Publish.html
+        payload:
+          type: string
+          pattern: '^hello .+$'
+      bindings:
+        sns:
+          Action: Publish
+          # This is a concatenation of topicArnPrefix +
topicArnTopicName from servers/channels above
+          TopicArn: "arn:aws:sns:us-east-1:123456789012:helloWorld"

I think the whole TopicArn is redundant, since the template/client should
use the topicArnPrefix together with this TopicArn variable. Therefore I
suggest removing arn:aws:sns:us-east-1:123456789012: to it only contain
helloWorld.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/asyncapi/asyncapi/pull/298?email_source=notifications&email_token=ACN2DRZG3R2UKM6MSGHS5SLQYIU3DA5CNFSM4JYOIZY2YY3PNVWWK3TUL52HS4DFWFIHK3DMKJSXC5LFON2FEZLWNFSXPKTDN5WW2ZLOORPWSZGOCO62AQY#pullrequestreview-331194435>,
or unsubscribe

karlwbrown:
IMO it feels easier to me to start with a reasonable/simple example, and
then we can work backwards from that to the bindings spec.

On Dec 12, 2019, at 07:33, Jonas Lagoni <notifications@github.com> wrote:

﻿

Another comment, even though the example is great, we need to specify the
binding for SNS more so then giving an example 😄 Maybe its better to start
specifying the bindings here <https://github.com/asyncapi/bindings>?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/asyncapi/asyncapi/pull/298?email_source=notifications&email_token=ACN2DRYYUKSAK7DPQGIZL63QYIVQ3A5CNFSM4JYOIZY2YY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOEGWQTXI#issuecomment-564988381>,
or unsubscribe

jonaslagoni:
Ahh okay, I am starting to understand the `TopicArn` now, I think :smile: I dont think this should be included in the bindings that specific, but rather in the implementation of the template :thinking: I would maybe rename `TopicArn` to `TopicArnPostfix` so it only contain `helloWorld`. That way in the template you can always concat it like so:
```nunjucks
// Only example, not complete nunjucks code.
{{server.bindings().sns().TopicArnPrefix() |  parseToARNString }}:currentUser.uid():{{operation.bindings().sns().TopicArnPostfix() }}
```
Maybe there are even more use-cases etc, but I think it all comes down to the implementation part of the template.

Therefor I  would suggest that you try to start implementing a template using your current bindings, this is the same way I am doing with the NATS bindings. It helps you to get an understanding on how you can actually use it in practice :+1: 

And to answer your question about concatenation of strings in JSON schema then I don't think they have it no :slightly_smiling_face:

karlwbrown:
@jonaslagoni can you explain more what you mean by that nunjucks code?

Also you said this: 
"I dont think this should be included in the bindings that specific, but rather in the implementation of the template "

Can you explain what you mean? I want to be sure we're talking about the same thing.

For me, there is the 'bindings spec' which is here:
https://github.com/asyncapi/bindings/blob/master/sns/README.md#channel

then there is the schema which you generate for your messages, which might be something like this:
https://github.com/asyncapi/asyncapi/pull/298/files#diff-1c4aa514c49eea36bab76f18f17e471f

jonaslagoni:
What I mean by nunjucks code is when used with the [async api generator](https://github.com/asyncapi/generator) you can template your way out different kind of situations like I suggested with `TopicArn`. Even if you use a custom generator, you always have to balance between what should go into the bindings and what should not. 

>Can you explain what you mean? I want to be sure we're talking about the same thing.
...
 IMO, if you are specifying in the 'schema' (is this a template?) what the message format should be, that would include the topicArn, as it should be to a known place, even if the message content will differ.

Yes and no, the problem you phase is that you have some dynamic data that should be included in your topicArn, which are not specified until at runtime (users) and some static data which never changes like the server you specify in your Async API file. That's why I am saying that some of the data might need to be in the Async API file and some should not. What is what I can honestly not tell you since it all depends on how SNS works, and even when you try to explain parts of it I cannot help you without fully understand it😄

karlwbrown:
Ah ok, I think I understand. So the templating is not part of asyncapi, but rather a feature of the generator. That makes sense.

I think re: SNS, the topicArn is not dynamic, or at least, in the simple cases I'm thinking of, it's not  dynamic; in most cases it could likely be fully specified in the original schema. Could also consider to what extent parameters might be useful here.

I think you can (reasonably) bind a server to an 'aws account' namespace, and then you could also somehow map a 'channel' (as a semantic concept) to a set of different *actual* channel/topicArns that will differ on a per-environment basis but still retain the same semantic meaning. 

For example, hello-world channel in AWS account 123456789 may be called 'hello-world-dev' and 'hello-world-test' and 'hello-world-prod'

The best way to specify that remains to be seen...

examples/2.0.0/aws-sns-publish.yml
fmvilas: Actually, I don't think that `helloWorld` should even be there since that would always be the name of the channel. Am I right?

IMO, all this data from the topicArn should be included as an AWS extension and not as part of the protocol binding. This data has nothing to do with the protocol but more with infrastructure/user details.

What do you think?

karlwbrown: As this is the SNS protocol, it seems like part of the bindings (SNS is AWS-only) - so everything that is AWS specific *should* go IMO into the SNS binding, no?

As for the 'name' of the channel, that's another question I have - because the name of the channel can be sometimes different across different environments, even though these channels have the same semantic meaning - e.g. maybe the channel is the 'helloWorld' channel but the actual topic in SNS is called 'helloWorld-dev-1818727' and 'helloWorld-test-11288273' or something.

fmvilas: > because the name of the channel can be sometimes different across different environments, even though these channels have the same semantic meaning

You just said it yourself. The topic name includes the environment, which is not part of the protocol but part of their product. By "protocol" I mean the "wire protocol", and bindings should only have the protocol features, not product features. Think about this way: tomorrow AWS starts offering to connect to their services using AMQP instead of their proprietary protocol. You still have environments but environments are not a feature of the AMQP protocol. Sure the information about the environment will be carried somewhere. The difficulty with AWS is that they have everything (infrastructure, product, and protocol) mixed up and bundled as a single thing.

Therefore, your channel or topic is actually `helloWorld`. If you deploy the same app to another region, environment or using another user, your channel should be the same. If that's not true, then maybe you'd need to have a different AsyncAPI file per environment and user, which I don't think is what you want.

For instance, given the following file:

```yaml
...
servers:
  dev:
    url: yoursnsurl.com
    protocol: sns
    x-aws-sns:
      environment: dev
      user: 1818727
  production:
    url: yoursnsurlprod.com
    protocol: sns
    x-aws-sns:
      environment: prod
      user: 1818727
channels:
  helloWorld:
    bindings:
      sns:
        Action: Publish
        # ...and more stuff here
```

The code generator and docs generator can process the information and give you the topicArn as `helloWorld-dev-1818727` or `helloWorld-prod-1818727` (depending on the server you choose), which is what you'll use in your code.

karlwbrown: Ah ok, I'm sorry my example above was probably confusing.

AWS itself does not have any restrictions on the topic name (see https://aws.amazon.com/sns/faqs/#Features_and_functionality)

I've pushed another commit that simplifies the bindings somewhat.

The main thing we need to be able to communicate through the spec is the topicArn for a given channel, while noting that this topicArn is a sort of 'global' locator for Amazon that uniquely identifies a *particular* channel, which is a combination of "AWS account" + "region" + "topicName".

karlwbrown:
I've updated the example to be a bit simpler now - welcome your additional comments...

I got rid of the "Action" binding, as I now feel it is distracting and also a bit confusing, and there are so many Actions that are beyond the scope of this API that I don't think it's worth specifying at all.

The new example assumes there are three environments - a dev/test environment (in one AWS account), and a production environment (in a different AWS account). This is a simple topology, many others are possible, but I just wanted to start with this example.

The main thing we need to communicate to the consumers, outside the context of the message, is the topicArn where they can read their messages.

These topicArns are a combination of region + account + topicName, and are given by AWS when you first create the topic.

There are a couple ways to do this - 
1. For each "semantic" topic, describe a channel for it, then use bindings or extensions to show the mapping between semantic topic name and topicArn for different environment/server/etc combinations. The sample posted tries to do this.
2. For each actual topic that exists, create a channel for it. This would lead to some amount of duplication, especially if you have more or less the same channel across multiple environments.
3. There are other ways I'm sure...

examples/2.0.0/aws-sns-publish.yml
karlwbrown: Also another note re: this:
'By "protocol" I mean the "wire protocol", and bindings should only have the protocol features, not product features.'

So maybe a core issue here is that SNS isn't really a 'wire protocol', per se. Rather, it is a product, that you interact with via HTTPS with certain parameters (to send messages to), and it in term can distribute messages over a number of different protocols, including SQS, HTTPS, SMTP, SMS, etc.

examples/2.0.0/aws-sns-publish.yml
```diff
@@ -0,0 +1,80 @@
...
+  development:
+    # This URL doesn't really describe a development server, but it does describe the
+    # top-level URL used to subscribe to messages from a given AWS region
+    url: 'https://sns.us-east-1.amazonaws.com/'
+    description: 'Development/test account for message publishing'
+    protocol: 'sns'
+    protocolVersion: '2010-03-31'
+    bindings:
+      # AccountID for this environment
+      awsAccountID: "123456789012"
```
fmvilas: What about putting the region here too? It's a feature of the server in the end:

```suggestion
      awsAccountID: "123456789012"
      region: "us-east-1"
```

We can do the same for the other servers.

examples/2.0.0/aws-sns-publish.yml
```diff
@@ -0,0 +1,80 @@
...
+      message:
+        oneOf:
+          - $ref: '#/components/messages/helloWorldMessage'
+          - $ref: '#/components/messages/goodNightWorldMessage'
+components:
+  channelBindings:
+    sns:
+      helloWorld:
+        development:
+          topicArn: 'arn:aws:sns:us-east-1:123456789012:helloWorld-dev'
```
fmvilas: If we put the region information suggested above, then we would not need this, right? The generator (or other software) could come up with the topicArn string by joining all the information: `arn:aws:sns:` + region from the server + `:` + account from the server + `:` + channel name.

Let me know if this makes sense. I'm trying to decouple the binding information from the server definitions.

PR #301, opened 2019-12-27
Add bindings definition

fmvilas:
### What?
I'm adding the definition of "binding" (or "protocol binding") and an incorrect piece of text saying that a "binding" can contain topology-related information.

### Why?
Make clear what is a "binding". We had definitions for what channel bindings, operation bindings, message bindings, and server bindings must be. However, we didn't have a definition of "binding" itself and it was leading to multiple interpretations.

### Why am I removing "and/or the topology"?
Bindings are one of the most exciting features of AsyncAPI. It lets you dig down to the specifics of each protocol without polluting the core spec. That said, bindings are also one of the most dangerous features in terms of spec evolution. We don't want bindings to become a black box where anyone can put whatever they want, from topology information to infrastructure/deployment details. We have specification extensions for that purpose and we'll give them more power soon with the creation of an extension registry.

### DIFF
--- a/versions/2.0.0/asyncapi.md
+++ b/versions/2.0.0/asyncapi.md
@@ -90,10 +90,10 @@ A message is the mechanism by which information is exchanged via a channel betwe
 A channel is an addressable component, made available by the server, for the organization of [messages](#definitionsMessage). [Producer](#definitionsProducer) applications send messages to channels and [consumer](#definitionsConsumer) applications consume messages from channels. Servers MAY support many channel instances, allowing messages with different content to be addressed to different channels. Depending on the server implementation, the channel MAY be included in the message via protocol-defined headers.
 
 #### <a name="definitionsProtocol"></a>Protocol
-A protocol is the mechanism (wireline protocol OR API) by which [messages](#definitionsMessage) are exchanged between the application and the [channel](#definitionsChannel). Example protocol include, but are not limited to, AMQP, HTTP, JMS, Kafka, MQTT, STOMP, WebSocket.  
+A protocol is the mechanism (wireline protocol or API) by which [messages](#definitionsMessage) are exchanged between the application and the [channel](#definitionsChannel). Example protocols include, but are not limited to, AMQP, HTTP, JMS, Kafka, MQTT, STOMP, WebSocket.  
 
 #### <a name="definitionsBindings"></a>Bindings
-An AsyncAPI binding (or "protocol binding") is a mechanism to define protocol-specific information. Therefore, infrastructure details, deployment information, and everything that's not related to the protocol MUST be left apart.
+A "binding" (or "protocol binding") is a mechanism to define protocol-specific information. Therefore, a protocol binding MUST define protocol-specific information only. 
 
 ## <a name="specification"></a>Specification
 
--- a/versions/2.0.0/asyncapi.md
+++ b/versions/2.0.0/asyncapi.md
@@ -93,7 +93,7 @@ A channel is an addressable component, made available by the server, for the org
 A protocol is the mechanism (wireline protocol OR API) by which [messages](#definitionsMessage) are exchanged between the application and the [channel](#definitionsChannel). Example protocol include, but are not limited to, AMQP, HTTP, JMS, Kafka, MQTT, STOMP, WebSocket.  
 
 #### <a name="definitionsBindings"></a>Bindings
-An AsyncAPI binding (or "protocol binding") is a mechanism to define protocol-specific information that's necessary to utilize the API. Therefore, infrastructure details, deployment information, and everything that's not related to how to communicate with the API MUST be left apart.
+An AsyncAPI binding (or "protocol binding") is a mechanism to define protocol-specific information. Therefore, infrastructure details, deployment information, and everything that's not related to the protocol MUST be left apart.
 
 ## <a name="specification"></a>Specification
 
--- a/versions/2.0.0/asyncapi.md
+++ b/versions/2.0.0/asyncapi.md
@@ -27,7 +27,7 @@ user/signedup:
 
 It means that the [application](#definitionsApplication) allows [consumers](#definitionsConsumer) to subscribe to the `user/signedup` [channel](#definitionsChannel) to receive userSignUp [messages](#definitionsMessage).
 
-**The AsyncAPI specification does not assume any kind of software topology, architecture or pattern.** Therefore, a server MAY be a message broker, a web server or any other kind of computer program capable of sending and/or receiving data. However, AsyncAPI offers a mechanism called "bindings" that aims to help with more specific information about the protocol and/or the topology.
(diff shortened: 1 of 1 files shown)


versions/2.0.0/asyncapi.md
```diff
@@ -92,6 +92,9 @@ A channel is an addressable component, made available by the server, for the org
 #### <a name="definitionsProtocol"></a>Protocol
 A protocol is the mechanism (wireline protocol OR API) by which [messages](#definitionsMessage) are exchanged between the application and the [channel](#definitionsChannel). Example protocol include, but are not limited to, AMQP, HTTP, JMS, Kafka, MQTT, STOMP, WebSocket.  
 
+#### <a name="definitionsBindings"></a>Bindings
+An AsyncAPI binding (or "protocol binding") is a mechanism to define protocol-specific information that's necessary to utilize the API. Therefore, infrastructure details, deployment information, and everything that's not related to how to communicate with the API MUST be left apart.
```
jonaslagoni: > An AsyncAPI binding (or "protocol binding") is a mechanism to define protocol-specific information that's necessary to utilize the **API**. Therefore, infrastructure details, deployment information, and everything that's not related to how to communicate with the **API** MUST be left apart.

Not sure people will truly understand that the API in this case is protocol interface, I know that is what you mean, but for someone who are new to the protocols and the spec might confuse it🤔 Ofcourse they read it in the section above, but still. We could write it like so:

> An AsyncAPI binding (or "protocol binding") is a mechanism to define protocol-specific information. Therefore, infrastructure details, deployment information, and everything that's not related to the protocol MUST be left apart.

fmvilas: That's a great suggestion, Jonas. Changing it.

versions/2.0.0/asyncapi.md
```diff
@@ -92,6 +92,9 @@ A channel is an addressable component, made available by the server, for the org
 #### <a name="definitionsProtocol"></a>Protocol
 A protocol is the mechanism (wireline protocol OR API) by which [messages](#definitionsMessage) are exchanged between the application and the [channel](#definitionsChannel). Example protocol include, but are not limited to, AMQP, HTTP, JMS, Kafka, MQTT, STOMP, WebSocket.  
 
+#### <a name="definitionsBindings"></a>Bindings
+An AsyncAPI binding (or "protocol binding") is a mechanism to define protocol-specific information. Therefore, infrastructure details, deployment information, and everything that's not related to the protocol MUST be left apart.
+
```
RobertDiebels: `deployment information,` No need for the comma after `information` if the summation is going to end.

`not related` I have a personal preference to use negations that are self-contained instead of adding the `not` so maybe `unrelated` is worth considering.

`MUST be left apart` maybe `MUST NOT be defined in a protocol binding` would be better here. I'm guessing that only protocol-specific information should be added here. Nothing else.

If I'm right and what's intended is `MUST NOT be defined [...]` then, you could consider dropping the double negation `not related` and `MUST NOT`. By changing it to a positive statement such as: `Therefore, a protocol binding MUST only define protocol-specific information.`. After the statement you can add examples of valid information to add. Since the range of valid definitions is probably much smaller than that of invalid definitions. 

This might seem redundant because the first sentence already mentioned `protocol-specific information`. However, that sentence introduced the concept of a `protocol binding` into the specification. So you can mention it again in the next one without to much of a problem.

derberg: If I got it right, you suggest this without examples:

```
An AsyncAPI binding (or "protocol binding") is a mechanism to define protocol-specific information. Therefore, infrastructure details, deployment information and everything unrelated to the protocol MUST NOT be defined in a binding.
```


or this but with examples:

```
An AsyncAPI binding (or "protocol binding") is a mechanism to define protocol-specific information. Therefore, a binding MUST define only protocol-specific information.
```

GraphQL introduced a counter examples in their spec https://graphql.github.io/graphql-spec/June2018/ to show bad examples and I think here it would also make sense. I would say I'm with option 2, with examples, positive and negative

RobertDiebels: @derberg Yes, if `MUST be left apart` can be interpreted as `MUST NOT be defined in a protocol binding.`
Which would indeed turn my suggestion into:
```
An AsyncAPI binding (or "protocol binding") is a mechanism to define protocol-specific information. Therefore, infrastructure details, deployment information and everything unrelated to the protocol MUST NOT be defined in a binding.
```
However, I would remove the double negation for clarity and cognitive load. By changing it to:
```
An AsyncAPI binding (or "protocol binding") is a mechanism to define protocol-specific information. Therefore, a protocol binding MUST only define protocol-specific information.
```
This makes it clear that only protocol-specific  information is acceptable. Since people tend to remember what was said at the start and the end of sentence. You could also state:
```
An AsyncAPI binding (or "protocol binding") is a mechanism to define protocol-specific information. Therefore, a protocol binding MUST define protocol-specific information only.
```
Afterwards you could add examples of valid information. Optionally you could also add examples of invalid information such as topology. Though that might be redundant because the sentence already clearly states that only protocol specific information is allowed. An example would be:

```
An AsyncAPI binding (or "protocol binding") is a mechanism to define protocol-specific information. Therefore, a protocol binding MUST define protocol-specific information only. A protocol binding MUST NOT contain information on topology, infrastructure, etc.. etc.. (this list is probably quite long)
```

derberg: oh, I thought that by examples you actually mean an example yaml/json of binding information that is correct and that is incorrect. That would be my preference, but I do not have a strong opinion about it

Grammarly tells me that it is better to write `MUST define protocol-specific information only` rather than `MUST only define protocol-specific information` :) so I'm with the first one

RobertDiebels: @derberg 
> I thought that by examples you actually mean an example yaml/json of binding information that is correct and that is incorrect.

That's an option as well. However, I meant the examples of invalid information which are given in the PR.

> Grammarly tells me that it is better to write [...]
Lets go with grammarly's suggestion then.
Which would give us this one:
```
An AsyncAPI binding (or "protocol binding") is a mechanism to define protocol-specific information. Therefore, a protocol binding MUST define protocol-specific information only. A protocol binding MUST NOT contain information on topology, infrastructure, etc.. etc.. (this list is probably quite long)
```
or this one:
```
An AsyncAPI binding (or "protocol binding") is a mechanism to define protocol-specific information. Therefore, a protocol binding MUST define protocol-specific information only. 
```
My preference would be the later. Since that one doesn't mention topology at all. So there is no reason to assume that it is valid information.

Fully-fledged yaml examples, both valid and in valid, can always be added later I presume.

fmvilas:
Addressed the comments from @RobertDiebels and @derberg. Not considering adding examples of good and bad use cases in this PR right now but I agree they would be great.

derberg:

Issue #312, opened 2020-02-11
Standardize basic contribution rules across all repos and basic repository settings


derberg:
Proposal:

### Forks only

Doesn't matter who you are, org member or not, all work on AsyncAPI should be done in Forks. PRs should come from Forks, and there should be no feature branches in the upstream.

**Why**
- A common practice for larger projects
- Get rid of issue of polluted list of branches, and zombie branches that were already merged.

**What has to be done**
- Clear instruction on how to setup your Fork so it is up to date with upstream master
  - explain the usage of [Pull bot](https://probot.github.io/apps/pull/)
  - explain how to set your local fork in the way that master is pointing to upstream

### Disable Wiki and Projects

**Why**
We do not use this feature, let's just disable it for all repositories.

**What has to be done**
* cleanup if we have some old stuff there
* disable those features and document for future

### Squash merging only

All repos should allow only `squash merging` of PRs

**Why**
To not have `Merge pull request ...` commits that are not very descriptive, but always have a commit message that introduces the commit. The advantage is also that you always have a final word on commit message before merging in.

Example of squash and merge, and just a merge
<img width="718" alt="Screenshot 2020-02-11 at 10 04 00" src="https://user-images.githubusercontent.com/6995927/74222946-e3088780-4cb5-11ea-9a81-367097964a23.png">

**What has to be done**
* set squash and merge  for all repos and document for future

### Introduce CODEOWNERS

It is a GitHub https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners that helps out to automate the assignment of PRs to proper group of people.

**Why**
You can, for example, specify that specific part of the repository can be only reviewed by a given person. If someone donates a template into the generator, he can become a main code owner to assure that only he/she can approve changes in it. 
Code owners is also well integrated with branch protection, so you can set a clear PR requirement

<img width="699" alt="Screenshot 2020-02-11 at 10 13 12" src="https://user-images.githubusercontent.com/6995927/74223527-30392900-4cb7-11ea-8a30-cab32ff0e251.png">

**What has to be done**
* create those files in all repos, specify code owners and then set proper branch protection
* document for future repos

Reason: 
To encourage contributions and tools adoption, we need to make sure work done on those projects is done consistently and in clean way.

jonaslagoni:
> ...
> ### Disable Wiki and Projects
> 
> **Why**
> We do not use this feature, let's just disable it for all repositories.
> ...


[...]

PR #317, opened 2020-02-20
Add info about what settings we have in all the repositories

derberg:
We agreed here #312 and on a SIG call that we will follow certain settings in all repos. Now it has to be documented


derberg:

PR #325, opened 2020-03-09
Add information to repository setting doc about branch deletion

derberg:
- Add info that `Automatically delete head branches` must be selected
- Add info about sponsoring option

See also: #312

### DIFF
--- a/documents/repository-settings.md
+++ b/documents/repository-settings.md
@@ -8,6 +8,8 @@ Under the repository name, choose the **Settings** tab. The **Options** view ope
     - Wikis
     - Projects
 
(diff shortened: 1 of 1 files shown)


fnobilia:
I missed the last call, did you decide to use CODEOWNERS or not? In case you would like to use this config, we could add a section here to specify that `master` branch should be protected and at least one CODEOWNER should approve before allowing a developer to merge.

derberg:
@fnobilia usage of CODEOWNERS is already agreed. We will do it, just a matter of setup, and yes, we want it set up with a branch protection

derberg:

Issue #314, opened 2020-02-20
Setup monitoring for all stale issues and PRs in all asyncapi repositories


derberg:
**Reason**
* We do not want to forget work done by the community and respect their effort in creatin issues or pull requests. We need a kind of alerting that will show that there is a work that is stale and still waiting for action
* We do not want to confuse community by keeping open work that is no longer relevant. To scale easily we need an automated way of picking up old things that need review and then maybe closing as no longer valid (with proper comment). 

**Description**
* Instead of using yet another tool in the project (that would normally be a probot and it's stale bot) we should use what we already have, in this case it is GitHub actions and this action https://github.com/marketplace/actions/close-stale-issues
* I propose we set 60days for marking issue/pr as `stale` and then 30days for automated closing of `stale` issue/pr
* only the following label should prevent us from staling the issue: `keep-open`. I know that previously in asyncapi repo we used stalebot with much reacher exclusion list https://github.com/asyncapi/asyncapi/blob/master/.github/stale.yml.  IMHO stale action should be used to remind us and regularly triage issues/prs ( ) and not just eventually close with no action. For example I do not want to keep `Make AsyncAPI a welcoming community` issue opened forever but be reminded that there is still some work needed to improve community experience. 
* stale label will be `stale` everywhere
* how we know stuff is staled:
  * we can have a slack messages for it https://github.com/marketplace/actions/post-slack-message, but I'm not sure if it won't be annoying and it will only consume our messages history on Slack
  * have a simple dashboard that would list all stale issues and we could always go through it during SIG calls. Problem is with ZenHub though, it doesn't list all issues in closed issues pipeline but only those that are not older by x. This is why we should make sure stale issues don't get close too soon, this is why I proposed 30d and not 7d to close stale issue.
* Document somewhere in our community docs how we are handling old issues and how to check them

PR #326, opened 2020-03-10
Add GitHub Actions powered stale bot and new contrib welcome

derberg:
- Official GH Stale action is used here https://github.com/marketplace/actions/close-stale-issues
- Official GH welcome action is used here https://github.com/marketplace/actions/first-interaction
- Probot stalebot config is out

fixes: #314


derberg:

PR #327, opened 2020-03-10
Add info in repo settings instruction about what basic workflows we need

derberg:
Final fix for #314 to document what workflows we should have in each repo


derberg:

PR #382, opened 2020-05-05
docs: recognize new contributors and people that helped to organize conference

derberg:
**Description**

- add some missing info to existing contributors
- change to display 3 in a row, not 2
- add new contributors


derberg:

PR #403, opened 2020-07-01
chore: remove funding config file to apply global from .github repo

derberg:
**Description**

- remove funding config file to apply global from .github repo


derberg:

PR #424, opened 2020-08-06
fix: Schema should not enforce minProperties

derberg:
**Description**

- Schema should not enforce minProperties. Schema is not a spec but a tool that helps to work with spec, and should not add rules not described in the spec.

**Related issue(s)**
Resolves https://github.com/asyncapi/tck/issues/30


derberg:

PR #426, opened 2020-08-10
fix: List of scopes can be empty for oauth2 and openIdConnect

derberg:
**Description**

- List of scopes can be empty for oauth2 and openIdConnect

**Related issue(s)**
See also https://github.com/asyncapi/tck/issues/22#issuecomment-671407523


derberg:

derberg:

PR #452, opened 2020-10-06
Ref usage Fixes #406

### LINKED ISSUE #406: $ref are not defined in every level it may be defined
Issue part of #397 

$ref not defined in every level it may be defined. Check OAS specs for instance, it’s strictly defined where $ref field can be used and where it can’t, it’s not a free form inclusion to the AST. For example, there can’t be a $ref in an info object. A place where this is completely misleading is in the use of traits. The operation has an array of trait objects, but the field $ref is not there, so it can’t be referenced. It also states that the objects can be present in that array, which makes no sense the traits facet should only have refs. What would be the point of having an inline trait? Not having refs strictly defined (as OAS has) is confusing for the user.
https://github.com/asyncapi/asyncapi/blob/master/versions/2.0.0/asyncapi.md#file-structure

BlockLucas:
Define Reference Object usage in every part it could be used as discussed in #406


versions/2.0.0/asyncapi.md
```diff
@@ -348,7 +348,7 @@ Field Name | Type | Description
...
 <a name="serverObjectVariables"></a>variables | Map[`string`, [Server Variable Object](#serverVariableObject)] | A map between a variable name and its value.  The value is used for substitution in the server's URL template.
 <a name="serverObjectSecurity"></a>security | [[Security Requirement Object](#securityRequirementObject)] | A declaration of which security mechanisms can be used with this server. The list of values includes alternative security requirement objects that can be used. Only one of the security requirement objects need to be satisfied to authorize a connection or operation.
-<a name="serverObjectBindings"></a>bindings | [Server Bindings Object](#serverBindingsObject) | A map where the keys describe the name of the protocol and the values describe protocol-specific definitions for the server.
+<a name="serverObjectBindings"></a>bindings | [Server Bindings Object](#serverBindingsObject) \| [Reference Object](#referenceObject) | A map where the keys describe the name of the protocol and the values describe protocol-specific definitions for the server.
```
fmvilas, line 351: This would result in something like:

```yaml
servers:
  myServer:
    bindings:
      $ref: '...'
```

Is this what you want to mean here?

versions/2.0.0/asyncapi.md
```diff
@@ -547,12 +547,12 @@ Describes the operations available on a single channel.
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="channelItemObjectRef"></a>$ref | `string` | Allows for an external definition of this channel item. The referenced structure MUST be in the format of a [Channel Item Object](#channelItemObject). If there are conflicts between the referenced definition and this Channel Item's definition, the behavior is *undefined*.
+<a name="channelItemObjectRef"></a>$ref | [Reference Object](#referenceObject) | Allows for an external definition of this channel item. The referenced structure MUST be in the format of a [Channel Item Object](#channelItemObject). If there are conflicts between the referenced definition and this Channel Item's definition, the behavior is *undefined*.
```
fmvilas, line 550: If I understood correctly, this is not correct. This would result in something like:

```yaml
channels:
  test:
    $ref:
      $ref: '...'
```

I think it should be done here: https://github.com/asyncapi/asyncapi/blob/master/versions/2.0.0/asyncapi.md#channelsObjectChannel. Something like `{channel} | [Channel Item Object](#channelItemObject) \| [Reference Object](#referenceObject)`

versions/2.0.0/asyncapi.md
```diff
@@ -650,8 +650,8 @@ Field Name | Type | Description
...
-<a name="operationObjectTraits"></a>traits | [[Operation Trait Object](#operationTraitObject) &#124; [Reference Object](#referenceObject) ] | A list of traits to apply to the operation object. Traits MUST be merged into the operation object using the [JSON Merge Patch](https://tools.ietf.org/html/rfc7386) algorithm in the same order they are defined here.
+<a name="operationObjectBindings"></a>bindings | [Operation Bindings Object](#operationBindingsObject) \| [Reference Object](#referenceObject)| A map where the keys describe the name of the protocol and the values describe protocol-specific definitions for the operation.
+<a name="operationObjectTraits"></a>traits | [ [Reference Object](#referenceObject) ] | A list of traits to apply to the operation object. Only references are allowed here as inlined traits would not be reusable. Traits MUST be merged into the operation object using the [JSON Merge Patch](https://tools.ietf.org/html/rfc7386) algorithm in the same order they are defined here.
```
fmvilas, line 654: I'd not change this as it would be a breaking change. Maybe this can be proposed for next major version on a separate issue.

fmvilas (changes requested):

derberg:
@fmvilas hey, bot closed this PR which is still important. The good thing is that @AgustinBettati took over here
https://github.com/asyncapi/asyncapi/pull/498. Can you have a look?

PR #453, opened 2020-10-15
Clarify message examples definition

fmvilas:
**Description**
This clarifies the structure of examples on the Message Object.

**Pending stuff**
- [x] Update the JSON Schema definition in https://github.com/asyncapi/asyncapi-node

**Related issue(s)**
See also https://github.com/asyncapi/html-template/issues/59


versions/2.0.0/asyncapi.md
```diff
@@ -1003,7 +1003,7 @@ Field Name | Type | Description
 <a name="messageObjectTags"></a>tags | [Tags Object](#tagsObject) | A list of tags for API documentation control. Tags can be used for logical grouping of messages.
 <a name="messageObjectExternalDocs"></a>externalDocs | [External Documentation Object](#externalDocumentationObject) | Additional external documentation for this message.
 <a name="messageObjectBindings"></a>bindings | [Message Bindings Object](#messageBindingsObject) | A map where the keys describe the name of the protocol and the values describe protocol-specific definitions for the message.
-<a name="messageObjectExamples"></a>examples | [Map[`string`, `any`]] | An array with examples of valid message objects.
+<a name="messageObjectExamples"></a>examples | [Map[`string`, `any`]] | An array of key/value pairs where keys MUST be either "headers" and/or "payload". Values MUST contain examples of valid headers or payload definitions, respectively.
```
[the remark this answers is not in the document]
Tenischev: `A map of two elements. The first one has a 'key'="headers" where a 'value' is an example of defined headers. The second element has a 'key'="payload" where a 'value' is an example of defined payload`.
...or something like this?...
...It will be great to set links to the correspond elements `headers` and `payload`...

versions/2.0.0/asyncapi.md
```diff
@@ -1105,6 +1121,15 @@ correlationId:
...
   - $ref: "#/components/messageTraits/commonHeaders"
+examples:
+  - headers:
+      correlationId: my-correlation-id
+      applicationInstanceId: myInstanceId
+    payload:
+      user:
+        someUserKey: someUserValue
+      signup:
+        someSignupKey: someSignupValue
```
Tenischev, line 1132: Is it a good idea to show example of not `Schema object`?

versions/2.0.0/asyncapi.md
derberg, line 1006:
```suggestion
<a name="messageObjectExamples"></a>examples | [Map[`string`, `any`]] | An array of key/value pairs where keys MUST be either **headers** and/or **payload**. Values MUST contain valid examples of `headers` or `payload` fields, respectively.
```

derberg:
more important is I think to clarify:
- is this related to 2.0.0 or maybe a patch. If 2.0.0 then it must be fixed in website. 
- what about the json schema?

versions/2.0.0/asyncapi.md
fmvilas: This shouldn't be a Schema object. It's an example of what you will transfer down the wire not an AsyncAPI example.

fmvilas:
@derberg 

> is this related to 2.0.0 or maybe a patch. If 2.0.0 then it must be fixed in website.

This is something outside the concern of this PR. There are a lot of changes that are not reflected on the website, including the patches or clarifications like this one. Agree it should be there but not something to tackle in this PR.

> what about the json schema?

Thanks! Completely forgot about it.

versions/2.0.0/schema.json
```diff
@@ -690,7 +690,14 @@
...
                   "type": "array",
                   "items": {
-                    "type": "object"
+                    "type": "object",
+                    "additionalProperties": false,
+                    "properties": {
+                      "headers": {
+                        "type": "object"
+                      },
+                      "payload": {}
```
fmvilas, line 699: Since `payload` can be anything we don't validate it at all.

derberg (approved):

derberg:

derberg:

PR #495, opened 2021-02-03
test: support testing on windows

derberg:
**Description**

- support testing on windows. With new version of mdv we would have this solved as you can see here https://github.com/asyncapi/asyncapi/pull/495/checks?check_run_id=1823847384

**Related issue(s)**
See also https://github.com/asyncapi/asyncapi/pull/493/checks?check_run_id=1817232881

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,9 +24,17 @@
       }
     },
     "ansi-regex": {
-      "version": "3.0.0",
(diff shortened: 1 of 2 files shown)


derberg:
after contributing fix to mdv it was already released and works like a charm 🥳

fmvilas (approved):

derberg:

derberg:

PR #498, opened 2021-02-11
fix: clarify use of reference objects in async 2.0.0

AgustinBettati:
**Description**
Explicitly clarify all the places where reference objects may be used. This will make the spec more coherent as currently the reference object is only defined in certain places which leads to confusion.
I do not consider these changes as breaking, as they are simply additions which reflect expected behavior.

**Related issue(s)**
Relates to https://github.com/asyncapi/asyncapi/issues/406


derberg:
@AgustinBettati hey, can you update the branch and resolve conflicts 🙏🏼

fmvilas (approved):
LGTM 👍

PR #502, opened 2021-02-25
feat: add SASL security scheme types for use by Kafka specs

### LINKED ISSUE #466: Add security mechanism info to Kafka bindings for server objects
**Is your feature request related to a problem? Please describe.**
Kafka has several values that are essential to know when preparing to connect to a secured Kafka cluster, such as `sasl.mechanism` and `security.protocol`. Adding these to the Kafka binding for server objects would allow these to be documented.

**Can't it be tackled using specification extensions?**
It could be captured in extensions.

e.g.
```
asyncapi: '2.0.0'
...
servers:
  production:
    url: localhost:9092
    protocol: kafka-secure
    security:
    - saslScramCreds: []
...
components:
  securitySchemes:
    saslScramCreds:
      type: userPassword
      x-mykafka-sasl-mechanism: 'SCRAM-SHA-256'
      x-mykafka-security-protocol: 'SASL_SSL'
```

However I think this is a common enough request that there would be value in doing it in a consistent way. In particular, it would enable code generation to set up the right Kafka client properties.

**Describe the solution you'd like**
Strawman suggestion:
```
asyncapi: '2.0.0'
...
servers:
  production:
    url: localhost:9092
    protocol: kafka-secure
    security:
    - saslScramCreds: []
    bindings:
      kafka:
        saslMechanism: 'SCRAM-SHA-256'
        securityProtocol: 'SASL_SSL'
...
components:
  securitySchemes:
    saslScramCreds:
      type: userPassword
```

**Describe alternatives you've considered**
Fran Méndez [suggested](https://asyncapi.slack.com/archives/C34F2JV0U/p1606406595418100?thread_ts=1606403605.417600&cid=C34F2JV0U): 
> We should make it part of the security schemes, it's not only for Kafka after all.

dalelane:
As discussed in https://github.com/asyncapi/bindings/pull/56 this
adds additional security scheme types. The motivation for adding
them is to enable description of secured Kafka clusters, however
the security protocols and mechanisms being added are not unique
to Kafka, so this commit adds them as generic security schemes so
they could be used by other protocols as well.

Fixes #466

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -1965,7 +1965,7 @@ Defines a security scheme that can be used by the operations. Supported schemes 
 ##### Fixed Fields
 Field Name | Type | Applies To | Description
 ---|:---:|---|---
-<a name="securitySchemeObjectType"></a>type | `string` | Any | **REQUIRED**. The type of the security scheme. Valid values are `"userPassword"`, `"apiKey"`, `"X509"`, `"symmetricEncryption"`, `"asymmetricEncryption"`, `"httpApiKey"`, `"http"`, `oauth2`, and `openIdConnect`.
+<a name="securitySchemeObjectType"></a>type | `string` | Any | **REQUIRED**. The type of the security scheme. Valid values are `"userPassword"`, `"apiKey"`, `"X509"`, `"symmetricEncryption"`, `"asymmetricEncryption"`, `"httpApiKey"`, `"http"`, `"oauth2"`, and `"openIdConnect"`, `"plain"`, `"scramSha256"`, `"scramSha512"`, `"gssapi"`.
 <a name="securitySchemeObjectDescription"></a>description | `string` | Any | A short description for security scheme. [CommonMark syntax](http://spec.commonmark.org/) MAY be used for rich text representation.
 <a name="securitySchemeObjectName"></a>name | `string` | `httpApiKey` | **REQUIRED**. The name of the header, query or cookie parameter to be used.
 <a name="securitySchemeObjectIn"></a>in | `string` | `apiKey` \| `httpApiKey` | **REQUIRED**. The location of the API key. Valid values are `"user"` and `"password"` for `apiKey` and `"query"`, `"header"` or `"cookie"` for `httpApiKey`.
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -1961,6 +1961,7 @@ Defines a security scheme that can be used by the operations. Supported schemes 
 * HTTP API key.
 * OAuth2's common flows (Implicit, Resource Owner Protected Credentials, Client Credentials and Authorization Code) as defined in [RFC6749](https://tools.ietf.org/html/rfc6749).
 * [OpenID Connect Discovery](https://tools.ietf.org/html/draft-ietf-oauth-discovery-06).
+* SASL (Simple Authentication and Security Layer) as defined in [RFC4422](https://tools.ietf.org/html/rfc4422).
 
 ##### Fixed Fields
 Field Name | Type | Applies To | Description
@@ -2099,6 +2100,18 @@ flows:
     scopes:
       write:pets: modify pets in your account
       read:pets: read your pets
+```
+
+###### SASL Sample
+
+```json
+{
+  "type": "scramSha512"
+}
+```
+
+```yaml
+type: scramSha512
 ```
 
 #### <a name="oauthFlowsObject"></a>OAuth Flows Object
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -1965,7 +1965,7 @@ Defines a security scheme that can be used by the operations. Supported schemes 
 ##### Fixed Fields
 Field Name | Type | Applies To | Description
 ---|:---:|---|---
-<a name="securitySchemeObjectType"></a>type | `string` | Any | **REQUIRED**. The type of the security scheme. Valid values are `"userPassword"`, `"apiKey"`, `"X509"`, `"symmetricEncryption"`, `"asymmetricEncryption"`, `"httpApiKey"`, `"http"`, `"oauth2"`, and `"openIdConnect"`, `"plain"`, `"scramSha256"`, `"scramSha512"`, `"gssapi"`.
+<a name="securitySchemeObjectType"></a>type | `string` | Any | **REQUIRED**. The type of the security scheme. Valid values are `"userPassword"`, `"apiKey"`, `"X509"`, `"symmetricEncryption"`, `"asymmetricEncryption"`, `"httpApiKey"`, `"http"`, `"oauth2"`, `"openIdConnect"`, `"plain"`, `"scramSha256"`, `"scramSha512"`, and `"gssapi"`.
 <a name="securitySchemeObjectDescription"></a>description | `string` | Any | A short description for security scheme. [CommonMark syntax](http://spec.commonmark.org/) MAY be used for rich text representation.
 <a name="securitySchemeObjectName"></a>name | `string` | `httpApiKey` | **REQUIRED**. The name of the header, query or cookie parameter to be used.
 <a name="securitySchemeObjectIn"></a>in | `string` | `apiKey` \| `httpApiKey` | **REQUIRED**. The location of the API key. Valid values are `"user"` and `"password"` for `apiKey` and `"query"`, `"header"` or `"cookie"` for `httpApiKey`.


nictownsend:
_Originally posted by @dalelane in https://github.com/asyncapi/bindings/issues/56#issuecomment-785834586_

AsyncAPI: <br>server [`protocol`](https://www.asyncapi.com/docs/specifications/2.0.0#fixed-fields-4) | AsyncAPI: <br> security scheme [`type`](https://www.asyncapi.com/docs/specifications/2.0.0#fixed-fields-21) | meaning<br>encryption? | meaning<br>auth? | maps to Kafka client parameter<br>[`sasl.mechanism`](https://kafka.apache.org/documentation/#consumerconfigs_sasl.mechanism) | maps to Kafka client parameter<br>[`security.protocol`](https://kafka.apache.org/documentation/#consumerconfigs_security.protocol)
---|:---:|:---:|:---:|:---:|---
`kafka` |  | no | no | _unset_ | `PLAINTEXT` 
`kafka` | `plain` <sup>*</sup> | no | yes, using SASL/PLAIN | `PLAIN` | `SASL_PLAINTEXT`
`kafka` | `scramSha256` <sup>*</sup> | no | yes, using SASL/SCRAM | `SCRAM-SHA-256` | `SASL_PLAINTEXT` 
`kafka` | `scramSha512` <sup>*</sup> | no | yes, using SASL/SCRAM | `SCRAM-SHA-512` | `SASL_PLAINTEXT` 
`kafka` | `oauth2`  | no | yes, using OAuth | `OAUTHBEARER` | `SASL_PLAINTEXT` 
`kafka` | `gssapi` <sup>*</sup> | no | yes, using GSSAPI | `GSSAPI` | `SASL_PLAINTEXT` 
`kafka-secure` |  |  yes | no | _unset_ | `SSL`
`kafka-secure` | `plain` <sup>*</sup> | yes | yes, using SASL/PLAIN | `PLAIN` | `SASL_SSL`
`kafka-secure` | `scramSha256` <sup>*</sup> | yes | yes, using SASL/SCRAM | `SCRAM-SHA-256` | `SASL_SSL` 
`kafka-secure` | `scramSha512` <sup>*</sup> | yes | yes, using SASL/SCRAM | `SCRAM-SHA-512` | `SASL_SSL` 
`kafka-secure` | `oauth2`  | yes | yes, using OAuth | `OAUTHBEARER` | `SASL_SSL` 
`kafka-secure` | `gssapi` <sup>*</sup> | yes | yes, using GSSAPI | `GSSAPI` | `SASL_SSL` 
`kafka-secure` | `X509` | yes | yes, using mutual TLS | _unset_ | `SSL` 


<sup>*</sup> - a new security scheme type to be added

This is all working on the assumption that `kafka/kafka-secure` is equivalent to `http/https` - https://asyncapi.slack.com/archives/C34F2JV0U/p1614255214067200 @fmvilas for reference

If it's not, then there will need to be more to achieve the table above.

nictownsend:
This is all working on the assumption that `kafka/kafka-secure` is equivalent to `http/https` - https://asyncapi.slack.com/archives/C34F2JV0U/p1614255214067200 @fmvilas for reference

If it's not, then there will need to be more to achieve the table above.

versions/2.0.0/asyncapi.md
```diff
@@ -1961,7 +1961,7 @@ Defines a security scheme that can be used by the operations. Supported schemes
 ##### Fixed Fields
 Field Name | Type | Applies To | Description
 ---|:---:|---|---
-<a name="securitySchemeObjectType"></a>type | `string` | Any | **REQUIRED**. The type of the security scheme. Valid values are `"userPassword"`, `"apiKey"`, `"X509"`, `"symmetricEncryption"`, `"asymmetricEncryption"`, `"httpApiKey"`, `"http"`, `oauth2`, and `openIdConnect`.
+<a name="securitySchemeObjectType"></a>type | `string` | Any | **REQUIRED**. The type of the security scheme. Valid values are `"userPassword"`, `"apiKey"`, `"X509"`, `"symmetricEncryption"`, `"asymmetricEncryption"`, `"httpApiKey"`, `"http"`, `"oauth2"`, and `"openIdConnect"`, `"plain"`, `"scramSha256"`, `"scramSha512"`, `"gssapi"`.
```
smoya, line 1964: Wouldn't `plain` be covered already by the `userPassword`?

smoya: Would it make sense to reuse the `userPassword` schema, adding an optional field for the encryption?

nictownsend: I think that depends whether we want `userPassword` and `X509` to be reusable across protocol.

i.e it's inferred (or we state) - that for `kafka`, `userPassword` = SASL plain, `X509` = mutual TLS, `scramShaX` = SASL scram etc

nictownsend: I guess it becomes a challenge if kafka (or other protocols) add support for clashing ideas (e.g SASL x509 + mutual tls)

nictownsend: > adding an optional field for the encryption
Can we assume that protocol `kafka-secure` means encrypted and `kafka` is plain?

smoya: > i.e it's inferred (or we state) - that for `kafka`, `userPassword` = SASL plain, `X509` = mutual TLS, `scramShaX` = SASL scram etc

Perhaps it makes sense to add the concept of aliases that at the end are just reusing the same JSON Schema types?  What I would like to avoid is redefining the same concept with a different name and maintain both at the same time.

> Can we assume that protocol kafka-secure means encrypted and kafka is plain?

Yeah, that's a good point 👍 . What does it **should** mean `-secure` suffix here? I guess the question is: Do we talk about encryption or the protocol?

nictownsend: > What does it should mean -secure suffix here? I guess the question is: Do we talk about encryption or the protocol?

If we compare `http/https` - the `s` represents `secure` protocol (via SSL encryption) - at least, I believe that's the case. If so, then that maps accordingly to `kafka/kafka-secure` - the `secure` means it's a secure version of the protocol?

dalelane: I agree with this. I think being consistent with the distinction between http and https would be best - so in the same way that using https means encryption and does not require authentication, I think kafka-secure should mean encryption but not require auth.

dalelane: > Would it make sense to reuse the `userPassword` schema, adding an optional field for the encryption? 

I don't really mind. 

The root requirement is that there at least three SASL mechanisms typically used for Kafka that could all be described as username/password-based (SASL/PLAIN, SASL/SCRAM-SHA-256, SASL/SCRAM-SHA-512) so we either need to describe them as different security scheme types, or make this SASL mechanism information a new additional attribute of the security scheme. 

Either is totally fine with me, as long as we have a way to articulate the required config. 

@fmvilas Any strong preferences?

fmvilas: Not really. Either way is fine for me. Like you said, as long as we have a way to articulate the config, everything is fine. Just let's make sure it's not too verbose and it's maintainable in the long-term.

dalelane:
Updated the PR to address merge conflicts. 

I _think_ all the comments above are addressed and this is ready for re-review now, please.

dalelane:
@fmvilas @derberg Is there anything else needed for this one, please?

fmvilas:
I think it's pretty much done on the spec side. The only missing part for me is to update the JSON Schema definition and the JS Parser.

Regarding the JSON Schema definition, create a new document called `2.1.0.json` and add these new security schemes to the definition [here](https://github.com/asyncapi/asyncapi-node/blob/master/schemas/2.0.0.json#L933). Just create a pull request, doesn't have to be merged.

Regarding the JS Parser, similar thing: create a PR and [make the `@asyncapi/specs` (this is `asyncapi-node` above) point to the branch you created above](https://github.com/asyncapi/parser-js/blob/master/package.json#L71). Again, PR doesn't have to be merged.

If you meet these two criteria, we'll move this to the next stage ([Draft RFC 2](https://github.com/asyncapi/spec/blob/master/CONTRIBUTING.md#stage-2-draft)).

If you're not familiar with JS, we're happy to help on that front.

fmvilas:
Moving this to the next stage 🎉 It's now a Draft, pending testing it on the JS parser.

derberg:
In my opinion, we should also get examples of new schemes like in case of others from the list https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md#security-scheme-object-example, also for me it is not clear what is the structure of information that one has to provide when using `scramSha512` for example.

+ above the table that is modified in this PR is the list of supported schemes, would be good to add the new ones there and even link them to some vendor-neutral links (if available) that explain more.

Thoughts?

derberg:
what do you think about extending the [streetlight example](https://github.com/asyncapi/spec/blob/master/examples/2.0.0/streetlights.yml)?

derberg:
@dalelane can you have a look at my last comment?

derberg:
@fmvilas I checked the streetlights example and I actually revert my request to @dalelane to extend it with SASL, as SASL is for Kafka and example is for MQTT. Problem is that this example, even though for mqtt, has kafka bindings. So the example is a bit messy and I would not like to block this PR because of it.

I had quick chat with @dalelane and he, as our Kafka ❤️ AsyncAPI expert, will work on another PR to fix the existing example and provide a separate Kafka example.

@fmvilas I think we are ready to go with these PRs related to SASL. The way to merge should be:
1. `spec` PR merge
2. `asyncapi-node` PR merge
3. wait for `asyncapi-node` release candidate for NPM
4. bump this candidate in PR in `parser-js`
5. `parser-js` PR merge
I think this is it 🤔

dalelane:
@derberg FYI - I've had a go at preparing a new version of the Streetlight example in https://github.com/asyncapi/spec/pull/548

PR #511, opened 2021-03-14
chore: improve contributing guidelines

fmvilas:
**Description**

- Adds information about the new stages: strawman, proposal, draft, and accepted (or rejected).
- Adds some guiding principles.

**To-do**

- [ ] Clean labels of this repository and add a few ones for each stage, `Needs Champion`, etc. Look at https://github.com/graphql/graphql-spec/labels for inspiration.

**Related issue(s)**
https://github.com/asyncapi/.github/pull/37

PR #516, opened 2021-03-16
chore: add new issue and PR templates according to the new spec process

fmvilas:
### Description

This PR removes the previous issue templates and creates a new one for issues and another for pull requests. They give you hints on how to contribute according to the new spec process.

### Related issues
#511


aayushmau5:
Is it feasible to do the PR template as a checkbox and Q/A thing? Like 

- [ ] Have you read the PR guidelines?

- Describe your PR:
   **answer**
- Issue related
  #**issue no.**

I feel like this kind of thing forces people to actually do read the PR guidelines, and helps with the format.

fmvilas:
This is more or less what we have right now but after triaging some issues it probably still makes sense to have this format of Q/A 👍

fmvilas:
What about now? Many more options 😅

aayushmau5:
Yeah, that's good :laughing:

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,15 @@
...
+
+Before creating your issue:
+
+* **Have a question?** Find community resources at [asyncapi/community](https://www.github.com/asyncapi/community).
+
+* **Found an editing mistake?** Create a Pull Request with the edited fix. The Github UI allows you to edit files directly, find the source files at [asyncapi/spec/tree/master/spec](https://www.github.com/asyncapi/spec/tree/master/spec).
+
+* **Improvements to documentation?** Head over to [asyncapi/website](https://www.github.com/asyncapi/website).
+
+* **Feature request?** First read [asyncapi/spec/blob/master/CONTRIBUTING.md](https://www.github.com/asyncapi/spec/blob/master/CONTRIBUTING.md) and prefer creating a Pull Request!
```
derberg, line 13: where will this display and what do you mean by `and prefer creating a Pull Request!`?

.github/ISSUE_TEMPLATE/question.md
```diff
@@ -0,0 +1,23 @@
...
+assignees: ''
+---
+
+<!--
+
+!!! IMPORTANT !!!
+
+Before creating your issue:
+
+* **Have a question?** Find community resources at [asyncapi/community](https://www.github.com/asyncapi/community).
```
derberg, line 15: you have some plans to extend https://www.github.com/asyncapi/community with some resources? maybe better link here https://github.com/asyncapi/community/issues/1 ?

.github/PULL_REQUEST_TEMPLATE/strawman.md
```diff
@@ -0,0 +1,19 @@
+---
+name: 💡 Proposal (RFC 1)
```
derberg, line 2: did you check how those templates and prs would show up? on some test repo? Last time I checked GH supported only one PR template

.github/ISSUE_TEMPLATE.md
fmvilas: This will be displayed every time someone creates an issue that's not using an issue template, like when you click "Create blank new issue". And the "prefer a PR for a feature request" is because of this: https://github.com/asyncapi/spec/pull/511/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R104.

.github/ISSUE_TEMPLATE/question.md
fmvilas: That's a good point. I'll change it.

.github/PULL_REQUEST_TEMPLATE/strawman.md
fmvilas: No, I haven't. And thanks for bringing the comment up. This file is wrong. It says "strawman" but in fact, it's a proposal RFC 1.

smoya: You can choose the template to use by adding the `?template=issue_template.md`. Unfortunately, there is no UI for that yet.

More info at https://docs.github.com/en/github/managing-your-work-on-github/about-automation-for-issues-and-pull-requests-with-query-parameters#supported-query-parameters

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,15 @@
+<!--
```
smoya, line 1: > You are using an old version of issue templates. Please update to the new issue template workflow.

It seems we are using an old template version. You can check the warning when watching the file: https://github.com/asyncapi/spec/blob/3535538236c899d282847aa979425acb0e54567e/.github/ISSUE_TEMPLATE.md

fmvilas: Good catch. Thanks, Sergio.

.github/ISSUE_TEMPLATE/question.md
fmvilas: Changed!

.github/PULL_REQUEST_TEMPLATE/strawman.md
fmvilas: Replaced Proposal (RFC 1) with Strawman (RFC 0).

derberg: @smoya how would it work with PR, no example in docs? I think we need one template with all sections and contributor just need to leave the one that is relevant

.github/ISSUE_TEMPLATE.md
fmvilas: I added a frontmatter but it's still complaining. Not sure what to do. It doesn't say anything on the referenced docs 🤷

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,20 @@
...
+about: ""
+---
+
+<!--
+
+!!! IMPORTANT !!!
+
+Before creating your issue:
+
+* **Have a question?** Find community resources at https://github.com/asyncapi/community/issues/1.
```
smoya, line 12: Not sure if already discussed but you can include those in the template chooser so it looks much cleaner: https://docs.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository#configuring-the-template-chooser

.github/ISSUE_TEMPLATE.md
smoya: You can create them via https://github.com/asyncapi/spec/issues/templates/edit

.github/PULL_REQUEST_TEMPLATE/strawman.md
smoya: If it's gonna be always the same template, yes one should be enough. Otherwise you can also set a default one but also have more templates people can use (by adding the query parameter)

derberg: the thing is that if for now, the only way to use different PR templates is to use some mysterious query param, it is not very helpful. I mean who would even remember the names of templates anyway 🤷🏼 this is why better to consolidate all into one template that will always show up

smoya: Yup, makes sense as the ux doesnt help this time 👍

.github/ISSUE_TEMPLATE.md
fmvilas: It is included. There's a "question" type of issue. This one is for those who consider choosing "Open a new blank issue". It's just so they think twice 😄

.github/PULL_REQUEST_TEMPLATE/strawman.md
fmvilas: Ok, I'll remove them. It doesn't make sense with the PR workflow if there's no UI for that.

fmvilas: How about now?

derberg (approved):
:shipit: 
but probably once the other PR is merged

.github/ISSUE_TEMPLATE.md
fmvilas: Yeah, that doesn't help. It allows you to create templates for different types of issues. What I'm editing here is the default one when people choose "Open new blank issue".

smoya (approved):
LGTM 🚀

derberg:

derberg:

Issue #505, opened 2021-03-04
[BUG] application of message traits (intentionally) replacing existing attributes


c-pius:
**Describe the bug**
According to specification, [message traits](https://github.com/asyncapi/asyncapi/blob/master/versions/2.0.0/asyncapi.md#messageObject) are applied using `json-merge-patch` protocol.

> A list of traits to apply to the message object. Traits MUST be merged into the message object using the JSON Merge Patch algorithm in the same order they are defined here. The resulting object MUST be a valid Message Object.

The implication of this is that properties defined in a trait will "override" those defined in the message itself. I am not sure if this is actually a bug (it is just how `json-merge-patch` works) but I personally think this is a bit unfortunate.

A concrete scenario where I run into troubles is that I have a trait that is defining some standard message headers where some of them are required. Then I have several messages that include some specific header attributes per message where also some of them are required. Now the problem is, that when the trait is applied, the required attributes from the message are replaced by the ones from the trait (see the examples below for more clarity).

Of course I could also just remove the required array from the trait and add them into the message, but then the trait is not really self-contained anymore... Therefore I wanted to check if this behavior of traits is really the intended one? Or if there are any other mitigations I could take?

**To Reproduce**
1. Take the sample document below
2. Parse it with `parser-js`
3. Check the resulting message

```json5
{
  // ...
  "components": {
    "messages": {
      "channelMessage": {
        "headers": {
          // ...
          "required": [ // ["req-header"] has been replaced by the trait
            "t-req-header"
          ]
        }
      }
    }
    // ...
  }
}
```

**Expected behavior**

```json5
{
  // ...
  "components": {
    "messages": {
      "channelMessage": {
        "headers": {
          // ...
          "required": [
            "req-header",
            "t-req-header"
          ]
        }
      }
    }
    // ...
  }
}
```

**Sample document**

```yaml
asyncapi: 2.0.0
info:
  title: My API
  version: '1.0.0'

channels:
  mychannel:
    publish:
      message:
        $ref: '#/components/messages/channelMessage'

components:
  messages:
    channelMessage:
      traits:
        - $ref: '#/components/messageTraits/some-trait'
      headers:
        type: object
        properties:
          req-header:
            type: integer
          non-req-header:
            type: integer
        required:
          - req-header
  schemas:
    testSchema:
      type: object
      properties:
        name:
          type: string
  messageTraits:
    some-trait:
      headers:
        type: object
        properties:
          t-req-header:
            type: string
[...]

PR #517, opened 2021-03-16
fix: clarify that traits overwrite the objects they are applied to

Fannon:
**Description**
This PR will add some clarification (basically a warning) that a traits properties will overwrite the properties of the object it is applied to. This is potentially counter-intuitive. For more explanations, please see #505 and this comment: https://github.com/asyncapi/spec/issues/505#issuecomment-797933700 

**Related issue(s)**
See also #505


spec/asyncapi.md
```diff
@@ -657,7 +657,7 @@ Field Name | Type | Description
...
-<a name="operationObjectTraits"></a>traits | [[Operation Trait Object](#operationTraitObject) &#124; [Reference Object](#referenceObject) ] | A list of traits to apply to the operation object. Traits MUST be merged into the operation object using the [JSON Merge Patch](https://tools.ietf.org/html/rfc7386) algorithm in the same order they are defined here.
+<a name="operationObjectTraits"></a>traits | [[Operation Trait Object](#operationTraitObject) &#124; [Reference Object](#referenceObject) ] | A list of traits to apply to the operation object. Traits MUST be merged into the operation object using the [JSON Merge Patch](https://tools.ietf.org/html/rfc7386) algorithm in the same order they are defined here. The resulting object MUST be a valid [Operation Object](#operationObject). **WARNING**: Please be aware that due to the defined order of merge patching, a trait will actually *overwrite* the properties of the object they are applied to.
```
Fannon, line 660: This description might get a bit too long and is partially duplicated with the Message Trait Object description.

One option would be to introduce a dedicated section in the spec where both can point to. This section could then also carry more information and explain the traits in more detail.

smoya: I wonder if it would make sense to clarify what https://github.com/asyncapi/spec/issues/505 estates: as the [JSON Merge Patch](https://tools.ietf.org/html/rfc7386) mentions, arrays are fully replaced as a whole value (by the ones in the trait).

Fannon: Yes, this should also be clarified probably.

If we need further clarification, having a dedicated section where we can refer to would make all the more sense.

Btw. I've created an alternative PR which has such a section, but also redefines the behavior: #532

Fannon:
@smoya : This PR could indeed be deprecated in favor of #532 
But this is up to you, which approach you think makes more sense.

PR #518, opened 2021-03-17
docs: clarify schemaFormat and payload

### LINKED ISSUE #503: Clarifications in documentation of `schemaFormat` and `payload`
**Describe the bug**
The documentation for `schemaFormat` : https://www.asyncapi.com/docs/specifications/2.0.0#messageObject states:

```
A string containing the name of the schema format used to define the message payload. If omitted, implementations should parse the payload as a Schema object.
```

My understanding following : https://asyncapi.slack.com/archives/C34F2JV0U/p1614177165056400 is that the `schemaFormat` addition of `+json/+yaml` is primarily valuable when using `$ref` for remote files, to state the format of the schema (so that the parser knows how to handle the file once resolved).

When inlining the schema, while I guess for YAML AsyncAPI document you could put a JSON object - you can't for JSON AsyncAPI as it would not be valid.

My proposal/requested solution is additional help in the documentation for inlining. Something like:
```
When inlining the schema, you should omit the `+json/+yaml` as this is only needed to inform an implementation on the format of a schema by remote reference.
```

I also think the `payload` needs clarifying : 
```
Definition of the message payload. It can be of any type but defaults to Schema object.
```

to include:
```
It should match the format provided in schemaFormat.
```

as initially myself and colleagues thought that meant you could inline the value as a string, and `schemaFormat` was just telling the parser to parse that string as the given format.

nictownsend:
**Description**

- Clarify schemaFormat and how it relates to payload and `$ref`

**Related issue(s)**
Resolves #503


spec/asyncapi.md
```diff
@@ -998,9 +998,9 @@ Describes a message received on a given channel and operation.
...
+<a name="messageObjectSchemaFormat"></a>schemaFormat | `string` | A string containing the name of the schema format used to define the message payload. If omitted, implementations should parse the payload as a [Schema object](#schemaObject). When the payload is defined using a `$ref` to a remote file, the schema format should include the file encoding type to allow implementations to parse the file correctly.<br/><br/>Check out the [supported schema formats table](#messageObjectSchemaFormatTable) for more information. Custom values are allowed but their implementation is OPTIONAL. A custom value MUST NOT refer to one of the schema formats listed in the [table](#messageObjectSchemaFormatTable).
```
derberg, line 1003: > the schema format should include the file encoding type

what do you mean?

derberg:
@fmvilas assigning you as you were involved in the issue

spec/asyncapi.md
fmvilas: I'd not use "should" here but instead "it is recommended". We want to keep it flexible. There are ways for tooling to guess the content type. Linters like [Spectral](https://github.com/stoplightio/spectral/) can add rules for authors to optinally make it easier for tooling.

My suggestion:

```suggestion
<a name="messageObjectSchemaFormat"></a>schemaFormat | `string` | A string containing the name of the schema format used to define the message payload. If omitted, implementations should parse the payload as a [Schema object](#schemaObject). When the payload is defined using a `$ref` to a remote file, it is RECOMMENDED the schema format includes the file encoding type to allow implementations to parse the file correctly. E.g., adding `+yaml` if content type is `application/vnd.apache.avro` results in `application/vnd.apache.avro+yaml`.<br/><br/>Check out the [supported schema formats table](#messageObjectSchemaFormatTable) for more information. Custom values are allowed but their implementation is OPTIONAL. A custom value MUST NOT refer to one of the schema formats listed in the [table](#messageObjectSchemaFormatTable).
```

spec/asyncapi.md
```diff
@@ -998,9 +998,9 @@ Describes a message received on a given channel and operation.
 Field Name | Type | Description
 ---|:---:|---
 <a name="messageObjectHeaders"></a>headers | [Schema Object](#schemaObject) &#124; [Reference Object](#referenceObject) | Schema definition of the application headers. Schema MUST be of type "object". It **MUST NOT** define the protocol headers.
-<a name="messageObjectPayload"></a>payload | `any` | Definition of the message payload. It can be of any type but defaults to [Schema object](#schemaObject).
+<a name="messageObjectPayload"></a>payload | `any` | Definition of the message payload. It can be of any type but defaults to [Schema object](#schemaObject). It must match the schema format, including encoding type - e.g Avro should be inlined as either a YAML or JSON object NOT a string to be parsed as YAML or JSON.
```
fmvilas, line 1001: One suggestion regarding formal language:

```suggestion
<a name="messageObjectPayload"></a>payload | `any` | Definition of the message payload. It can be of any type but defaults to [Schema object](#schemaObject). It MUST match the schema format defined in [`schemaFormat`](#messageObjectSchemaFormat), including encoding type - e.g Avro should be inlined as either a YAML or JSON object NOT a string to be parsed as YAML or JSON.
```

And why can't it be inlined as a string? I think it's fine to have a YAML Avro document inlined in a JSON AsyncAPI file:

```json
{
  "payload": "type: record\nfields:\n  - name: displayName\n    type: string"
}
```

I mean, not the best UX but may be fine for tooling-generated AsyncAPI files that are not meant to be consumed by humans.

fmvilas (changes requested):

spec/asyncapi.md
nictownsend: > And why can't it be inlined as a string? I think it's fine to have a YAML Avro document inlined in a JSON AsyncAPI file

This was the reason for raising the PR - the JS parser treats a `String` as a primitive Avro schema (the string is the name of a primitive type e.g `"string"`). And as such, the inlined schema is not able to be parsed.

If however the thought is that it should be acceptable, I'm happy to raise an issue instead on the parser to support it

fmvilas: Oh yeah! You're completely right. A string is a valid Avro schema and is a valid JSON Schema too. Thanks for noticing!

spec/asyncapi.md
fmvilas: @derberg do you know what it means now?

fmvilas (approved):
LGTM 👍

spec/asyncapi.md
derberg: @fmvilas I'm not good person to get this question as I talked with Nic in the past about this topic, so I know the background.

only worth noticing that now it sounds like someone could have a yaml file and point to schemas in json files and specify it in the `schemaFormat` and our parsers should support it, and I don't think they do now.

maybe ask 3rd person

fmvilas: It's exactly how it should work. If it's not working like this in the parser, it's an issue of the parser.

PR #530, opened 2021-04-19
fix: add requirement for updating JSON Schema when contributing

derberg:
Updating contribution guide as there is a bug that makes only JS Parser a requirement for introducing changes in the spec. JSON Schema is moved to a separate repo and it also must be updated as JS Parser depends on it. So no other option here, you have to accept this PR @fmvilas 😄


fmvilas (approved):

derberg:
It is actually not included in the release. This was an unintentional release caused by the release automation initial run. 1.0 version of the spec was released long time ago. The released mentioned by the bot was removed.

derberg:

Issue #475, opened 2021-01-08
Eanble the association of channels with servers/brokers


bruno-baloi-mulesoft:
There is a significant need for the ability to design Async APIs that perform multiple activities.

For instance consuming events over one type of transport/server (i.e. MQTT) and subsequently generating events over other types of transports (i.e. AMQP). Typically if an Async API implements some type of event choreography that occurs quite often. Additionally there can be APIs that act as Transport bridges and applying certain types event filtering or processing.

In these types of scenarios the Async API spec will need to define the servers, and the channels need to have a mechanism whereby they can indicate what server they are bound to.

fmvilas:
Duplicate of #255?

bruno-baloi-mulesoft:
Thx Fran. So can we expect the update to the spec then ?

fmvilas:
I'm working on the roadmap for 2.1.0. Don't have a date yet but will update here as soon as we know. Thanks for your patience.

Issue #255, opened 2019-08-08
Given a channel name, how do you know which broker (server) should be used?


treeder:
Some things to consider:

* Multiple servers could have the same channel name - I subscribe to "orders" channel on server A and publish to "orders" on server B
* The websocket example from #253 , you the message has to pass the channel name along in the message, but which broker/server should it go to? 
* A message provides a channel to reply (request/reply), again, which broker?

jmaister:
Same issue here.

The "channels" section has already a "channel.operation.bindings" field, which does not apply for this problem.

As I have to make the code work, for the moment I added a non-official property called "channel.operation.binder" to the channel to point to the ID of the server.

    asyncapi: '2.0.0'
    ...
    
    servers:
      nats1:
        url: nats://10.10.10.10:4222
        description: Primary NATS server
        protocol: nats
      kafka1:
        url: 10.10.10.10:9092
        description: Primary Kafka server
        protocol: kafka
    
    channels:
      multi_convert_request:
        subscribe:
          binder: nats1
          message:
            ...
    
      multi_convert_response:
        publish:
          binder: kafka1
          message:
            ...

fmvilas:
Yeah, we'll work it out in future versions. In the meantime, make sure you use compatible property names so your AsyncAPI document is still valid. You can just rename your `binder` property to `x-binder` and everything should be fine.

rvowles:
yay this is being added, we just had a question about this in our team

GeraldLoeffler:
Another use case are all kinds of "protocol adapter" apps, i.e., apps that read from server S1 using protocol P1 and forward messages to server S2 using protocol P2.

For example: Imagine an app that receives messages from an IBM MQ server and sends them to an MQTT server: it has a channel with a publish operation that must be linked to the IBM MQ server, and another channel with a subscribe operation that must be linked to the MQTT server. This should be expressed in one AsyncAPI document. Because if it were split into two separate AsyncAPI documents (one with IBM MQ as the server, the other with MQTT as the server) then the fact that the app essentially just routes incoming IBM MQ messages to an MQTT server is lost.

smoya:
The concept behind https://github.com/asyncapi/spec/issues/244 shows a more distributed-systems-oriented approach where servers can be grouped (e.g. clusters). Then Operations would just specify in which group(s) or cluster(s) can be found.

derberg:
This is now supported by 2.2 release of the spec 🥳

Issue #244, opened 2019-07-25
Allow grouping topics across multiple broker clusters


lwielek:
**Scenario**
A single application connects to 2 (or more) broker clusters.
It publishes/subscribes to topics maintained by both broker clusters.
Topics are not being shared among brokers - a topic that exists on broker 1, does not exist on broker 2.

**Problem**
It is not possible to document the application's APIs in a single Async-API specification.
From the spec, it is not possible to know, which topic is maintained on which broker cluster.

**Proposal**
Extend the specification with 2 additional fields.
- `server.group` (string) - all servers that belong to a single cluster can be identifier by the group (cluster) name. Optional field. If not present, the server is assigned to a *default* group.
- `topic.serverGroup` (string) - the broker group (cluster) name, where the topic exists. Optional field. If not present, the *default* group is assumed.

Example:
```
servers:
  - url: api.streetlights.smartylighting.com:1234
    scheme: mqtt
    description: API broker
    group: api
  - url: audit.streetlights.smartylighting.com:1234
    scheme: mqtt
    description: Audit broker
    group: audit
  
topics:
  event.{streetlightId}.lighting.measured:
    serverGroup: api
    publish:
      $ref: '#/components/messages/lightMeasured'
  event.lighting.audit:
    serverGroup: audit
    publish:
      $ref: '#/components/messages/auditMessage'
```

**Workaround**
Possible workaround is to create 2 separate specifications - separate for broker 1 topics, and broker 2 topics.

derberg:
I'm paying out with websocket atm. I have an API that has 2 different connection points, we can call it private and public. One server is opened, and the other one is secured with token. As typical Websocket, there is one channel, one endpoint. There are messages in the API that can go only to private server.

So I don't even need to wire a topic (channel) with the server, but actually a message with the server.

Possible mitigation, I can have 2 AsyncAPI files of course that share messages/schemas with `$ref`. Nevertheless would be nice to look more into the subject

derberg:
possible with AsyncAPI 2.2 🥳

Issue #513, opened 2021-03-15
Define a release cadence


fmvilas:
### Description

Currently, there's no information on what would be the next version. I don't even know myself, to be honest. Many people ask me when we'll release the next version of the spec and what version would it be. E.g., 2.0.1, 2.1.0, or maybe 3.0.0?

Therefore, to make sure we always set the right expectations, I propose we define a fixed release cadence. Below is my proposal but feel free to leave yours as well.

### Release cadence proposal

* **Continuous releases for patch versions**. We should not be waiting to fix a typo, a wrong example, or something that doesn't change the spec but just corrects the specification text.
* **3 months cadence for minor and major versions**. We set fixed dates for the next releases. People know exactly when a new version of the spec will be released. We release whatever we have. If what's being released has breaking changes, it's a major version, otherwise, it's a minor one. We work on long-lived branches that have the name of the target release month. E.g., `2021-05`.

### Related issues

#463

aayushmau5:
I like these ideas. As for the release version, for major release we should go with 2.1.0 and minors or patches 2.0.1. Not really sure if it's in complaint with Semantic Versioning.

WaleedAshraf:
Related/Duplicate ticket: https://github.com/asyncapi/spec/issues/463

fmvilas:
Thanks, Waleed! Added it to the description. Although, please take into account **this one is only about the cadence**. As in how often should we be releasing new spec versions.

fmvilas:
@aayushmau5 we're already following Semver :) This issue is about the release cadence.

lornajane:
I suggest quarterly releases, it takes the pressure off because if something isn't quite ready or doesn't make the cut, I don't think anyone could be upset about waiting 12 weeks, but half a year can seem too long and cause us to wait for "just one more thing" .... forever. I also suggest the dates are set for mid quarter, so we avoid other end-of-quarter noise and would therefore release in February, May, August and November every year.

fmvilas:
Totally agree, @lornajane. Changing my proposal to every quarter.

derberg:
now when I'm trying to imagine the flow here I think it was not the best idea to move spec's JSON Schema away from `spec` repo. Now it will be hard to make PRs, and imho PR with change in the spec should also update JSON Schema if only possible of course. Why? because after merging to `2021-05-01` branch we should generate a release candidate artifact so others can play with it. 

[...]

PR #531, opened 2021-04-27
feat: assign channels to servers

GeraldLoeffler:
---
title: "Assign channels to servers through their names"
---

Add `servers` field to channel items, so that channels can be assigned to a sub-set of servers using the names of those servers. (A server's name is the string key used for defining that server in the top-level `servers` object.)

Solves the issues discussed in the related issues:

---
**Related issue(s):**

#244 
#255 
#475 
---


smoya:
I like the approach of using **labels** here. In fact, It made me think about Kubernetes labels and selectors. 
Perhaps, we could move one step further and add the same `metadata` concept for servers so labels can be a map of string => string as well. In this way we could also tag servers with labels such as `env: production` or similar.

fmvilas:
@smoya I think going the same path as k8s is a bit of a premature optimization for now IMHO. AFAIK, nobody has asked for a more sophisticated mechanism, and adding it will introduce potentially unnecessary complexity to the spec.

derberg:
@GeraldLoeffler could you clarify more about why it is Server and Channel relation and not Server and Message relation? There is a use case that one channel can have multiple messages on a given operation (`messages.oneOf`). You can have 2 servers with the same channel, but some messages are available only on 1 server. What would be the possible problem if we have `serverLabel` on the message and not the channel?

Another suggestion from my side would be a bit more modification on the example. Now `labels` kind of duplicate information like server key and protocol. This may cause confusion.

Another thinking exercise is: at first sight `labels` seem to be nifty solution but is it really necessary here, do we have a use case for so many servers that they need some kind of label? wouldn't it be better -> the channel/message just have a `server` object that is a list of servers? we would leverage `$ref` usage here of course? I think it would be more aligned with the spec 🤔 like that users could reuse existing solutions like `$ref` and `traits` to optimize AsyncAPI document and keep it DRY, instead of the `label` system.

Sorry for all these challenges 😅 

Huge additional request, please explicitly mention this PR as a comment in all the linked issues so people that are interested the most are notified.

GeraldLoeffler:
> @GeraldLoeffler could you clarify more about why it is Server and Channel relation and not Server and Message relation? There is a use case that one channel can have multiple messages on a given operation (`messages.oneOf`). You can have 2 servers with the same channel, but some messages are available only on 1 server. What would be the possible problem if we have `serverLabel` on the message and not the channel?

it was simply a question of modelling the most obvious association, the one between server and channel:
- a queue/topic (= channel) is inherently (often statically) bound to a message broker (= server)
- a URL path (= channel) is inherently (typically statically) bound to a HTTP server (= server)
So far in the spec it was assumed that in the scope of a given AsyncAPI document, all channels are available on all servers. I'm trying to relax that requirement.

You are saying that the same channel may exist on several servers, but some messages may only be sent/received via that channel to/from a subset of those servers. True, that can happen too, but i didn't have the need to model that and so it's not in my proposal. If we wanted to include it we would need an optional `serverLabels` field on the message (in addition to the one on the channel).

GeraldLoeffler:
> Another suggestion from my side would be a bit more modification on the example. Now `labels` kind of duplicate information like server key and protocol. This may cause confusion.

agreed, the examples could be better: happy to refine them once we have agreement on the approach.

GeraldLoeffler:
> Another thinking exercise is: at first sight `labels` seem to be nifty solution but is it really necessary here, do we have a use case for so many servers that they need some kind of label? wouldn't it be better -> the channel/message just have a `server` object that is a list of servers? we would leverage `$ref` usage here of course? I think it would be more aligned with the spec 🤔 like that users could reuse existing solutions like `$ref` and `traits` to optimize AsyncAPI document and keep it DRY, instead of the `label` system.

your proposal is to bind channels to servers by explicitly listing all servers to which a channel applies. This is certainly possible, but cumbersome in the typical case of having many environments. For example, for a protocol translator from `mqtt` to `jms`:
```
servers:
 mqtt-dev:
    url: ...
  mqtt-test:
    url: ...
  mqtt-prod:
    url: ...
  jms-dev:
    url: ...
  jms-test:
    url: ...
  jms-prod:
    url: ...

from-channel-1:
  servers:
    - $ref "...mqtt-dev"
    - $ref "...mqtt-test"
    - $ref "...mqtt-prod"
  ...
from-channel-2:
  servers:
    - $ref "...mqtt-dev"
    - $ref "...mqtt-test"
    - $ref "...mqtt-prod"
  ...
to-channel-1:
  servers:
    - $ref "...jms-dev"
    - $ref "...jms-test"
    - $ref "...jms-prod"
  ...
to-channel-2:
  servers:
    - $ref "...jms-dev"
    - $ref "...jms-test"
    - $ref "...jms-prod"
  ...
```
versus
```
servers:
 mqtt-dev:
    url: ...
    label: ["from"]
  mqtt-test:
    url: ...
    label: ["from"]
  mqtt-prod:
    url: ...
    label: ["from"]
  jms-dev:
    url: ...
    label: ["to"]
  jms-test:
    url: ...
    label: ["to"]
  jms-prod:
    url: ...
    label: ["to"]

from-channel-1:
  serverLabels: ["from"]
  ...
from-channel-2:
  serverLabels: ["from"]
  ...
to-channel-1:
  serverLabels: ["to"]
  ...
to-channel-2:
  serverLabels: ["to"]
  ...
```
I think this example shows the tradeoff between the two approaches.

smoya:
> @GeraldLoeffler could you clarify more about why it is Server and Channel relation and not Server and Message relation? There is a use case that one channel can have multiple messages on a given operation (`messages.oneOf`). You can have 2 servers with the same channel, but some messages are available only on 1 server. What would be the possible problem if we have `serverLabel` on the message and not the channel?

Would it make sense to add, instead, an extra link between Servers and Operations? This is something we have considered when designing the new Intent-driven API for the Parser(s). See https://github.com/asyncapi/parser-api/blob/master/docs/v1.md#operation.

However, having several resources linking to servers can introduce issues such as mutual exclusion. What if a channel specifies its relation to some servers but their operations are linked to others?

GeraldLoeffler:
Calling all interested parties to please **support one of the following proposals** on this topic:

1. associate channels to servers via simple string labels on servers ([original proposal](https://github.com/asyncapi/spec/pull/531/files) by @GeraldLoeffler, also shown in the 2nd part [here](https://github.com/asyncapi/spec/pull/531#issuecomment-828252634))
2. ditto, but extend the labelling scheme to be name-value pairs exactly as in Kubernetes ([proposal](https://github.com/asyncapi/spec/pull/531#issuecomment-827782537) by @smoya)
3. associate channels to servers by referring to the servers directly, without any labels (proposal by @derberg, shown in the 1st part [here](https://github.com/asyncapi/spec/pull/531#issuecomment-828252634))

In addition, there were proposal to go beyond the association of channels to servers:

4. also associate operations to servers ([proposal](https://github.com/asyncapi/spec/pull/531#issuecomment-828293409) by @smoya)
5. also associate messages to servers ([proposal](https://github.com/asyncapi/spec/pull/531#issuecomment-828238520) by @derberg)

Proposals 4. and 5. are additional associations. So the mechanism (string labels, key-value-pair labels, or direct reference) to link to servers in 4. and 5. should match whatever is chosen to link channels to servers. I therefore propose to cover 4.-5. by a separate PR and focus here on 1.-3. .

Please vote on 1., 2., or 3., and also state your support for 4. and/or 5. (@smoya, @derberg, @fmvilas).

tyazid:
> > @GeraldLoeffler could you clarify more about why it is Server and Channel relation and not Server and Message relation? There is a use case that one channel can have multiple messages on a given operation (`messages.oneOf`). You can have 2 servers with the same channel, but some messages are available only on 1 server. What would be the possible problem if we have `serverLabel` on the message and not the channel?
> 
> Would it make sense to add, instead, an extra link between Servers and Operations? This is something we have considered when designing the new Intent-driven API for the Parser(s). See https://github.com/asyncapi/parser-api/blob/master/docs/v1.md#operation.
> 
> However, having several resources linking to servers can introduce issues such as mutual exclusion. What if a channel specifies its relation to some servers but their operations are linked to others?

Rather consider the association of an operation with a server as optional because it is not the general case. Indeed, an operation should (and it is) be associated with a message which can be exchanged with one or several servers... in my opinion.

DavidBiesack:
See also a similar request for OpenAPI to define reusable sets of parameters (or parameter sets). There may be some ideas presented there that you can use to refine this feature proposal.  I.e. labels are juts one way to define a set of servers (all servers sharing a label constitute a server set), and you want to be able to apply or associate a channel to one or more server sets.

derberg:
@GeraldLoeffler I think it is hard to perform any voting if there are still some unknowns. 

> agreed, the examples could be better: happy to refine them once we have agreement on the approach.

I think good example is necessary for others to follow proposal

> your proposal is to bind channels to servers by explicitly listing all servers to which a channel applies. This is certainly possible, but cumbersome in the typical case of having many environments.

The advantage is that you have validation out of the box, parsing fails if you want to parse outdated reference. With labeling, one can change a label on a channel to a name that is not on the server list. Of course this could also be validated but therefore needs to be specified in the spec so parser know it is mandatory to implement.

---

Can you provide a full example of AsyncAPI file with thes new properties. I'd like to have a closer look on a real use case and see how it works for websocket. We have over representation of people supporting concepts with broker in the middle and I just want to explore how this would work for websocket, if this is of any use actually.

> also associate messages to servers (proposal by @derberg)

Let's drop it. I see this in websockets only and for the server code generation where in fact in my opinion this should be really solved by having 2 separate spec documents, with external common parts

GeraldLoeffler:
@derberg @smoya , @fmvilas : going the path of least resistence here and adopting @derberg's proposal, i.e., associating channels to servers by referring to the servers directly, without any labels. This works and is the simplest possible solution.

Have also added a (i think) convincing example:

```
servers:
  webSocketsDev:
    $ref: "#/components/servers/wsDev"
  webSocketsProd:
    $ref: "#/components/servers/wsProd"
  amqpProd:
    $ref: "#/components/servers/rabbitmqBrokerInProd"
  amqpStage:
    $ref: "#/components/servers/rabbitmqBrokerInStaging"

webUICommandsQueue:
  description: This application publishes WebUICommand messages to this AMQP queue which it had received via WebSockets on another channel
  servers:
    - $ref: "#/components/servers/rabbitmqBrokerInProd"
    - $ref: "#/components/servers/rabbitmqBrokerInStaging"
  subscribe:
    message:
      $ref: "#/components/messages/WebUICommand"
  bindings:
    amqp:
      is: queue
...

components:
  servers:
    rabbitmqBrokerInProd:
      url: staging.gigantic-server.com
      description: Staging server
      protocol: amqp
      protocolVersion: 0.9.1
   rabbitmqBrokerInStaging
      url: api.gigantic-server.com
      description: Production server
      protocol: amqp
      protocolVersion: 0.9.1
    wsDev:
      ...
    wsProd:
      ...
```

Please review https://github.com/asyncapi/spec/pull/531/files and approve.

spec/asyncapi.md
```diff
@@ -640,6 +656,40 @@ subscribe:
 ```
 
 
+Using eplicit references to the servers on which the channel is available:
```
fmvilas, line 659:
```suggestion
Using explicit references to the servers on which the channel is available:
```

spec/asyncapi.md
```diff
@@ -1322,6 +1372,7 @@ All objects defined within the components object will have no effect on the API
 
 Field Name | Type | Description
 ---|:---|---
+<a name="componentsServers"></a> servers | Map[`string`, [Server Object](#serverObject) \| [Reference Object](#referenceObject)] | An object to hold reusable [Server Objects](#serverObject).
```
fmvilas, line 1375: Do we really need this? Why can't we point directly to `#/servers/{name-of-server}`?

spec/asyncapi.md
```diff
@@ -640,6 +656,40 @@ subscribe:
...
+    }
+  }
+}
+```
+
+```yaml
+description: This application publishes WebUICommand messages to this AMQP queue which it had received via WebSockets on another channel
+servers:
+  - $ref: "#/components/servers/rabbitmqBrokerInProd"
+  - $ref: "#/components/servers/rabbitmqBrokerInStaging"
```
fmvilas, line 685: Why not like this:

```suggestion
  - $ref: "#/servers/amqpBrokerInProd"
  - $ref: "#/servers/amqpBrokerInStaging"
```

fmvilas (changes requested):
Left some comments, Gerald. I think we can simplify this even a bit more.

fmvilas:
Also, related to your comment in #513, I think it's too early for this proposal to get into the next release. It's still a Strawman (RFC 0) proposal and I don't think we're giving enough time for people to review and implement support in the parsers.

spec/asyncapi.md
GeraldLoeffler: the only other instances i found in the spec where a field points to another field where when they pointed to component objects: this is the only reason why i've modelled it this way. If you say we can point directly to `#/servers/{name-of-server}` and get the same kind of static validation of the reference then i'm more than happy to make that change. Let me know @fmvilas .

GeraldLoeffler:
> Also, related to your comment in #513, I think it's too early for this proposal to get into the next release. It's still a Strawman (RFC 0) proposal and I don't think we're giving enough time for people to review and implement support in the parsers.

that's very unfortunate, but there's nothing i can do about it other than try to shepherd this through the required process...

spec/asyncapi.md
GeraldLoeffler: i understand this is your proposal to linking directly to `#/servers/{name-of-server}`: but `{name-of-server}` (eg, `amqpBrokerInProd`) in this case would be name of a server in the map of server objects (https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md#servers-object): if you say this works just as well as pointing to a components object then i'm OK with that.

GeraldLoeffler:
i've now implemented the further simplification suggested by @fmvilas , which looks like this:
```
servers:
  rabbitmqBrokerInProd:
    url: staging.gigantic-server.com
    description: Staging server
    protocol: amqp
    protocolVersion: 0.9.1
 rabbitmqBrokerInStaging
    url: api.gigantic-server.com
    description: Production server
    protocol: amqp
    protocolVersion: 0.9.1
  wsDev:
    ...
  wsProd:
    ...

webUICommandsQueue:
  description: This application publishes to this AMQP queue WebUICommand messages which it had received via WebSockets on another channel
  servers:
    - $ref: "#/servers/rabbitmqBrokerInProd"
    - $ref: "#/servers/rabbitmqBrokerInStaging"
  subscribe:
    message:
      $ref: "#/components/messages/WebUICommand"
  bindings:
    amqp:
      is: queue
...
```

I think this is now the simplest possible design that still allows AsyncAPI document authors to assign channels to a sub-set of the defined servers.

@derberg , @fmvilas : please review and merge

GeraldLoeffler:
please dear gods of GitHub, unstale this PR! - I still think this can/should be merged ASAP, because it is fully backwards-compatible: i'm hoping for the September release of AsyncAPI...

derberg:
I think this proposal can reach `Stage 1: Proposal`
- it is already opened for 3 months
- there was a good discussion
- we shared it with the community on all possible channels, including https://www.asyncapi.com/blog/may-2021-at-asyncapi#assigning-channels-to-servers
- I assume @GeraldLoeffler is obviously a champion here that will drive it further, including changes in the JS Parser and JSON Schema
- all are clearly explained, and the use case is pretty obvious. In slack we heard many people asking for such support on channel level

From the definition of Stage 1 all requirements are met imho. Opinions @fmvilas ?

I still have some concerns (non-bloking promotion to Stage 1) with approach that we use refs to Servers Object here @fmvilas. Spec says that `components` are there for reuse purposes, so this can cause confusion that servers are reused without components. Unless, unless this way we want to make sure the user provides `Servers Object`. We set a kind of requirement on the user that to use a specific server on a channel level, all servers need to be listed under `Servers Object`. I don't see why it is needed though. Isn't it an unnecessary restriction? 🤔 Somehow I have a feeling that in case we add Servers on Components, usage of ref definitely makes sense, but using ref on Servers directly sounds weird 🤷🏼 I kinda immediately think it should be just a list of names (or label like in case of initial request)

magicmatatjahu:
I wasn't involved with this proposal from the beginning, but as I read the discussion two things wonder me: 
- what in the case if someone add reference to the server, but from the other AsyncAPI doc, like `$ref: ./asyncapi-two.yaml/servers/production`. In the current proposal it's possible, but there should be only possibility to reference to the servers from current AsyncAPI document, so as @derberg you wrote:

  > The advantage is that you have validation out of the box, parsing fails if you want to parse outdated reference...
  
  We must also implement custom parsing for this case - check if references server is included in the `servers` section.

- why not using the normal server names, without any labels/references, similar like in https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md#security-requirement-object?

  ```yaml
  servers:
    rabbitmqBrokerInProd:
      url: staging.gigantic-server.com
      description: Staging server
      protocol: amqp
      protocolVersion: 0.9.1
   rabbitmqBrokerInStaging
      url: api.gigantic-server.com
      description: Production server
      protocol: amqp
      protocolVersion: 0.9.1
    wsDev:
      ...
    wsProd:
      ...
      
  webUICommandsQueue:
    description: This application publishes to this AMQP queue WebUICommand messages which it had received via WebSockets on another channel
    servers: ['rabbitmqBrokerInProd', 'rabbitmqBrokerInStaging']
    subscribe:
      message:
        $ref: "#/components/messages/WebUICommand"
    bindings:
      amqp:
        is: queue
  ```
  
Maybe there are answers to these questions in the attached issues or in this discussion, but I have not found them.

fmvilas:
Yeah, I think this can be considered a Stage 1 Proposal.

I was actually thinking that it may be better to use server names instead of `$ref`. The problem with $ref is that we lose the name of the server when it's dereferenced.

For instance:

```yaml
asyncapi: 2.1.0
servers:
  serverA:
    url: ws://server-a.com
  serverB:
    url: ws://server-b.com

channels:
  mychannel:
    servers:
      - $ref: '#/servers/serverA'
```

becomes

```yaml
asyncapi: 2.1.0
servers:
  serverA:
    url: ws://server-a.com
  serverB:
    url: ws://server-b.com

channels:
  mychannel:
    servers:
      - url: ws://server-a.com
```

As you can see, in the second example, we lose the `serverA` identifier of the server.

I realized about this while working on Glee. In Glee, you can use a `x-servers` extension (not documented yet) to define which servers you want to associate with the channel. It's user-configurable too so having the `serverA` identifier is necessary.

So, should we instead use plain strings to refer to the servers? Example:

```yaml
asyncapi: 2.1.0
servers:
  serverA:
    url: ws://server-a.com
  serverB:
    url: ws://server-b.com

channels:
  mychannel:
    servers:
      - serverA
```

magicmatatjahu:
@fmvilas Great minds think the same 😄

GeraldLoeffler:
> So, should we instead use plain strings to refer to the servers?

string-references would not be validated at design-time by a YAML editor like `$ref`s would, correct?

fmvilas:
> string-references would not be validated at design-time by a YAML editor like `$refs` would, correct?

`$ref` is not part of YAML specification so I guess a YAML editor would behave the same either with `$ref` and plain string references.

magicmatatjahu:
@GeraldLoeffler won't be, but with using `$ref` there are problems described by me and Fran in previous comments:

- how to check if given server is in current AsyncAPI doc?
- what in case when someone define plain server in the `channel/servers` field, without ref? 

So at the end we still need custom validation.

fmvilas:
I'm confused. How is a YAML editor helping you with `$ref`s? I'm assuming you're referring to specific OpenAPI/AsyncAPI/JSON Schema editors, right? If that's the case, these editors would have to implement support for validation at design time. Just the same way they implemented support for `$ref`.

GeraldLoeffler:
> I'm confused. How is a YAML editor helping you with `$ref`s? I'm assuming you're referring to specific OpenAPI/AsyncAPI/JSON Schema editors, right? If that's the case, these editors would have to implement support for validation at design time. Just the same way they implemented support for `$ref`.

VSCode, for example, with the Red Hat YAML extension, understands YAML and JSON Reference and JSON Pointer, so it knows if a `$ref` is correct, and if it is it shows it as a hyperlink that you can command-click to follow:
yaml-with-ref

derberg:
+1 for server names instead of $refs because of dereferencing. I don't really get why `how to check if given server is in current AsyncAPI doc?` would be a problem. 

and yes, there will have to be a custom validation added to the parser to check if the name added in the channel server matches the server from servers. We already perform such custom validation for server security, if it matches security from components.

the lack of support in YAML editor would be solved once we invest more time in https://github.com/asyncapi/vs-asyncapi-preview. Anyway, not related to this proposal imho.

magicmatatjahu:
> I don't really get why `how to check if given server is in current AsyncAPI doc?` would be a problem.

If we will use the name of server, it won't be a problem, but it will be with references. After dereferencing you will end with object of the server (without name) and additionally as Fran wrote, you will be able  write the plain server in the `servers` field of channel, so there is only one possibility to check that given server is defined in current doc... by deeply checking the entire server shape (or checking the url). In the string case, you only check the array of server names.

GeraldLoeffler:
changed this proposal to refer to servers by name rather than by ref (`$ref$).

fmvilas (changes requested):

GeraldLoeffler:
@fmvilas : made change requested by you and newly fetched from upstream (`asyncapi:master`)

GeraldLoeffler:
@fmvilas : this PR was still in state "changes requested" (by you), but **i have already addressed all requested changes**: i've therefore **re-requested your review**. In other words: i'm done with this and i think it addresses all requests and comments by you and @derberg and @magicmatatjahu .

spec/asyncapi.md
```diff
@@ -562,7 +562,8 @@ Describes the operations available on a single channel.
 Field Name | Type | Description
 ---|:---:|---
 <a name="channelItemObjectRef"></a>$ref | `string` | Allows for an external definition of this channel item. The referenced structure MUST be in the format of a [Channel Item Object](#channelItemObject). If there are conflicts between the referenced definition and this Channel Item's definition, the behavior is *undefined*.
-<a name="channelItemObjectDescription"></a>description | `string` | An optional description of this channel item. [CommonMark syntax](https://spec.commonmark.org/) can be used for rich text representation.
+<a name="channelItemObjectDescription"></a>description | `string` | An optional description of this channel item. [CommonMark syntax](http://spec.commonmark.org/) can be used for rich text representation.
```
fmvilas, line 565:
```suggestion
<a name="channelItemObjectDescription"></a>description | `string` | An optional description of this channel item. [CommonMark syntax](https://spec.commonmark.org/) can be used for rich text representation.
```

fmvilas (changes requested):
Minor thing. Otherwise 👍

GeraldLoeffler:
> Minor thing. Otherwise 👍

requested fix accepted!

magicmatatjahu (approved):

derberg:
@GeraldLoeffler all good from my side here but PR is against `master`, so please change base branch to release branch and rebase

derberg:
as for the other PRs, JSON Schema merged and RC produced. We are missing full implementation in parser -> https://github.com/asyncapi/parser-js/pull/364. So far we are at Stage 1 here

GeraldLoeffler:
@derberg : rebased to release branch - done

spec/asyncapi.md
```diff
@@ -563,6 +563,7 @@ Field Name | Type | Description
 ---|:---:|---
 <a name="channelItemObjectRef"></a>$ref | `string` | Allows for an external definition of this channel item. The referenced structure MUST be in the format of a [Channel Item Object](#channelItemObject). If there are conflicts between the referenced definition and this Channel Item's definition, the behavior is *undefined*.
 <a name="channelItemObjectDescription"></a>description | `string` | An optional description of this channel item. [CommonMark syntax](https://spec.commonmark.org/) can be used for rich text representation.
+<a name="channelItemObjectServers"></a>servers | [[Server Object](#serverObject)] | The servers on which this channel is available, specified as an optional unordered list of names (string keys) of [Server Objects](#serverObject) defined in the [Servers Object](#serversObject) (a map). If `servers` is absent or empty then this channel must be available on all servers defined in the [Servers Object](#serversObject).
```
derberg, line 566:
```suggestion
<a name="channelItemObjectServers"></a>servers | [`string`] | The servers on which this channel is available, specified as an optional unordered list of names (string keys) of [Server Objects](#serverObject) defined in the [Servers Object](#serversObject) (a map). If `servers` is absent or empty then this channel must be available on all servers defined in the [Servers Object](#serversObject).
```

I think this type is leftover from the initial proposal where we advised that `servers` is an array of objects or references, while now it is actually an array of strings

derberg (commented):
A reference implementation in JSON Schema and Parser are already released in individual release candidates. I changed this proposal to Stage 3.

There is one error IMHO here in this PR, please have a look at the individual comment to markdown file cc @fmvilas @magicmatatjahu

spec/asyncapi.md
magicmatatjahu: I agree! In addition, the PR description would have to be changed, as it still points to the old idea related to labels.

derberg:
@GeraldLoeffler fyi we cannot make fixes on our own as you did PR from some kind of org-level fork and we do not have rights as maintainer to push to this PR.

GeraldLoeffler:
> @GeraldLoeffler fyi we cannot make fixes on our own as you did PR from some kind of org-level fork and we do not have rights as maintainer to push to this PR.

i accepted your suggestion and gave you write access to my repo.

derberg:
@GeraldLoeffler perfect, the only part that is missing is updated description that reflects what is actually implemented

GeraldLoeffler:
> @GeraldLoeffler perfect, the only part that is missing is updated description that reflects what is actually implemented

done

derberg (approved):
👏🏼

fmvilas:
Super happy to see thins merged 👏

Issue #536, opened 2021-05-14
Work on 2.1 release


derberg:
Release 2.1 is scheduled for June -> https://github.com/asyncapi/spec/issues/513

Release branches:
https://github.com/asyncapi/spec/tree/2021-06-release
https://github.com/asyncapi/asyncapi-node/tree/2021-06-release
https://github.com/asyncapi/parser-js/tree/2021-06-release

Contribution guide:
https://github.com/asyncapi/spec/blob/master/CONTRIBUTING.md

Things to do for the release:
- https://github.com/asyncapi/website/issues/86 - @aayushmau5
- automation setup. Release candidates artifacts publishing - @derberg
- social media communication - @derberg 
- release notes - @derberg 
- press release - @Barbanio 
- review docs and examples and update if needed
- draft release process basing on this first experience - @derberg 
- review proposals that were already there for some time and people got familiar with proposed solutions, help push what is possible towards `Stage 3: Accepted`

fmvilas:
I created a new milestone to track all the PRs that will be included in the 2.1.0 release: https://github.com/asyncapi/spec/milestone/17.

fmvilas:
We should be setting a deadline date to include changes in the spec for this release. What about May 30?

derberg:
@fmvilas you think we need it? if we have automation in place before the end of May and also I work on a release notes simultaneously and not leave it until the end, we should be able to accept things that come in later, I think. I think better to think about any deadline once we see there are enough features and then we say like "last 2 weeks" or something like that/

Thoughts?

I still dream about pushing https://github.com/asyncapi/spec/issues/377 😅

fmvilas:
Haha! I think it's important we communicate clearly that we're not just randomly choosing what goes in and what doesn't. A deadline might help.

derberg:
The last time people asked I said that whatever reaches `Stage 3: Accepted` gets in. And it is on us to make sure that there is a setup that supports it.

derberg:
cc @dalelane @magicmatatjahu 

I'm tagging you folks as you should keep a closer look at this issue to learn about new additions if you want to add support to other tools you mentioned in the past, like java template or react component

derberg:
draft PR with release notes opened https://github.com/asyncapi/website/pull/260

@fmvilas I know you did an initial review of changes done between 2.0 tag and the latest master. Please provide the list of changes here or just contribute directly to the release notes article

dalelane:
[...]

PR #539, opened 2021-05-19
chore: add release pipeline for the spec

derberg:
This PR introduces release automation for the spec using conventional commits spec and following the release cadence agreement.

assumptions:
- all changes for release happen in release branch and every change in release branch produces a prerelease
- merge from release into master branch triggers normal release, like minor or major
- it is possible to merge from just a fork with for example `fix` PR and trigger patch release

See also: https://github.com/asyncapi/spec/issues/536

---

Commit like `fix: spelling mistake in Operations object description` to master branch

Screenshot 2021-05-19 at 13 43 10

Results in a patch release (base was 2.0.1):
Screenshot 2021-05-19 at 13 45 21

Commit like `feat: add something new in some non-release branch` added to branch that doesn't't follow pattern with `-release` suffx

Screenshot 2021-05-19 at 14 36 21

did not trigger a release

Commit like `feat: add something new in release branch` is a branch that has `2021-06-release` suffix 

Screenshot 2021-05-19 at 15 56 57

Results in proper pre release `v2.1.0-2021-06-release.1`

Screenshot 2021-05-19 at 15 55 19

Another commit to release branch `feat: add something more in release branch` doesn't close bump from 2.1 to 2.2 but proper bump of suffix from 1 to 2

Screenshot 2021-05-19 at 16 02 47

### DIFF
--- a/.releaserc
+++ b/.releaserc
@@ -1,11 +1,11 @@
 ---
 branches:
 - master
-- name: .+release
+- name: 2021-06-release
   prerelease: true
 plugins:
 - - "@semantic-release/commit-analyzer"
-  - preset: conventionalcommits
+  - preset: conventionalcommits 
 - - "@semantic-release/release-notes-generator"
   - preset: conventionalcommits
 - - "@semantic-release/github"
--- a/.github/workflows/.releaserc
+++ b/.github/workflows/.releaserc
@@ -1,0 +1,11 @@
+---
+branches:
+- master
+- name: .+release
+  prerelease: true
+plugins:
+- - "@semantic-release/commit-analyzer"
+  - preset: conventionalcommits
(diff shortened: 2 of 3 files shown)


.releaserc
```diff
@@ -0,0 +1,11 @@
+---
+branches:
+- master
+- name: 2021-06-release
```
derberg, line 4: yes, we need to be hardcoded here 🤷🏼 believe me or not, this is also strange for me but `semantic-release` package doesn't support regex here (I tried everything). I will make sure this step is part of the release process procedure

magicmatatjahu (approved):

fmvilas (approved):
I like it. It's exactly what I'd do manually but automated 💯

derberg:

derberg:

Issue #540, opened 2021-05-19
examples should be maintained on master only for latest


derberg:
We have all examples here https://github.com/asyncapi/spec/tree/master/examples but spec 2.0.0 is only on master. Examples should also in my opinion be there for 2.0.0 and for other versions user can check out given tag

Thoughts?

derberg:
It was already fixed but somehow comment from PR -> `Fixes #540` -> did not close it

PR #548, opened 2021-05-31
docs: separate streetlight examples for Kafka and MQTT

dalelane:
Main motivation of this was to provide a place to demonstrate
the new Kafka-specific security schemes introduced in
AsyncAPI 2.1.0. It was also a chance to remove the Kafka
bindings from the otherwise-MQTT sample.

The need for this was discussed in
https://github.com/asyncapi/spec/pull/502

Contributes to: #466

### DIFF
--- a/examples/2.1.0/streetlights-kafka.yml
+++ b/examples/2.1.0/streetlights-kafka.yml
@@ -1,0 +1,163 @@
+asyncapi: '2.1.0'
+info:
+  title: Streetlights Kafka API
+  version: '1.0.0'
+  description: |
+    The Smartylighting Streetlights API allows you to remotely manage the city lights.
+
+    ### Check out its awesome features:
+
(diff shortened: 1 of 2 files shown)


derberg:
Awesome stuff @dalelane 
Thanks 🙏🏼 

@fmvilas wanna have a look? I think we need to solve https://github.com/asyncapi/spec/issues/540 now or after release, it will be hard to review examples long term if we after all create new instances of those + it is kind of inconsistent, that spec is only latest but examples are there for all versions

derberg (approved):

fmvilas (approved):
LGTM 💯

fmvilas:
@derberg indeed it's weird to have examples for all versions but only the latest version of the spec. I'd get rid of the examples of older versions.

derberg:
@fmvilas I think we can merge and I'll create a followup cleanup

PR #550, opened 2021-06-02
fix: clarify support for Boolean JSON Schemas

magicmatatjahu:
---
title: "Mention about true/false JSON Schemas in Schema Object"
---

As Schema Object in spec is defined as superset of the [JSON Schema Specification Draft 07](http://json-schema.org/), it would be nice to specify that the Schema Object itself can be boolean value. I know it results from the definition of draft 07, so I'd like to hear your opinion about the changes.

Description for `true/false` schemas I took from https://github.com/OAI/OpenAPI-Specification/pull/2609/files Thanks to @MikeRalphson 

---

**Related issue(s):**
https://github.com/asyncapi/parser-js/issues/232
Fixes https://github.com/asyncapi/spec/issues/554

**Related PR(s):**
https://github.com/asyncapi/asyncapi-node/pull/63 - fix bug with true/false schemas in JSON Schema of spec
https://github.com/asyncapi/parser-js/pull/311 - handle true/false JSON Schema in parser 
https://github.com/asyncapi/asyncapi-react/pull/367 - handle Boolean JSON Schema in the `react-component` 

---

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -839,7 +839,7 @@ Describes a parameter included in a channel name.
 Field Name | Type | Description
 ---|:---:|---
 <a name="parameterObjectDescription"></a>description | `string` | A verbose explanation of the parameter. [CommonMark syntax](http://spec.commonmark.org/) can be used for rich text representation.
-<a name="parameterObjectSchema"></a>schema | [Schema Object](#schemaObject) &#124; [Reference Object](#referenceObject) &#124; `boolean` | Definition of the parameter.
+<a name="parameterObjectSchema"></a>schema | [Schema Object](#schemaObject) &#124; [Reference Object](#referenceObject) | Definition of the parameter.
 location | `string` | A [runtime expression](#runtimeExpression) that specifies the location of the parameter value. Even when a definition for the target field exists, it MUST NOT be used to validate this parameter but, instead, the `schema` property MUST be used.
 
 This object can be extended with [Specification Extensions](#specificationExtensions).
@@ -1001,7 +1001,7 @@ Describes a message received on a given channel and operation.
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="messageObjectHeaders"></a>headers | [Schema Object](#schemaObject) &#124; [Reference Object](#referenceObject) &#124; `boolean` | Schema definition of the application headers. Schema MUST be of type "object". It **MUST NOT** define the protocol headers.
+<a name="messageObjectHeaders"></a>headers | [Schema Object](#schemaObject) &#124; [Reference Object](#referenceObject) | Schema definition of the application headers. Schema MUST be of type "object". It **MUST NOT** define the protocol headers.
 <a name="messageObjectPayload"></a>payload | `any` | Definition of the message payload. It can be of any type but defaults to [Schema object](#schemaObject).
 <a name="messageObjectCorrelationId"></a>correlationId | [Correlation ID Object](#correlationIdObject) &#124; [Reference Object](#referenceObject) | Definition of the correlation ID used for message tracing or matching.
 <a name="messageObjectSchemaFormat"></a>schemaFormat | `string` | A string containing the name of the schema format used to define the message payload. If omitted, implementations should parse the payload as a [Schema object](#schemaObject). Check out the [supported schema formats table](#messageObjectSchemaFormatTable) for more information. Custom values are allowed but their implementation is OPTIONAL. A custom value MUST NOT refer to one of the schema formats listed in the [table](#messageObjectSchemaFormatTable).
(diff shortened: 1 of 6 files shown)


jonaslagoni:
@magicmatatjahu is this really necessary? When we refer to the Schema object, to me that implicitly say that boolean is allowed? 🤔

magicmatatjahu:
@jonaslagoni Do you mean union type `Schema Object | boolean`? I honestly wonder. We can describe that Schema Object can be boolean (as I did) and won't add a union type.

spec/asyncapi.md
```diff
@@ -1510,7 +1510,7 @@ components:
 
 The Schema Object allows the definition of input and output data types.
 These types can be objects, but also primitives and arrays.
-This object is a superset of the [JSON Schema Specification Draft 07](http://json-schema.org/).
+This object is a superset of the [JSON Schema Specification Draft 07](http://json-schema.org/). This means that Schema Object can be a boolean value. `True` means that the type can be anything, while `false` means that the given schema cannot be defined.
```
jonaslagoni, line 1514: This is somewhat incorrect depending on how you look at it. If we are talking purely JSON Schema, then `true` means everything will be accepted, `false` means everything will be rejected. Not sure how to best describe that in this sentence though 🤔

magicmatatjahu: `then true means everything will be accepted` - `True` means that the type can be anything.
`false means everything will be rejected` - `false` means that the given schema cannot be defined.

I think that both phrases seem to have the same meaning.

The bigger problem to describe is for some parts of spec, because when we will allow making schema as boolean value (as described in spec after merging this PR), some parts, like headers` in message, MUST be of `object` type, not boolean. People may not understand exactly when they will be able to use scheme as boolean. The schema for the Channel Parameters is another example where the boolean schema doesn't make much sense, but if it's a Schema Object then they have possibility to do that...

jonaslagoni: > `then true means everything will be accepted` - `True` means that the type can be anything.
> `false means everything will be rejected` - `false` means that the given schema cannot be defined.

Hehe, I see your point, but it is two different perspectives, the one I proposed is from a validation point of view, the other is from a definition point of view. - and as JSON Schema is validation, that is why I proposed to change the wording 🙂 

> The bigger problem to describe is for some parts of spec, ...

From my perspective, it is not a problem, as soon as you say it can be a `Schema Object` that entails that it can be a boolean. Why would you want to restrict this? 🤔 IMO, whether it is used or not, that's up to the user, not us 🙂

jonaslagoni: I am starting to wonder if it is intentional that the schema object is not allowed to be a boolean or if that was just how it came to be? 🤔

magicmatatjahu: In spec we use usually definition point of view, in some places also validation POV, but only by purpose, we define that some parts must be in given shape or must be unique. Now I understand why you want describe `true/false` in this way, but as I read `Schema Object` section in spec I see description from definition POV, not validation 😅

magicmatatjahu: > I am starting to wonder if it is intentional that the schema object is not allowed to be a boolean or if that was just how it came to be? 🤔

Maybe @fmvilas forgot about possibility of `true/false` schemas when he described the spec 😅

jonaslagoni: > In spec we use usually definition point of view, in some places also validation POV, but only by purpose, we define that some parts must be in given shape or must be unique. Now I understand why you want describe `true/false` in this way, but as I read `Schema Object` section in spec I see description from definition POV, not validation 😅

Yeea, and I agree we should be consistent. 

IMO it needs to be re-written to use the correct perspective, but it might become unclear for other people then... 🤷

magicmatatjahu: > IMO it needs to be re-written to use the correct perspective, but it might become unclear for other people then...

Ok, let's wait for other comments :)

MikeRalphson:
From our co-incidentally raised PR over at OAS https://github.com/OAI/OpenAPI-Specification/pull/2609/files

magicmatatjahu:
As I can see AsyncAPI not only copied the architecture (of spec), but also full sentences from the specification 😅 

@jonaslagoni Do you agree that

> The empty schema (which allows any instance to validate) MAY be represented by the `boolean` value `true` and a schema which allows no instance to validate MAY be represented by the `boolean` value `false`.

is much better that current description in PR? Will you accept it? As you see, there we have desc from validation POV.

@MikeRalphson will you have any problem so that we can reuse it in the AsyncAPI specification, or do you want us to pay you? 😄

MikeRalphson:
I think we share a common license, as well as having common interests!

spec/asyncapi.md
smoya: > IMO it needs to be re-written to use the correct perspective, but it might become unclear for other people then

I agree we should advocate for consistency since having two perspectives are adding confusion when reading.

> In spec we use usually definition point of view, in some places also validation POV, but only by purpose, we define that some parts must be in given shape or must be unique. Now I understand why you want describe true/false in this way, but as I read Schema Object section in spec I see description from definition POV, not validation 😅

IMHO, definition/describing POV looks more easy to understand for the first reader. Specs are made for describing, tooling around does other things such as validation.

jonaslagoni: > Specs are made for describing, tooling around does other things such as validation.

Then what if you describe what your data (payload) should validate against? 😆

magicmatatjahu: I used the description from https://github.com/OAI/OpenAPI-Specification/pull/2609/files PR:

> The empty schema (which allows any instance to validate) MAY be represented by the `boolean` value `true` and a schema which allows no instance to validate MAY be represented by the `boolean` value `false`.

It's from validation pov, but for me it's a better than previous one.

smoya (approved):
LGTM 🚀 🌔

fmvilas (approved):
LGTM 👍

fmvilas (approved):

magicmatatjahu:
@fmvilas I see that you changed two times the target branch and we end to master. I thought that we will merge it to master and then rebase the `2021-06-release`. Am I right?

fmvilas:
Yes, I got confused at first. If it's just an editorial change and it doesn't change the behavior, it should point to master and should trigger a patch release when merged.

derberg (approved):

derberg:
@magicmatatjahu @fmvilas 
We did not trigger official 2.1 yet using automation.

> If it's just an editorial change and it doesn't change the behavior, it should point to master
yes, but in the future after 2.1, otherwise merge of this fix will trigger 2.1

now it should go to release candidate branch with `fix:` and it will produce `v2.1.0-2021-06-release.3`

derberg:
> yes, but in the future after 2.1, otherwise merge of this fix will trigger 2.1

why? because it already happened twice 😓
semantic-release analyze all commits between tags, so after merging this to master it will analyze all commits since 2.0 tag, and therefore trigger release that we do not want

magicmatatjahu:
Another solution can be go with `docs:`, rebase in pre-release and make patch manually, but merging it to pre-release branch is fine for me.

derberg:
@magicmatatjahu 

> Another solution can be go with docs:

no, unless you provide in commit details `[skip: ci]`

> rebase in pre-release and make patch manually

good luck 😄

magicmatatjahu:
@fmvilas I'm changing the target to prerelease due to problems described by Łukasz :) I need only info from you that you accept our decision, or if you have problems, please write them.

fmvilas:
That's fine. I don't think we too many options right now 😄

magicmatatjahu:
Merged in https://github.com/asyncapi/spec/pull/561 Thanks guys for review!

PR #562, opened 2021-06-22
chore: update topic names in Kafka example

dalelane:
Update the Kafka streetlights example so that it uses channel
names that use characters that are valid for use in Kafka topic
names.

### DIFF
--- a/examples/streetlights-kafka.yml
+++ b/examples/streetlights-kafka.yml
@@ -25,7 +25,7 @@ servers:
 defaultContentType: application/json
 
 channels:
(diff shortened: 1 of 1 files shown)


derberg:
I'm ok with this one as per Slack discussion. Dots in a channel name are still making it a valid URI template. And it is standard for Kafka so I don't see why would we not accept it. Just want @fmvilas to have a look as I don't know all the history, maybe fact that dots work fine is that this is a kind of bug

fmvilas (approved):
LGTM :shipit:

PR #565, opened 2021-06-22
chore: remove `a` tags from spec headings

aayushmau5:
title: "Remove all the `a` tags from headings"
---

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -80,32 +80,32 @@ It means that the [application](#definitionsApplication) allows [consumers](#def
 
 <!-- /TOC -->
 
-## <a name="definitions"></a>Definitions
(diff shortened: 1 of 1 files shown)


aayushmau5:
:thinking:  The `lint` action is failing because of the missing `a` tags?

magicmatatjahu:
@aayushmau5 Instead of removing tags from spec, you can delete them in the https://github.com/asyncapi/website/blob/master/components/navigation/DocsNavItem.js when page is rendered.

aayushmau5:
oh, never really thought about that. I'm going to try that approach.

aayushmau5:
Will be closing this PR in favor of https://github.com/asyncapi/website/pull/283

PR #566, opened 2021-06-23
ci: add step to remove toc from spec in action

aayushmau5:
---
title: "Add step to remove Table of Contents from Spec file"
---

The spec file contains `Table Of Contents`, which would be good to strip out from the website. This action step will remove the TOC from the spec file.


derberg (approved):
This action is getting complex 😅 
Well done anyway, I don't see any other better way to handle that. For now it is perfect!

PR #573, opened 2021-06-25
docs: update websocket example with new example props

derberg:
as a followup of changes proposed by Laurent

### DIFF
--- a/examples/websocket-gemini.yml
+++ b/examples/websocket-gemini.yml
@@ -111,7 +111,9 @@ components:
       payload:
         $ref: '#/components/schemas/market'
       examples:
-        - payload:
+        - name: updateMessage
+          summary: Example of an update message hat contains a change in price information.
+          payload:
             type: update
             eventId: 36902233362
             timestamp: 1619769673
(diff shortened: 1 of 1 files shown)


derberg:
@magicmatatjahu yo, here you have a real example that you can test against changes in markdown and html (react) templates

examples/websocket-gemini.yml
```diff
@@ -111,7 +111,9 @@ components:
       payload:
         $ref: '#/components/schemas/market'
       examples:
-        - payload:
+        - name: updateMessage
+          summary: Example of an update message hat contains a change in price information.
```
magicmatatjahu, line 115: Example of an update message **hat** 

I know that my english isn't good, but `message hat` Wat?! 😄

derberg: yeah, it is another word, like a synonym for `envelope` (you need to read more about CloudEvents). But yeah, I will switch to `that` this time 😆

magicmatatjahu: Thanks for clarification!

magicmatatjahu (approved):

Issue #418, opened 2020-08-01
[FEATURE REQUEST] security scope


cappelaere:
**Is your feature request related to a problem? Please describe.**
Current security scopes are at the full API level.  Given the granularity available, it would be great to have it at the message level.

**Can't it be tackled using specification extensions?**
Since scopes are already defined at securty scheme level, they ought to be used at the message level and used for code generation as well.

**Describe the solution you'd like**
I would like to have the ability to define an array of scopes at the message level that would be necessary for a user to have in order to publish or subscribe.  The array would allow anyone of the scopes to be acceptable,

## Example:
```yaml
smartylighting/streetlights/1/0/action/{streetlightId}/turn/on:
    parameters:
      streetlightId:
        $ref: '#/components/parameters/streetlightId'
    publish:
      operationId: turnOn
      traits:
        - $ref: '#/components/operationTraits/kafka'
      message:
        $ref: '#/components/messages/turnOn'
      security:
        - supportedOauthFlows:
            - streetlights:turnOn
```

derberg:
Hi @cappelaere, could you please format the code sample with markdown so it is easier to read? Thanks

cappelaere:
Updated the example to follow OpenAPI 3.0 to apply security scopes at the message level.

Neverbolt:
I too would very much like this feature to be included in the base specification, since as of now the authorization constructs are not very advanced and being able to specify scopes at an Operation level would allow a much more fine grained control and documentation as well as code generation:

derberg:
@Neverbolt contribution is always welcome, have a look at our contribution guide that explains how to contribute changes to the spec. Most important before PR is to discuss the solution first. Would love to see how you imagine it fixed in the spec.

derberg:
@cappelaere @Neverbolt you should definitely have a look at https://github.com/asyncapi/spec/pull/584

derberg:
@cappelaere @Neverbolt did you have a chance to look at #584 ?

derberg:
I'm closing this one as there are no responses
@cappelaere @Neverbolt please join discussion #584 ? every single voice from the community helps in adding the feature, especially that we have a champion there that drives the topic

PR #584, opened 2021-07-15
feat: add security at the operation level

sekharbans-ebay:
---
feat: Add security at the channel/operation level
---
Description:
We are using AsyncAPI specifications for publishing the subscriber specific details of our topics and associated payloads for our HTTP push use cases.

Our subscribable channels(or topics) are protected by a set of oauth scopes (authorization code or client credentials) - that a potential subscriber needs to be aware of.  

Here is a [sample contract](https://developer.ebay.com/api-docs/master/commerce/notification/asyncapi/authorization_revocation.yaml). 

We are able to accommodate almost **_every detail_** of interest  to an integrator/subscriber (THANK YOU) - with the exception of being able to call out the oauth scopes necessary for the subscription to the topic - i.e.  the security aspect at the channel or rather channel/operation level.  

The current specification is perhaps more aligned with a broker based topology - but ours is a pure push use case (to registered webhooks) and we need an ability to call out any security aspects at the channel operation level.


magicmatatjahu:
@sekharbans-ebay Hello! Please update the title with following [Semantic release](https://github.com/asyncapi/asyncapi-react/blob/master/CONTRIBUTING.md#conventional-commits) to something like:

```
feat: Add security at the channel/operation level
```

Also I think that this PR is related to the https://github.com/asyncapi/spec/issues/418 issue.

sekharbans-ebay:
Hi @magicmatatjahu 

I did change the title as asked - what would be the next steps please? Not sure if I am missing anything else.

Thanks
Sekhar

magicmatatjahu:
@sekharbans-ebay I meant to change the issue title 😅  

> what would be the next steps please

Discussion with core team and community, the best place will be [here](https://github.com/asyncapi/spec/issues/418). You should also inform about your PR and point to the mentioned issue in our slack. By this more people will be interested/involved in this topic.

derberg (changes requested):
@sekharbans-ebay could you provide some fixes here in the PR? 

- look at a failing check, there is an error with PR title
- update PR title that should talk only about operation, not channel
- please extend operation object examples with the new proposed security key
- clarify how should it work if someone provides security on a server object too, what has priority
- I think the operation traits object also needs to be extended

sekharbans-ebay:
> @sekharbans-ebay could you provide some fixes here in the PR?
> 
> * look at a failing check, there is an error with PR title
> * update PR title that should talk only about operation, not channel
> * please extend operation object examples with the new proposed security key
> * clarify how should it work if someone provides security on a server object too, what has priority
> * I think the operation traits object also needs to be extended


Apologies for the delay in responding - all of the review comments have been incorporated except for the 'operation object example' part. I have followed the approach that 'Server' has - where the examples on Security are referenced and provided in great details in an independent section on Security. The Security aspects are complex and require an independent treatment. 

Trust that is acceptable.

derberg:
@sekharbans-ebay I like that you choose the path of being consistent with the spec. In this case though the problem is that we are actually missing an example of security on the server, not a patter that we should follow. To actually find a good example of usage of `security` you have to go through examples and find stuff like https://github.com/asyncapi/spec/blob/master/examples/streetlights-kafka.yml#L22

spec/asyncapi.md
```diff
@@ -762,6 +763,7 @@ Field Name | Type | Description
 <a name="operationTraitObjectOperationId"></a>operationId | `string` | Unique string used to identify the operation. The id MUST be unique among all operations described in the API. The operationId value is **case-sensitive**. Tools and libraries MAY use the operationId to uniquely identify an operation, therefore, it is RECOMMENDED to follow common programming naming conventions.
 <a name="operationTraitObjectSummary"></a>summary | `string` | A short summary of what the operation is about.
 <a name="operationTraitObjectDescription"></a>description | `string` | A verbose explanation of the operation. [CommonMark syntax](https://spec.commonmark.org/) can be used for rich text representation.
+<a name="operationObjectSecurity"></a>security | [[Security Requirement Object](#securityRequirementObject)]| A declaration of which security mechanisms are associated with this operation. Only one of the security requirement objects need to be satisfied to authorize an operation. In cases where Server Security also applies, it is expected the security mechanisms specified at the operation level would take priority and ideally should match or be a subset of that required by the server.
```
derberg, line 766: please change the name of the anchor as it will cause navigation issues

```suggestion
<a name="operationTraitObjectSecurity"></a>security | [[Security Requirement Object](#securityRequirementObject)]| A declaration of which security mechanisms are associated with this operation. Only one of the security requirement objects need to be satisfied to authorize an operation. In cases where Server Security also applies, it is expected the security mechanisms specified at the operation level would take priority and ideally should match or be a subset of that required by the server. 
```

sekharbans-ebay:
> @sekharbans-ebay I like that you choose the path of being consistent with the spec. In this case though the problem is that we are actually missing an example of security on the server, not a patter that we should follow. To actually find a good example of usage of `security` you have to go through examples and find stuff like https://github.com/asyncapi/spec/blob/master/examples/streetlights-kafka.yml#L22

Question on security examples - Since there are different security mechanisms involved, the Security Requirements in the examples  can either be repeated for each of oauth, password or api_key else it seems incomplete. As a reader of the specs , I like what you have for server - i.e. pointer to the relevant security part of the specs.  If it is acceptable I can add - say only oauth as an example(both json/yaml). For example:

operationId: registerUser
summary: Action to sign a user up.
description: A longer description
**security:
  - petstore_auth:
        - write:pets
        - read:pets**
tags:
  - name: user
  - name: signup
  - name: register 
-----
----

derberg:
@sekharbans-ebay it is ok to add example for just one security req, oauth is the best example I think

sekharbans-ebay:
> @sekharbans-ebay it is ok to add example for just one security req, oauth is the best example I think

Thank you. Done.

derberg:
@fmvilas have a look. 
@sekharbans-ebay went through all comments and improved the proposal and therefore for me it is `Stage 1: Proposal` because I'm assuming @sekharbans-ebay is a champion for this proposal because he already went with PR creation.

I think we can label it properly and start pinging community for review

fmvilas:
Yes, let's start pinging the community for review. I like this idea a lot, @sekharbans-ebay 🙌

derberg:

spec/asyncapi.md
```diff
@@ -663,7 +663,8 @@ Field Name | Type | Description
...
 <a name="operationObjectSummary"></a>summary | `string` | A short summary of what the operation is about.
-<a name="operationObjectDescription"></a>description | `string` | A verbose explanation of the operation. [CommonMark syntax](https://spec.commonmark.org/) can be used for rich text representation.
+<a name="operationObjectDescription"></a>description | `string` | A verbose explanation of the operation. [CommonMark syntax](http://spec.commonmark.org/) can be used for rich text representation.
+<a name="operationObjectSecurity"></a>security | [[Security Requirement Object](#securityRequirementObject)]| A declaration of which security mechanisms are associated with this operation. Only one of the security requirement objects need to be satisfied to authorize an operation. In cases where Server Security also applies, it is expected the security mechanisms specified at the operation level would take priority and ideally should match or be a subset of that required by the server.
```
derberg, line 667: > the security mechanisms specified at the operation level would take priority and ideally should match or be a subset of that required by the server

because this is a spec and we need to be super precise, I have to say `ideally` is not the best word to use here. I think this must be a requirement, that security mechanism on operation level must match one of the security mechanisms on the server level. Then we can have a proper validation in place. Thoughts?

sekharbans-ebay: Suggest rewording to simply:  "    the security mechanisms specified at the operation level would take priority."    and omitting the recommendation on the subset altogether. If the invoker does not have access to the operation any component that the operation relies on is automatically shielded.

In some topologies - the operations can use a completely different set of authentication scopes to talk to the broker. For example operation A is exposed with an auth scope a. However the operation could interact with the broker using a different auth scope (defined at the server level) - lets say from an internal set of allowed scopes say 'b' that is granted to the operational component(s) authorized to interact with the broker).

Felix200521 (dismissed):

spec/asyncapi.md
derberg: so we are basically dropping ` and ideally should match or be a subset of that required by the server`? which is much clearer and better for me. Sounds good

derberg:
@fmvilas I'm in a position that we should go forward with this PR to `Draft` stage 2. We gave the community 6months to react and comment. We communicated it a few times on different AsyncAPI channels. I also offline pinged PayPal folks a few times to have a look (I knew that in the past they were asking in slack for something like this). This is coming directly from End User and I like it so much that I'm even happy to help @sekharbans-ebay to provide the initial implementation in AsyncAPI JS Parser + JSON Schema 😄 What is your opinion?

@sekharbans-ebay please enrich this PR with some official example that could land in the `examples` directory. Maybe an official ebay asyncapi file that already incorporated this new feature? Thoughts?

@Felix200521 I see you approved this PR. Can you share a comment on why you like this proposal, what is your use case? It will help us to get a better understanding of why this proposal makes sense for a wider audience.

sekharbans-ebay:
**> @sekharbans-ebay please enrich this PR with some official example that could land in the `examples` directory. Maybe an official ebay asyncapi file that already incorporated this new feature? Thoughts?**

Hi @derberg - I have made the alterations to reword the section. The asyncapi.md does include examples in this PR (relevant section(s) of the specification).
Will share the official asyncapi contract that has this feature incorporated - once (and if) this feature is approved. Trust the examples can wait a bit? Thanks a lot and wish you all a very happy and safe new year!

derberg:
@sekharbans-ebay 

> I have made the alterations to reword the section

but now we do not have a sentence that specifies that operation-level security has priority over server-level security. When I see https://developer.ebay.com/marketplace-account-deletion I assume you do not plan to use `server` at all, thus for your use case it might not be important, but we need to have this clear for others

> Will share the official asyncapi contract that has this feature incorporated - once (and if) this feature is approved

Just keep in mind that we do not need production-ready example here as we know it is added before the actual release. You can update it later. I assume you would use it here -> https://developer.ebay.com/marketplace-account-deletion. So for now you could just add a description that it is not yet an official production-level example, and then update it after release when you are ready.

It is just better to justify the feature and show it to others through a real use case based on a real example that will end up on production.

sekharbans-ebay:
> @sekharbans-ebay
> 
> > I have made the alterations to reword the section
> 
> but now we do not have a sentence that specifies that operation-level security has priority over server-level security. When I see https://developer.ebay.com/marketplace-account-deletion I assume you do not plan to use `server` at all, thus for your use case it might not be important, but we need to have this clear for others
> 
> > Will share the official asyncapi contract that has this feature incorporated - once (and if) this feature is approved
> 
> Just keep in mind that we do not need production-ready example here as we know it is added before the actual release. You can update it later. I assume you would use it here -> https://developer.ebay.com/marketplace-account-deletion. So for now you could just add a description that it is not yet an official production-level example, and then update it after release when you are ready.
> 
> It is just better to justify the feature and show it to others through a real use case based on a real example that will end up on production.

sekharbans-ebay:
> > @sekharbans-ebay
> > > I have made the alterations to reword the section
> > 
> > 
> > but now we do not have a sentence that specifies that operation-level security has priority over server-level security. When I see https://developer.ebay.com/marketplace-account-deletion I assume you do not plan to use `server` at all, thus for your use case it might not be important, but we need to have this clear for others

Agree. Will remove that line. The specification should not dictate a relation between these two - that should be left to the implementation. My original thoughts were based on the broker being a protected entity (i.e. accessible only thru the operations - in this scenario only the operations would need to know the security protocols for the broker which could be independently defined). 



> > > Will share the official asyncapi contract that has this feature incorporated - once (and if) this feature is approved
> > 
> > 
> > Just keep in mind that we do not need production-ready example here as we know it is added before the actual release. You can update it later. I assume you would use it here -> https://developer.ebay.com/marketplace-account-deletion. So for now you could just add a description that it is not yet an official production-level example, and then update it after release when you are ready.
> > It is just better to justify the feature and show it to others through a real use case based on a real example that will end up on production.

Ack. Will provide examples this week.

spec/asyncapi.md
```diff
@@ -702,7 +702,8 @@ Field Name | Type | Description
 ---|:---:|---
 <a name="operationObjectOperationId"></a>operationId | `string` | Unique string used to identify the operation. The id MUST be unique among all operations described in the API. The operationId value is **case-sensitive**. Tools and libraries MAY use the operationId to uniquely identify an operation, therefore, it is RECOMMENDED to follow common programming naming conventions.
 <a name="operationObjectSummary"></a>summary | `string` | A short summary of what the operation is about.
-<a name="operationObjectDescription"></a>description | `string` | A verbose explanation of the operation. [CommonMark syntax](https://spec.commonmark.org/) can be used for rich text representation.
+<a name="operationObjectDescription"></a>description | `string` | A verbose explanation of the operation. [CommonMark syntax](http://spec.commonmark.org/) can be used for rich text representation.
+<a name="operationObjectSecurity"></a>security | [[Security Requirement Object](#securityRequirementObject)]| A declaration of which security mechanisms are associated with this operation. Only one of the security requirement objects need to be satisfied to authorize an operation.
```
fmvilas, line 706:
```suggestion
<a name="operationObjectSecurity"></a>security | [[Security Requirement Object](#securityRequirementObject)]| A declaration of which security mechanisms are associated with this operation. Only one of the security requirement objects MUST be satisfied to authorize an operation. 
```

spec/asyncapi.md
```diff
@@ -760,6 +769,10 @@ This object can be extended with [Specification Extensions](#specificationExtens
 operationId: registerUser
 summary: Action to sign a user up.
 description: A longer description
+security:
+  - petstore_auth:
+        - write:pets
+        - read:pets
```
fmvilas, line 775:
```suggestion
    - write:pets
    - read:pets
```

spec/asyncapi.md
```diff
@@ -801,6 +814,7 @@ Field Name | Type | Description
 <a name="operationTraitObjectOperationId"></a>operationId | `string` | Unique string used to identify the operation. The id MUST be unique among all operations described in the API. The operationId value is **case-sensitive**. Tools and libraries MAY use the operationId to uniquely identify an operation, therefore, it is RECOMMENDED to follow common programming naming conventions.
 <a name="operationTraitObjectSummary"></a>summary | `string` | A short summary of what the operation is about.
 <a name="operationTraitObjectDescription"></a>description | `string` | A verbose explanation of the operation. [CommonMark syntax](https://spec.commonmark.org/) can be used for rich text representation.
+<a name="operationTraitObjectSecurity"></a>security | [[Security Requirement Object](#securityRequirementObject)]| A declaration of which security mechanisms are associated with this operation. Only one of the security requirement objects need to be satisfied to authorize an operation.
```
fmvilas, line 817:
```suggestion
<a name="operationTraitObjectSecurity"></a>security | [[Security Requirement Object](#securityRequirementObject)]| A declaration of which security mechanisms are associated with this operation. Only one of the security requirement objects MUST be satisfied to authorize an operation. 
```

examples/operation-security.yml
```diff
@@ -0,0 +1,106 @@
+asyncapi: 2.0.0
```
fmvilas, line 1:
```suggestion
asyncapi: 2.3.0
```

fmvilas (changes requested):
Hey @sekharbans-ebay! Left some suggestions. I think this PR is in pretty good shape. Would love to push it for 2.3.0 release :)

smoya:
@sekharbans-ebay I created the following 2 Issues so you can use them for tracking what is next in terms of work to do:

- https://github.com/asyncapi/parser-js/issues/434
- https://github.com/asyncapi/spec-json-schemas/issues/147

As @derberg, I'm also happy to help you move this forward so it can be included in `v2.3.0` 🎉

sekharbans-ebay:
> Hey @sekharbans-ebay! Left some suggestions. I think this PR is in pretty good shape. Would love to push it for 2.3.0 release :)

Thank you! I incorporated all suggested changes. Hope i did not miss any!

sekharbans-ebay:
> @sekharbans-ebay I created the following 2 Issues so you can use them for tracking what is next in terms of work to do:
> 
> * [Support `security` field at Operation level parser-js#434](https://github.com/asyncapi/parser-js/issues/434)
> * [Support `security` field at Operation level spec-json-schemas#147](https://github.com/asyncapi/spec-json-schemas/issues/147)
> 
> As @derberg, I'm also happy to help you move this forward so it can be included in `v2.3.0` 🎉
> 
> Be mindful this PR should target `2022-01-release` branch then.

I will need assistance with these two requirements.

smoya:
> Agree. That has been removed. The specification should not dictate a relation between these two - that should be left to the implementation. My original thoughts were based on the broker being a protected entity (i.e. accessible only thru the operations - in this scenario only the operations would need to know the security protocols for the broker which could be independently defined).
> 
> > > >

I do think we should specify that priority is on Operation security. It is not related with implementation since that priority is crucial whenever you design your AsyncAPI document. IMHO tooling should not make assumptions related with the spec.

smoya:
> I will need assistance with these two requirements.

I dropped a line on each issue description with what I think the change should be. They both are small changes. 
After looking to those lines, do you think you need help with the implementation? Whatever you think I can do for helping, feel free to say it :)

smoya:
I would like to suggest we also change a bit the `security` field description at Server level, mentioning users can specify those at operation level as well (for increasing granularity) and include a link into the Security at Operation level.

WDYT?

fmvilas:
I think it's a good idea, @smoya.

sekharbans-ebay:
> I would like to suggest we also change a bit the `security` field description on the [Server Object](https://github.com/asyncapi/spec/pull/584/files#diff-34bca0c91901dffce093114c7d4cac16477410fbca7cabf3a1b888d7dfc3eb4dL359), mentioning users can specify those at operation level as well (for increasing granularity) and include a link into the Security at Operation level.
> 
> WDYT?

Agree. Will make that change

sekharbans-ebay:
> > Agree. That has been removed. The specification should not dictate a relation between these two - that should be left to the implementation. My original thoughts were based on the broker being a protected entity (i.e. accessible only thru the operations - in this scenario only the operations would need to know the security protocols for the broker which could be independently defined).
> > > > >
> 
> I do think we should specify that priority is on Operation security. It is not related with implementation since that priority is crucial whenever you design your AsyncAPI document. IMHO tooling should not make assumptions related with the spec.

Can we have a quick discussion (perhaps zoom?) It would be great if @derberg can join as well. I had it and then removed it on second thoughts - will explain a few scenarios behind the hesitation.

fmvilas:
@sekharbans-ebay we can have this discussion at the next community meeting, what do you think? https://github.com/asyncapi/community/issues/207

sekharbans-ebay:
> @sekharbans-ebay we can have this discussion at the next community meeting, what do you think?
> 
> Feel free to add it to the agenda by leaving a comment here: [asyncapi/community#207](https://github.com/asyncapi/community/issues/207).

Thanks for adding will join.

sekharbans-ebay:
> > I will need assistance with these two requirements.
> 
> I dropped a line on each issue description with what I think the change should be. They both are small changes. After looking to those lines, do you think you need help with the implementation? Whatever you think I can do for helping, feel free to say it :)

Thanks! I ll try get these in by Friday.

jonaslagoni:
Based on the SIG meeting, I would suggest to try applying the changes currently suggestions, to find out where the limitation is and to which protocols. I.e. apply it to the [social media example](https://github.com/asyncapi/spec/tree/master/examples/social-media), where you try to apply different security mechanisms to the servers and operations. That way we get a sense of the constraints and where it is "needed". 

Just to mention it again, if it is only related to a specific protocol (or a few) it might be better to let it [recede in the bindings](https://github.com/asyncapi/bindings/), just keep it in mind 🙂

smoya:
> * limit to only `oauth2` scheme

Wouldn't this also apply to others like `http`, `httpApiKey`, etc? Like a path that requires a Bearer token (i.e. JWT)? This could be different depending on the operation and not only based on the server.

> specify something like Only one of the security requirement objects from MUST be satisfied to authorize an operation, including security requirements from Server Object

I'm thinking that a channel may have several layers of security and might use different security schemes. For example, `asymmetricEncryption` for connecting to the server, but in particular for the operation, another one?
Maybe I'm confused by how this all works under the hood so please bear with me!

derberg:
> If user wants different, they can easily create another server, with proper description, without security and in their operation, they can point `servers: ['theonewithoutsec']. So this approach lets you do both, make operation to also apply server, and you can always ignore server by providing one without security.

> There is no need implied in this PR to restrict this to only oauth2

See more sense now to have it as it is, for all schemes, especially in regards to the part I described above. You can have `server` with no security and then specify that one channel is secured with `oauth2` and the other with `userPassword `. Especially with websocket where you get a lot of flexibility, and every path/channel may need a separate connection and can be secured separately anyway you want.

I think it should be doable to apply it here https://github.com/asyncapi/spec/tree/master/examples/social-media as pointed out by @jonaslagoni. There is no security there so it is always easier to add something that try to modify something existing. @sekharbans-ebay is it something you think you could try?

sekharbans-ebay:
> 



> > If user wants different, they can easily create another server, with proper description, without security and in their operation, they can point `servers: ['theonewithoutsec']. So this approach lets you do both, make operation to also apply server, and you can always ignore server by providing one without security.
> 
> > There is no need implied in this PR to restrict this to only oauth2
> 
> See more sense now to have it as it is, for all schemes, especially in regards to the part I described above. You can have `server` with no security and then specify that one channel is secured with `oauth2` and the other with `userPassword `. Especially with websocket where you get a lot of flexibility, and every path/channel may need a separate connection and can be secured separately anyway you want.

Yes - that was the intent - why i also think this is an important feature that the specification is currently lacking.  As i was mentioning above - . a contract between the operations and the broker could be (and if i may state - _would_ be ) different than that between a channel operation and an integrator(for example a consumer application to a subscribe op).
> 
> I think it should be doable to apply it here https://github.com/asyncapi/spec/tree/master/examples/social-media as pointed out by @jonaslagoni. There is no security there so it is always easier to add something that try to modify something existing. @sekharbans-ebay is it something you think you could try?

Sure i can. I ll PR a modified version of this one with clear example that demonstrates that the concerns on sever/broker level security could be different that one used to expose operations to integrators. 

I assume since this is targeted now for April release - i can take on this a bit later? I am swamped a bit this month :)

derberg:
> I assume since this is targeted now for April release - i can take on this a bit later? I am swamped a bit this month :(

from my point of view it doesn't have to be moved to April, depending on our speed. But it is always recommended to do things slowly, not under time pressure 😉 so if it is not that bad for you to wait few months more that I definitely recommend we skip January-window

sekharbans-ebay:
> > I assume since this is targeted now for April release - i can take on this a bit later? I am swamped a bit this month :(
> 
> from my point of view it doesn't have to be moved to April, depending on our speed. But it is always recommended to do things slowly, not under time pressure 😉 so if it is not that bad for you to wait few months more that I definitely recommend we skip January-window

So we dont need to wait for broader community feedback?

smoya:
After reading https://github.com/asyncapi/parser-js/pull/445#issuecomment-1012870837 from @derberg:

> So far your PR follows approach when on operation level, only operation security applies, but we are getting conclusions there that server security might apply too, so server function should then return not only operation sec requirements but also the ones provided in the associated server.

Is it what we all agreed on here? I think this is still confusing the user, since if the security requirements from the server are returned at operation level as well, the user won't know which one applies **only** to the operation.
In fact, if we follow the statement found in the docs: 

> Only one of the security requirement objects need to be satisfied to authorize a connection or operation.

If this is also applied to security at Operation level, that means you won't ever need to know the requirements from server as you only can pick one of it. It does not make sense to return server requirements IMHO.

derberg:
> Is it what we all agreed on here? I think this is still confusing the user, since if the security requirements from the server are returned at operation level as well, the user won't know which one applies only to the operation.
In fact, if we follow the statement found in the docs:

this is what proper spec documentation is for to describe it. So basically one of the security requirements, from operation or server applies. 

> If this is also applied to security at Operation level, that means you won't ever need to know the requirements from server as you only can pick one of it. It does not make sense to return server requirements IMHO.

It is up to the user, to decide. Spec at current shape, as I describe in previous comments, allows you to do a mix. This is the point here. If you want only one of security req from operation, then do not provide them in server.

smoya:
> It is up to the user, to decide. Spec at current shape, as I describe in previous comments, allows you to do a mix. This is the point here. If you want only one of security req from operation, then do not provide them in server.

But what's the point of adding a security requirement on a channel if there are ones already declared at Server level? Why won't you add it directly into the Server? 

I think we should treat the `security` field at Operation level as a completely independent set of security requirements and not complementary to the ones declared at Server level.
I also think this is what @sekharbans-ebay suggested since the very beginning if I'm right.

derberg:
> But with that statement in mind, what's the point of adding a security requirement on a channel if there are ones already declared at Server level? Why won't you add it directly into the Server?
I think we should treat the security field at Operation level as a completely independent set of security requirements and not complementary to the ones declared at Server level. This also reduces friction and complexity.
I also think this is what @sekharbans-ebay suggested since the very beginning if I'm right.

well this is exactly what @sekharbans-ebay is about to show in this PR after updating description and the examples (social media) where we will be able to see use case on oauth2 scopes. At least this is my understanding 😄

sekharbans-ebay:
> > But with that statement in mind, what's the point of adding a security requirement on a channel if there are ones already declared at Server level? Why won't you add it directly into the Server?
> > I think we should treat the security field at Operation level as a completely independent set of security requirements and not complementary to the ones declared at Server level. This also reduces friction and complexity.
> > I also think this is what @sekharbans-ebay suggested since the very beginning if I'm right.
> 
> well this is exactly what @sekharbans-ebay is about to show in this PR after updating description and the examples (social media) where we will be able to see use case on oauth2 scopes. At least this is my understanding 😄

Is this rephrase acceptable?:
 
"A declaration of which security mechanisms are associated with this operation. Only one of the security requirement objects MUST be satisfied to authorize an operation. Security mechanisms specified at the operation level is completely independent, specific to the operation and not complementary to the ones declared at server level. "

derberg:
I have to admit I'm confused 😆 might be my language skills, so let me ask:

`Security mechanisms specified at the operation level is completely independent, specific to the operation and not complementary to the ones declared at server level.` basically means operation security overwrites server security? so we are back to the initial idea?

smoya:
> I have to admit I'm confused 😆 might be my language skills, so let me ask:
> 
> `Security mechanisms specified at the operation level is completely independent, specific to the operation and not complementary to the ones declared at server level.` basically means operation security overwrites server security? so we are back to the initial idea?

I think so. I'm ok with it. At the very begining I thought it would be better to be complementary but I totally changed my mind once I went deeply on the issue.

So yeah, I'm dropping the following questions to you @derberg:

- Do you see any red flag on this approach anyway?
- Do you believe it should be complementary instead?

If you don't see a red flag, I'm +1 to move forward with https://github.com/asyncapi/spec/pull/584#issuecomment-1014395390

derberg:
no red flags from my side, I was all the time with approach operation overrides server, but thought there is a use case to have it complimentary, with oauth2 and scopes. 

we still need @fmvilas voice, we both are codeowners here so far

fmvilas:
I'm still not quite sure I get why it should be completely separate from servers. Trying to sleep over it, that's what I didn't say anything but I'm reading all your comments. Are we finally going to provide examples as we spoke during the last meeting?

sekharbans-ebay:
> I'm still not quite sure I get why it should be completely separate from servers. Trying to sleep over it, that's what I didn't say anything but I'm reading all your comments. Are we finally going to provide examples as we spoke during the last meeting?

Yes i will add simple examples to the PR. 

But to help the decision documenting a few example scenarios that i can think of. Hopefully this will help make the right call on interdependencies or lack thereof (folks can add more scenarios and that might help us make the right call):
I will withhold making any recommendations - i have myself done a 180 on the interdependency question so i ll let the scenarios and consensus decide :) 

1.  There is a subscribe operation accessible thru a 'read' auth scope and a 'produce' operation accessible thru a write scope. The broker supports both 'read' and 'write' scopes
2. The operations are based on a different security mechanism altogether (say user/password) while the server supporting the use case uses oauth. Operations use oauth in this scenario to interact with the broker. External integrators only interact with channels and operations and do not need to know about the contract with the underlying message broker. 
3. A work in progress example: Developers are in the process for introducing new event streams with their own security protocols and are interested in the underlying server or message broker and the security protocols that apply. A contract on such a work in progress system topology may not have details of all operations and operation level security requirements but just the underlying broker.
4. The inheritance case(Server level security assumed for operations): This to me is more due to the inability to specify security at the operation level. When we came up with contracts for our use case - we had debated on using this ourselves but dropped this idea since repurposing server/broker level security for operation level seemed incorrect and would not scale across use cases

dalelane:
Sorry, I haven't had a chance to follow this as closely as I'd have liked. Are we coming to a consensus here for [including this in 2.3](https://github.com/asyncapi/spec/issues/675)? My take from a quick skim through to catch up is that we're not and that it would be better to defer this to the next release, but it's a long thread so I might be wrong!

fmvilas:
> One way to look at this is that the security on an operation level is an independent entity and of interest to consumers (or producers) reading or writing to a channel (i.e. a contract between the channel and its users)- where as the brokers themselves may have an independent security requirement (i.e a contract between the components that represent these operations and the underlying server).
>
> Request your thoughts please. Should we have another discussion around the example scenarios (i am sure there could be more) to decide?

Alright, let me frame my thoughts:

If these security mechanisms are a completely different set from the server one, does it mean I'd not need to use the ones defined at the server level at all? For instance, to publish a message to a broker I need to first connect and then send the message. Those are two different things. I'm assuming this PR is defining the second (publish operation). However, it would seem incorrect to me to say that I only need to meet the security mechanisms defined in the publish operation because, to be able to publish, I first have to connect, right? Therefore, it's not totally a different set of security mechanisms. They're somehow complementary. Unless we're having some protocols in mind like HTTP, where connecting and publishing are in practice the same thing. In such a case, maybe this should go on the HTTP binding instead?

Also, isn't it already possible to define two servers, define different security mechanisms for both, and assign the appropriate server at the operation level? By doing this, you don't need the `security` keyword at the operation level. It's arguable though that the DX might not be great. I know it's been mentioned already but that's still unclear to me.

However, if we make it complementary to the servers' security mechanisms, now it kinda resonates for me. I can see the benefit in this case, where you define your _default_ security mechanisms at the server level and the operation-specific ones at the operation level. This would mean you have to meet the security mechanisms at the server **and** the operation levels.

In general, what I'm lacking is a clear set of example use cases with their corresponding AsyncAPI files. @sekharbans-ebay I think the ones you outlined are great but I still fail to see how you'd apply them using this new keyword at the operation level.

> Are we coming to a consensus here for including this in 2.3?

Great point, @dalelane. IMHO, we shouldn't include this in the current release as we have less than 6 days and it might put some pressure on people to vote in a rush. Also, April is around the corner and would give us much more time to deeply think about it.

That said, I want to apologize to @sekharbans-ebay for not reviewing this PR before. I know it's been open for a while and it's definitely not your fault if it doesn't make it for this release. You've done more than we asked for 🙏

sekharbans-ebay:
> > One way to look at this is that the security on an operation level is an independent entity and of interest to consumers (or producers) reading or writing to a channel (i.e. a contract between the channel and its users)- where as the brokers themselves may have an independent security requirement (i.e a contract between the components that represent these operations and the underlying server).
> > Request your thoughts please. Should we have another discussion around the example scenarios (i am sure there could be more) to decide?
> 
> Alright, let me frame my thoughts:
> 
> If these security mechanisms are a completely different set from the server one, does it mean I'd not need to use the ones defined at the server level at all? For instance, to publish a message to a broker I need to first connect and then send the message. Those are two different things. I'm assuming this PR is defining the second (publish operation). However, it would seem incorrect to me to say that I only need to meet the security mechanisms defined in the publish operation because, to be able to publish, I first have to connect, right? Therefore, it's not totally a different set of security mechanisms. They're somehow complementary. Unless we're having some protocols in mind like HTTP, where connecting and publishing are in practice the same thing. In such a case, maybe this should go on the HTTP binding instead?

Yes its not in that sense - I guess I was thinking more in terms of components that represent these operations. But realize these are purely stream operations. 

> Also, isn't it already possible to define two servers, define different security mechanisms for both, and assign the appropriate server at the operation level? By doing this, you don't need the `security` keyword at the operation level. It's arguable though that the DX might not be great. I know it's been mentioned already but that's still unclear to me.

I have attached an example: https://github.com/sekharbans-ebay/spec/blob/master/examples/streetlights-operation-security.yml
which is a slight extension of the streetlights example with multiple security protocols that apply at different ports. The operations in this example need specific authorizations to be called out. 

> However, if we make it complementary to the servers' security mechanisms, now it kinda resonates for me. I can see the benefit in this case, where you define your _default_ security mechanisms at the server level and the operation-specific ones at the operation level. This would mean you have to meet the security mechanisms at the server **and** the operation levels.

Yes i agree. 

> In general, what I'm lacking is a clear set of example use cases with their corresponding AsyncAPI files. @sekharbans-ebay I think the ones you outlined are great but I still fail to see how you'd apply them using this new keyword at the operation level.

I have added an example of a kafka broker that supports multiple security mechanisms and operations specifically mention the auth(s) that apply. Hope this is a reasonable example: https://github.com/sekharbans-ebay/spec/blob/master/examples/streetlights-operation-security.yml

> > Are we coming to a consensus here for including this in 2.3?
> 
> Great point, @dalelane. IMHO, we shouldn't include this in the current release as we have less than 6 days and it might put some pressure on people to vote in a rush. Also, April is around the corner and would give us much more time to deeply think about it.

> That said, I want to apologize to @sekharbans-ebay for not reviewing this PR before. I know it's been open for a while and it's definitely not your fault if it doesn't make it for this release. You've done more than we asked for 🙏

Realize this requires some more thought - and i have been often delayed in my response! Next release sounds perfect. 
Grateful for the all the feedback on this PR and the support from this awesome community on this topic.

derberg:
I'm dropping link to https://github.com/asyncapi/spec/issues/418. Unfortunately folks do not respond there anymore but it is at least a good input that more people need it, and so far it is always just about the `oauth` and the scopes. So maybe as I mentioned in previous comments, we should only focus on a specific scheme first, and enable others if requirements show up

smoya:
Hi @sekharbans-ebay! I wonder if you will have some bandwidth to work on this for the next release, which is happening next month (April 2022). How can we help move it forward?

sekharbans-ebay:
> Hi @sekharbans-ebay! I wonder if you will have some bandwidth to work on this for the next release, which is happening next month (April 2022). How can we help move it forward?

Yes. I need consensus on the description on the operation level security support. Based on feedback from @fmvilas - here is a suggested text. If we all agree will move fwd with the PR asap:

_"A declaration of which security mechanisms are associated with this operation. Only one of the security requirement objects need to be satisfied to authorize an operation. In cases where Server Security also applies, it is expected the security mechanisms specified at the operation level would be complementary to those specified at the server."_

fmvilas:
> would be complementary to those specified at the server.

@sekharbans-ebay by "complementary" you mean both security mechanisms need to be satisfied?

Also, feel free to do these changes in the PR upfront and we can discuss them there. I think it will be easier for everyone.

sekharbans-ebay:
> > would be complementary to those specified at the server.
> 
> @sekharbans-ebay by "complementary" you mean both security mechanisms need to be satisfied?

Yes - I was kind of undecided but agree to your earlier comment. Will update the PR.

sekharbans-ebay:
> Left some suggestions to reword it. Please let me know if that's good enough. I feel we're close this time 🤞 😄

Accepted recommendations. **LokeshRishi** helped with the following associated PRs: 
https://github.com/asyncapi/spec-json-schemas/pull/189
https://github.com/asyncapi/parser-js/pull/505

Hope we make it this time :)

examples/operation-security.yml
```diff
@@ -0,0 +1,106 @@
...
+          method: POST
+          headers:
+            type: object
+            properties:
+              Content-Type:
+                type: string
+                enum:
+                  - application/json
+      security:
+        petstore_auth:
```
fmvilas, line 25: In this example, `petstore_auth` is a key in the `security` object but the spec (and the examples on the spec) say it should be an array like:

```yaml
security:
  - petstore_auth:
    - subscribe:auth_revocations
```

Honestly, I think this example is correct and we should change the spec description because [`security` is also a key in the Server Object](https://www.asyncapi.com/docs/specifications/v2.3.0#securityRequirementObject). We should be consistent with the Server Object IMHO.

sekharbans-ebay: Hi @fmvilas  - I see your first point - yes the example has it wrong. But to me, the proposed spec description is correct  and does have security: [[Security Requirement Object](https://github.com/asyncapi/spec/blob/3460840ee5a68b0b9f6d881da8f6262cc8602ef2/spec/asyncapi.md#securityRequirementObject)]

which is the same as on the server level. I ll fix the example.  Apologize if i misunderstood your comment

fmvilas: You're right, I got confused by the Security Requirement Object examples 🤦 It should be an array of objects 👍

examples/streetlights-operation-security.yml
```diff
@@ -0,0 +1,194 @@
...
+    publish:
+      summary: Inform about environmental lighting conditions of a particular streetlight.
+      operationId: receiveLightMeasurement
+      traits:
+        - $ref: '#/components/operationTraits/kafka'
+      message:
+        $ref: '#/components/messages/lightMeasured'
+      security:
+       streetlights_auth:
+       - streetlights:write
```
fmvilas, line 51:
```suggestion
         - streetlights:write  
```

examples/streetlights-operation-security.yml
```diff
@@ -0,0 +1,194 @@
...
+        $ref: '#/components/parameters/streetlightId'
+    subscribe:
+      operationId: turnOn
+      traits:
+        - $ref: '#/components/operationTraits/kafka'
+      message:
+        $ref: '#/components/messages/turnOnOff'
+      security:
+       streetlights_auth:
+       - streetlights:read
```
fmvilas, line 64:
```suggestion
         - streetlights:read  
```

examples/streetlights-operation-security.yml
```diff
@@ -0,0 +1,194 @@
...
+        $ref: '#/components/parameters/streetlightId'
+    subscribe:
+      operationId: turnOff
+      traits:
+        - $ref: '#/components/operationTraits/kafka'
+      message:
+        $ref: '#/components/messages/turnOnOff'
+      security:
+       streetlights_auth:
+       - streetlights:read
```
fmvilas, line 78:
```suggestion
         - streetlights:read    
```

examples/streetlights-operation-security.yml
```diff
@@ -0,0 +1,194 @@
...
+        $ref: '#/components/parameters/streetlightId'
+    subscribe:
+      operationId: dimLight
+      traits:
+        - $ref: '#/components/operationTraits/kafka'
+      message:
+        $ref: '#/components/messages/dimLight'
+      security:
+       streetlights_auth:
+       - streetlights:read
```
fmvilas, line 91:
```suggestion
         - streetlights:read  
```

fmvilas (changes requested):
Minor format suggestions this time. I know the current indentation is valid but would love to change them for the sake of consistency. Other than that, I'm happy with this PR 🚀

sekharbans-ebay:
> Minor format suggestions this time. I know the current indentation is valid but would love to change them for the sake of consistency. Other than that, I'm happy with this PR 🚀

Thanks so much! Committed your suggestions. Hope we are green for April release ?

fmvilas (approved):
Haven't checked the other PRs for parser and spec-json-schemas but for me this one is ready. Let's wait for others to review too. Great stuff @sekharbans-ebay and thanks a lot for your patience 🙏

fmvilas:
@sekharbans-ebay I just changed the base branch from `master` to `2022-04-release` and there are some conflicts you may want to have a look at.

sekharbans-ebay:
> @sekharbans-ebay I just changed the base branch from `master` to `2022-04-release` and there are some conflicts you may want to have a look at.

Done thanks.

derberg (commented):
Looks like we are super close to release this one 🚀 

I only have 2 questions:
- what new `examples/multiple-server-security.yml` is adding that `/examples/streetlights-mqtt.yml` doesn't have already?
- To approve I need to be 100% sure I get what is the current for `security` property. Now it sounds to me, that in nodejs-template I can generate code that (taking into account [one of examples](https://github.com/asyncapi/spec/pull/584/files?diff=unified&w=0#diff-ba32d827ae52b1b1dd9cc1bb8ad0426c78ea7251bcb0430b7bdf3f6677fa1f01)) sends a message to `smartylighting.streetlights.1.0.action.{streetlightId}.turn.on` with headers including token with `streetlights:read` scope. I can also, for the same channel, but when using `test_auth` server, totally ignore oauth and instead of passing token I can just use username/password for scramSha256?

If my ☝🏼 understanding is correct, could you please add proper inline comments to the [examples/streetlights-operation-security.yml](https://github.com/asyncapi/spec/pull/584/files?diff=unified&w=0#diff-ba32d827ae52b1b1dd9cc1bb8ad0426c78ea7251bcb0430b7bdf3f6677fa1f01) example? Maybe we could connect one of the channels using `server` property with only one server, so it would be super easy to explain in the inline comment?

README.md
```diff
@@ -41,7 +41,6 @@
 The latest draft specification can be found at [spec/asyncapi.md](./spec/asyncapi.md) which tracks the latest commit to the master branch in this repository.
 
 - [Version 2.4.0](https://github.com/asyncapi/spec/blob/v2.4.0/spec/asyncapi.md) (latest)
-- [Version 2.3.0](https://github.com/asyncapi/spec/blob/v2.3.0/spec/asyncapi.md)
```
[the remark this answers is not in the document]
sekharbans-ebay: > why was it removed?

Inadvertent. Will restore.

sekharbans-ebay:
> Looks like we are super close to release this one 🚀
> 
> I only have 2 questions:
> 
> * what new `examples/multiple-server-security.yml` is adding that `/examples/streetlights-mqtt.yml` doesn't have already?

Agreed. We don't need this. I was making a point on multiple security to @fmvilas  and had this as an example. If you wish i will delete this example. 

> * To approve I need to be 100% sure I get what is the current for `security` property. Now it sounds to me, that in nodejs-template I can generate code that (taking into account [one of examples](https://github.com/asyncapi/spec/pull/584/files?diff=unified&w=0#diff-ba32d827ae52b1b1dd9cc1bb8ad0426c78ea7251bcb0430b7bdf3f6677fa1f01)) sends a message to `smartylighting.streetlights.1.0.action.{streetlightId}.turn.on` with headers including token with `streetlights:read` scope. I can also, for the same channel, but when using `test_auth` server, totally ignore oauth and instead of passing token I can just use username/password for scramSha256?

Yes thats correct. 

> If my ☝🏼 understanding is correct, could you please add proper inline comments to the [examples/streetlights-operation-security.yml](https://github.com/asyncapi/spec/pull/584/files?diff=unified&w=0#diff-ba32d827ae52b1b1dd9cc1bb8ad0426c78ea7251bcb0430b7bdf3f6677fa1f01) example? Maybe we could connect one of the channels using `server` property with only one server, so it would be super easy to explain in the inline comment?

I added inline comments in the operation level security in the example. Hope this is ok.

derberg (approved):
LGTM

imho `examples/multiple-server-security.yml` can be removed because as I wrote, we have such example already in another file. Nevertheless not a strong opinion. 

@fmvilas @dalelane if you think it should stay, I'm also fine

anyway, IMHO proposal should reach `Stage2: Draft` as reference PRs are there:
- https://github.com/asyncapi/parser-js/pull/505
- https://github.com/asyncapi/spec-json-schemas/pull/189

examples/streetlights-operation-security.yml
```diff
@@ -0,0 +1,202 @@
...
+      streetlightId:
+        $ref: '#/components/parameters/streetlightId'
+    subscribe:
+      operationId: turnOn
+      traits:
+        - $ref: '#/components/operationTraits/kafka'
+      message:
+        $ref: '#/components/messages/turnOnOff'
+      security:
+      # This operation level security implies the ability to sends a message to
```
smoya, line 63:
```suggestion
      # This operation level security implies the ability to send a message to 
```

cc @alequetzalli in case you want to review this comment (it is a public example).

examples/streetlights-operation-security.yml
```diff
@@ -0,0 +1,202 @@
...
+      streetlightId:
+        $ref: '#/components/parameters/streetlightId'
+    subscribe:
+      operationId: dimLight
+      traits:
+        - $ref: '#/components/operationTraits/kafka'
+      message:
+        $ref: '#/components/messages/dimLight'
+      security:
+      # This operation level security implies the ability to sends a message to
```
smoya, line 94:
```suggestion
      # This operation level security implies the ability to send a message to 
```

cc @alequetzalli in case you want to review this comment (it is a public example).

dalelane:
> imho `examples/multiple-server-security.yml` can be removed because as I wrote, we have such example already in another file. Nevertheless not a strong opinion.
> 
> @fmvilas @dalelane if you think it should stay, I'm also fine

If it's not adding anything new, my vote would be to remove it, because otherwise it's another example we have to maintain

examples/operation-security.yml
```diff
@@ -0,0 +1,106 @@
+asyncapi: 2.3.0
```
dalelane, line 1:
```suggestion
asyncapi: 2.4.0
```

dalelane (approved):

sekharbans-ebay:
> > imho `examples/multiple-server-security.yml` can be removed because as I wrote, we have such example already in another file. Nevertheless not a strong opinion.
> > @fmvilas @dalelane if you think it should stay, I'm also fine
> 
> If it's not adding anything new, my vote would be to remove it, because otherwise it's another example we have to maintain

Deleted

alequetzalli (approved):
LGTM 🚀

smoya:
The rest of the PR's have been merged now:
- https://github.com/asyncapi/spec-json-schemas/pull/189
- https://github.com/asyncapi/parser-js/pull/505

Moving with the merge of this one as approvals from all Code Owners are present.

smoya:
I don't really understand the flow of the bots :(

derberg:
the bot that talks to you now is an old bug in a workflow, that was fixed on `master` only. @KhudaDad414 already fixed it. So do not worry about it. We just need to make sure to update release branch with latest master

`do-not-merge` label can be taken off only by maintainers, so there is no way that someone else merge release PRs, as maintainers need to make sure related PRs in other repos were merged properly.

I see you merged both related PRs but problem is that the parser PR do not use `@asyncapi/specs` release candidate `2.14.0-2022-04-release.1`. So it basically does not have the latest schema. In other words, someone using parser-js release candidate will not have validation included

smoya:
> the bot that talks to you now is an old bug in a workflow, that was fixed on `master` only. @KhudaDad414 already fixed it. So do not worry about it. We just need to make sure to update release branch with latest master
> 
> `do-not-merge` label can be taken off only by maintainers, so there is no way that someone else merge release PRs, as maintainers need to make sure related PRs in other repos were merged properly.
> 
> I see you merged both related PRs but problem is that the parser PR do not use `@asyncapi/specs` release candidate `2.14.0-2022-04-release.1`. So it basically does not have the latest schema. In other words, someone using parser-js release candidate will not have validation included

Thanks for the answer! 
I created a PR in the parser for pointing specs package to the latest release candidate. However, wondering if I could use some wildcard to not having to do this change every time we wanna merge a change.

https://github.com/asyncapi/parser-js/pull/521

smoya:
@derberg @fmvilas @dalelane if you agree, please remove the `do-not-merge` label and proceed merging this PR.

examples/streetlights-operation-security.yml
```diff
@@ -0,0 +1,202 @@
...
+    subscribe:
+      operationId: turnOn
+      traits:
+        - $ref: '#/components/operationTraits/kafka'
+      message:
+        $ref: '#/components/messages/turnOnOff'
+      security:
+      # This operation level security implies the ability to send a message to 
+      # `smartylighting.streetlights.1.0.action.{streetlightId}.turn.on` with Authorization headers 
+      # with `streetlights:read` scope. It is also possible for the same channel when using `test_auth`
```
smoya, line 65: I just noticed there is a typo here. It should be: `test_oauth`

examples/streetlights-operation-security.yml
```diff
@@ -0,0 +1,202 @@
...
+      operationId: turnOn
+      traits:
+        - $ref: '#/components/operationTraits/kafka'
+      message:
+        $ref: '#/components/messages/turnOnOff'
+      security:
+      # This operation level security implies the ability to send a message to 
+      # `smartylighting.streetlights.1.0.action.{streetlightId}.turn.on` with Authorization headers 
+      # with `streetlights:read` scope. It is also possible for the same channel when using `test_auth` 
+      # server to instead use credentials for security (scramSha256 in this example) specified on the server level
```
smoya, line 66: Now that I'm writing the release notes, this comment seems confusing to me.

The spec says: 

> In cases where Server Security also applies, it MUST also be satisfied.

However, by reading this comment, I understand that only one (the server or rather the operation) security requirement should be satisfied.

Would you mind clarifying this @sekharbans-ebay? Also pinging Code Owners @derberg @fmvilas @dalelane

sekharbans-ebay: > Now that I'm writing the release notes, this comment seems confusing to me.
> 
> The spec says:
> 
> > In cases where Server Security also applies, it MUST also be satisfied.
> 
> However, by reading this comment, I understand that only one (the server or rather the operation) security requirement should be satisfied.
> 
> Would you mind clarifying this @sekharbans-ebay? Also pinging Code Owners @derberg @fmvilas @dalelane

The example text is poorly worded - i ll reword this. My intention was to state that you could still  operate on the channel using security credentials supported at the server level perhaps thru a different operation.

smoya: Would you mind creating a PR against `2022-04-release` branch with such a fix @sekharbans-ebay ?

examples/streetlights-operation-security.yml
```diff
@@ -0,0 +1,202 @@
...
+      summary: Inform about environmental lighting conditions of a particular streetlight.
+      operationId: receiveLightMeasurement
+      traits:
+        - $ref: '#/components/operationTraits/kafka'
+      message:
+        $ref: '#/components/messages/lightMeasured'
+      security:
+       streetlights_auth:
+         - streetlights:write  
+  smartylighting.streetlights.1.0.action.{streetlightId}.turn.on:
```
smoya, line 52: Would it make sense to link the channel to a server(s)?

You can do this by adding `servers` field. For example:

```yaml
servers:
  - test_oauth
```

examples/streetlights-operation-security.yml
sekharbans-ebay: > Would you mind creating a PR against `2022-04-release` branch with such a fix (and the others I mentioned in other comments) @sekharbans-ebay ?

Done. Please review:
https://github.com/asyncapi/spec/pull/768

Issue #371, opened 2020-04-06
[BUG] Old server variable left out in example


fmvilas:
https://github.com/asyncapi/asyncapi/commit/471a0666c92cb416fd19118b7162b81b47b7eb9e#commitcomment-38317219

Thanks, @nulltoken 🙌

PR #589, opened 2021-07-27
docs: document missed server variable in the example

magicmatatjahu:
Reading the old issues, I checked that https://github.com/asyncapi/spec/issues/371 issue is still unresolved, so this PR fixes it.

Changes:
- document missed server variable (`{basePath}`) in the example.

Related issue(s)
Resolves https://github.com/asyncapi/spec/issues/371

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -446,6 +446,9 @@ The following shows how variables can be used for a server configuration:
             "8884"
           ],
           "default": "8883"
+        },
+        "basePath": {
+          "default": "v2"
         }
       }
     }
@@ -469,6 +472,9 @@ servers:
           - '8883'
           - '8884'
         default: '8883'
+      basePath:
+        # open meaning there is the opportunity to use special base paths as assigned by the provider, default is `v2`
+        default: v2
 ```
 
 


fmvilas (approved):

magicmatatjahu:
@derberg As I know you are preparing the release process, so probably my case is important for you. I only updated the example in the spec, not any description/sentences in spec, but only single example, so I think that it should be merged without release of the `2.1.1` spec.

derberg:
@magicmatatjahu My statement would be that examples are not part of the spec, so even though they are in `asyncapi.md` file, we should not release `2.1.1` because of example fix. We should only do patch releases when there are changes to the main content, the specification itself, the text that describes it.

The only issue with this approach is that this means we will merge this PR without release and it will not be reflected on the AsyncAPI Website, so we need to enhance a release pipeline here to cover such cases.

I wonder what others think

magicmatatjahu:
@derberg I think that we can use the `on.push.paths` https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths to run https://github.com/asyncapi/spec/blob/master/.github/workflows/new-spec-release.yml on every release and push

magicmatatjahu:
I added PR for updating spec on website - https://github.com/asyncapi/spec/pull/591

PR #592, opened 2021-07-27
docs: remove trailing commas from examples

smarek:
Per documentation https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas

JSON specification does not allow trailing commas

On related note the [Components Object Example (JSON)](https://github.com/asyncapi/spec/blob/v2.1.0/spec/asyncapi.md#components-object-example) is ambiguous, because as it currently is, it is not valid JSON object but only part of the object.
Where we have 2 possible resolutions
  1. Remove the `"components": ` part to show only the object containing the "Components Object"
  2. Wrap the example in another object, possibly just add the opening and closing curly brackets.
  
All other JSON objects currently in spec are valid afaict

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -1358,90 +1358,92 @@ my.org.User
 ##### Components Object Example
 
 ```json
-"components": {
-  "schemas": {
-    "Category": {
-      "type": "object",
-      "properties": {
-        "id": {
-          "type": "integer",
-          "format": "int64"
-        },
-        "name": {
-          "type": "string"
-        }
-      }
-    },
-    "Tag": {
-      "type": "object",
-      "properties": {
-        "id": {
-          "type": "integer",
-          "format": "int64"
-        },
-        "name": {
-          "type": "string"
-        }
-      }
-    }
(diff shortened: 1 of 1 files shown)


fmvilas (approved):
Good catch 🦅 👁️

smarek:
I've just also provided in separate commit fix for JSON in Components Object Example, if you'd prefer the other proposed solution, i can rewrite it, just say

derberg:
@smarek hard to figure out what you changed in the 2nd commit, am I right it is formatting and wrapping example in `{}`?

magicmatatjahu:
@derberg Yeap, wrapping in the `{}` (:

smarek:
@derberg yup, the git diff looks strange, but all i did was 2-spaces indent of the original code block and add the start/end brackets

magicmatatjahu (approved):
LGTM!

derberg:
let's wait with merge until @magicmatatjahu merges improvements to pipeline that will automatically pick up this merge and publish updated spec document on the website

PR #593, opened 2021-07-27
docs: all links should use secure protocol

smarek:
HTTP protocol is discouraged and all domains currently used in spec allow for secure version

Per my simple check, these are the domains (and their usage count) in current spec, so the manual verification of HTTPS support on all of these is trivial

```
> grep -i http spec/asyncapi.md | sed -rn 's@^.*\[(.*)\]\((.*)/.*$@\2@p' | awk -F'/' '{print $3}' | sort | uniq -c | sort -hr
     62 github.com
     14 tools.ietf.org
     14 spec.commonmark.org
      3 json-schema.org
      2 xml2rfc.ietf.org
      2 www.yaml.org
      1 www.ietf.org
      1 www.ecma-international.org
      1 www.apache.org
      1 avro.apache.org
```

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -6,9 +6,9 @@ Part of this content has been taken from the great work done by the folks at the
 
 #### Version 2.1.0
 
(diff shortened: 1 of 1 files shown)


magicmatatjahu (approved):
LGTM
@derberg Do you want to add something?

derberg:
@magicmatatjahu only prefix of PR title needs to be changes, I wrote in another PR that I think default package doesn't support it as it is Angular community prefix, afaik. So just changing to docs and you can merge

magicmatatjahu:
@smarek Thanks for contribution! :)

Issue #583, opened 2021-07-13
Inconsistencies for schemas in server.variables and channel.parameters


magicmatatjahu:
I'm wondering why schema (doc/constraints) for server variables uses only a subset of JSON Schema (enum, default, examples and description) rather a full set of JSON Schema keywords like `channel.parameters.[*].schema`. I can understand that if you have server's [url/host](https://github.com/asyncapi/spec/issues/547) then you should have in 99% cases "hardcoded" variables for it - variable's type can be only a string and with some set of values (`enum`), but maybe we should allow also more complex schema like `pattern` etc? If channel's parameter has it (which is more generic), so why not server variable? I need some clarification in this place. 

Another question/problem: if server variable can have defined `enum` then for what the `examples` exists? I mean if there is a case, when given variable have a predefined set of values (`enum`) then `examples` === `enum`, so `examples` is unnecessary. Other case, when `examples` field is used to describe possible "other" values for variable, so maybe then we should allow more complex schema to describe the variable (see my first question)?

I hope that you understand my concerns and thanks very much for comments!

fmvilas:
The reason it was defined as a small subset of JSON Schema was to remove complexity in an area that would probably not need much. I agree `pattern` could be added. I'd still not go for a full JSON Schema object because that would imply support for oneOf, anyOf, allOf, not, if, then, else, etc. And to be honest, I haven't seen a use case needing such complex schemas yet.

> if server variable can have defined enum then for what the examples exist?

Because `enum` is optional and your variable may be a free-form string. Therefore, you may need to provide examples.

> If channel's parameter has it (which is more generic), so why not server variable?

Now I'm wondering if we really need a full JSON Schema object on parameters 😄

magicmatatjahu:
> Because enum is optional and your variable may be a free-form string. Therefore, you may need to provide examples.

And for that we should enable the all JSON Schema keywords for string type like `minLength`, `maxLength`, `pattern` etc. and don't base "validation" on examples. Also, If we really go with current solution as subset of JSON Schema, I opt for enable also the `integer` and `number` type to describe much easier the variables which has form of numbers - for that someone can use the `minimum`, `maximum` etc keywords.

> Now I'm wondering if we really need a full JSON Schema object on parameters 😄

[...]

Issue #528, opened 2021-04-13
Avro schemas are not supported in the components/schemas section


MichaelDavisSolace:
### Describe the bug
Asyncapi supports Avro schemas, but only if they are in the message/payload section. It would be nice if we could put Avro schemas in the components/schemas section.

### To Reproduce
1. Create a document with an Avro schema in components/schemas. It will fail in the playground.

### Expected behavior
The document should parse as it does when the schema is in the message/payload section.

### Sample document
asyncapi: 2.0.0
info:
  title: Example with Avro
  version: 0.1.0
channels:
  example:
    publish:
      message:
        schemaFormat: 'application/vnd.apache.avro;version=1.9.0'
        payload:
          $ref: '#components/schemas/mySchema'
components:
  schemas:
    mySchema:
      type: record
      doc: User information
      fields:
        - name: displayName
          type: string

magicmatatjahu:
Thanks for the issue!

The problem as I see is that currently `schemaFormat` can only be used for `payload`. Using `schemaFormat` in schemas would entail creating something like this: 

```yaml
components:
  schemas:
    mySchema:
      schemaFormat: 'application/vnd.apache.avro;version=1.9.0'
      schema:
        type: record
        doc: User information
        fields:
          - name: displayName
            type: string
```

The solution for me is bad (in sense of user experience) because then user would have to write something like this if he wanted to refer to schema:

```yaml
payload:
  $ref: '#components/schemas/mySchema/schema'
```

Another suggestion would be to use a map for custom schemaFormats:

```yaml
components:

  schemaFormats:
    mySchema: 'application/vnd.apache.avro;version=1.9.0'
    
  schemas:
    mySchema:
      type: record
      doc: User information
      fields:
        - name: displayName
          type: string
```

TBH I don't know how it can be handled in proper way. Let's wait for other people and their insights :)

magicmatatjahu:
I thought a little more about problem and the `schemaFormat` inside the schema shouldn't cause any problems:

```yaml
components:
  schemas:
    mySchema:
      schemaFormat: 'application/vnd.apache.avro;version=1.9.0'
      type: record
      doc: User information
      fields:
        - name: displayName
          type: string
```

By default it will be JSON/Schema (aka AsyncAPI Schema) and in the tooling we can parse and validate schemas using `schemaFormat`.

MichaelDavisSolace:
Yes, I like that.

magicmatatjahu:
[...]

Issue #547, opened 2021-05-28
Replace server.url with server.host


fmvilas:
### Problem

Currently, AsyncAPI servers have a property called `url`. In many cases, we're showing examples of the `url` field as a host instead (without the protocol part of the URL) which is incorrect. On top of that, what's the point of having the protocol included in the `url` field if we already have the `protocol` field? This is an example of what I'm talking about:

```yaml
asyncapi: 3.0.0
servers:
  localhost:
    url: mqtt://localhost
    protocol: mqtt
  production:
    url: kafka://my-production-server.com
    protocol: kafka
```

It can also create inconsistencies where the protocol in the URL and the `protocol` field don't match:

```yaml
asyncapi: 3.0.0
servers:
  localhost:
    url: kafka://localhost
    protocol: mqtt
```

### Proposal

My proposal is that we rename the `url` field as `host`, which would be more correct. Example:

```yaml
asyncapi: 3.0.0
servers:
  localhost:
    host: localhost
    protocol: mqtt
  production:
    host: my-production-server.com
    protocol: kafka
```

---

We may also have to consider adding other parts of the URL like the path and query variables.

This is a proposal to fix #274

dedoussis:
This is indeed a glitch of the existing spec (`2.0.0`). And I think the root cause of the glitch is the fact the `url` field of the [Server Object](https://www.asyncapi.com/docs/specifications/2.0.0#serverObject) supports both absolute and relative URLs. This is a glitch because one cannot simply determine whether the `url` value is relative or not, as the scheme of the URL is expected to be defined through the separate `protocol` field.

The OpenAPI folks were able to support both relative and absolute URLs in their [OpenAPI Server Object](https://swagger.io/specification/#server-object) as the scope of the OpenAPI does not go beyond the HTTP and HTTPS protocols, which conveniently are the protocols one uses to server their API documentation. However, this is not the case in the AsyncAPI world where in most cases `server protocol != API docs endpoint protocol`.

So essentially, this proposal drops support for relative server URLs.

dedoussis:
Instead of dropping the `url` field and introducing new fields (for the different parts of the `url`), I would suggest the inverse approach:
Drop the `protocol` field and enforce the value of the existing `url` field to be an absolute URL.

The enforcement of the absolute url will provide support for all possible URL parts (even for niche parts such as `fragment` or `userinfo`)

To me, this feels like the most simple and future proof approach.

fmvilas:
[...]

PR #597, opened 2021-07-28
feat: allow to define Schema Object in Server variables

magicmatatjahu:
---
title: "Allow to define Schema Object in Server variables"
---

---

**Related issue(s):**
https://github.com/asyncapi/spec/issues/583

---

As I described in the https://github.com/asyncapi/spec/issues/583 issue, I'm wondering why we have inconsistencies for schemas in `server.variables` and `channel.parameters`. This PR introduces change of shape of single Server Variable Object:

- instead of having the 3 words `enum`, `default`, `examples`, which work identically to the JSON Schema keywords, we will now have the word `schema`, which is Schema Object
- `description` is still available, to be consistent with Channel Parameter Object. 
- we can remove `description` field (it is also defined in Schema Object) and flat the `schema` so definition of single variable can have shape like:

  ```yaml
  variables:
    username:
      type: string
      description: This value is assigned by the service provider, in this example `gigantic-server.com`
      default: demo
  ```

- another solution is to add the possibility to define other keywords like `format`, `pattern`, `minLength` etc and for numeric values (integer and number types), but in the current proposal someone can define a schema for server variable in `components/schemas` and we are reusing things that already exist in spec, not creating something from scratch. 
- with current proposal, in the future, when we will implement and resolve problems (https://github.com/asyncapi/spec/issues/528 https://github.com/asyncapi/spec/issues/216), user will be able to define variables in different formats than JSON Schema.

Comments are more than welcome!

BTW. It's a breaking change, so it should be (hopefully) merged in the `3.0.0`.

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -437,15 +437,20 @@ The following shows how variables can be used for a server configuration:
       "protocol": "secure-mqtt",
       "variables": {
         "username": {
-          "default": "demo",
-          "description": "This value is assigned by the service provider, in this example `gigantic-server.com`"
+          "description": "This value is assigned by the service provider, in this example `gigantic-server.com`",
+          "schema": {
+            "type": "string",
+            "default": "demo"
+          }
         },
         "port": {
-          "enum": [
-            "8883",
-            "8884"
-          ],
-          "default": "8883"
+          "schema": {
+            "enum": [
+              "8883",
+              "8884"
+            ],
+            "default": "8883"
+          }
         }
       }
     }
@@ -461,14 +466,16 @@ servers:
     protocol: secure-mqtt
     variables:
       username:
-        # note! no enum here means it is an open value
-        default: demo
         description: This value is assigned by the service provider, in this example `gigantic-server.com`
+        schema:
+          type: string
+          default: demo
       port:
-        enum:
-          - '8883'
-          - '8884'
-        default: '8883'
+        schema:
+          enum:
+            - '8883'
+            - '8884'
+          default: '8883'
 ```
 
 
@@ -480,10 +487,8 @@ An object representing a Server Variable for server URL template substitution.
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="serverVariableObjectEnum"></a>enum | [`string`] | An enumeration of string values to be used if the substitution options are from a limited set.
-<a name="serverVariableObjectDefault"></a>default | `string` | The default value to use for substitution, and to send, if an alternate value is _not_ supplied.
 <a name="serverVariableObjectDescription"></a>description | `string` | An optional description for the server variable. [CommonMark syntax](http://spec.commonmark.org/) MAY be used for rich text representation.
-<a name="serverVariableObjectExamples"></a>examples | [`string`] | An array of examples of the server variable.
+<a name="serverVariableObjectSchema"></a>schema | [Schema Object](#schemaObject) \| [Reference Object](#referenceObject) | Definition of the server variable.
 
 This object MAY be extended with [Specification Extensions](#specificationExtensions).
 


fmvilas:
I think we'll need to specify how to serialize them to strings (probably on channel parameters too) so when someone does something like this:

```yaml
variables:
  example:
    schema:
      type: object
      properties:
        a:
          type: string
          default: 'hello'
```

How is it serialized into the URL? As `{"a": "something"}`? Maybe as `?a=something`?

The reason you don't have a full schema object is to prevent people from using other things than a simple string. We can probably improve it and let people use other basic types like `number` and `boolean` which are pretty straightforward to serialize as a string inside the URL.

Food for thought 🤔

fmvilas:
Also, is there a use case driving this proposal? Would be great to know why you want such a change.

magicmatatjahu:
> The reason you don't have a full schema object is to prevent people from using other things than a simple string. We can probably improve it and let people use other basic types like number and boolean which are pretty straightforward to serialize as a string inside the URL.

I know that `object` and `array` in described case is unnecessary but my PR is like normal Strawman (I think that better is discuss under the PR than issue - you can see changes and how it can works), so using all keywords from Schema isn't good decision. As you wrote and I suggested in PR description, we can allow only a keywords for `string`, `number`, `integer`, `boolean` types - it's another good solution. With this approach we will still have possibility to define the Server Variables in the `components/schemas` and reuse them.

> How is it serialized into the URL? As {"a": "something"}? Maybe as ?a=something?

I don't think so if it should be serialized. In my opinion it's invalid. Like in Parameter Object, currently you have this possibility to define parameter as object (or array), but it's invalid usage, so we should also handled that problem in this proposal, for Server Variable and Parameter Object.

> Also, is there a use case driving this proposal? Would be great to know why you want such a change.

I don't have (for me) explicit use case, but I see that at the moment if you want define more specific constraints for your Server Variable, you must write examples in the `examples` array or describe constraints in the `description` field. When you will have more keywords which you can use, you can even make this:

```yaml
variable:
  type: integer
  minimum: 0
  maximum: 100 
```

rather than:

```yaml
variable:
  examples: [0, 1, 2 ... 100]
```

So at the end (after discussion) we should have answers for question: allow all keywords for the Server Variables? If yes, fine, if not, then we should chose the most important and change also the schema shape for Channel Parameter Object.

magicmatatjahu:
> How is it serialized into the URL? As {"a": "something"}? Maybe as ?a=something?

I am now wondering if in server variable user should have possibility to define query for configuration purposes 🤔  This may be off-proposal but it is quite an interesting topic.

derberg:
@fmvilas what do you mean by `serialized into url`? variables are provided under `url` and then described one by one under `Variables` object. So there is no need to say how is this going to be serialized into the url. As @magicmatatjahu write, we just need to clearly describe-restrict it to basic types only. We do it like this for `query` in WS Binding https://github.com/asyncapi/bindings/tree/master/websockets#channel-binding-object and this is completely fine.

For now I don't like it is not a specific use case driven, but still it is a good change that makes Variables consistent with Parameters. I would compare it to https://github.com/asyncapi/spec/issues/547, so like, nobody cares much (I mean, it is like it is, and things still work), but the change makes sense and we should strongly consider it.

Nevertheless, would be cool to hear community here.

fmvilas:
Yeah, sorry I should have been clearer. I meant what actual strings in the URL should be matching the validation described in the schema. So for instance, if you have a URL template like `ws://localhost/{username}` and you try to connect to a URL like `ws://localhost/["one", 2, false]`, would it be valid given a schema like the following?

```yaml
variables:
  username:
    type: array
```

If so, why are we assuming that URL variables should be deserialized as JSON? Why not a list of values like `ws://localhost/one,2,false`? JSON Schema is meant for validating JSON data but, in this case, I don't think it's JSON but just some strings and maybe with the ability to provide numbers, booleans, etc.

My point is, don't you think it's weird? And don't get me wrong, I think we should align with Parameter Object but probably by removing the `schema` keyword there because of the same issue.

Of course, listening from the community would be needed here. Maybe nobody is using the schema keyword beyond `type: string` and it's safe to remove it. Or maybe not. We need insights before proceeding.

magicmatatjahu:
> ... I don't think it's JSON but just some strings and maybe with the ability to provide numbers, booleans, etc.
>
> My point is, don't you think it's weird? And don't get me wrong, I think we should align with Parameter Object but probably by removing the schema keyword there because of the same issue.
>
> Of course, listening from the community would be needed here. Maybe nobody is using the schema keyword beyond type: string and it's safe to remove it. Or maybe not. We need insights before proceeding.

I agree with you that's it's weird that in the current proposal you can define also the `object` and `array` type, but as you mentioned it's still possible in the Parameter Object, so we should standardize it, I think not only for Server Variable and Parameter but also in the other parts of spec - for example, the message's `header` field has Schema Object shape, but in docs is mentioned that it should use only the `object` type:

image

But however if I define for headers the `type: string` as in example:

image

it still works. I know that spec =/= tooling (parser) but we should handle that problems.

derberg:
>  I think we should align with Parameter Object but probably by removing the schema keyword there because of the same issue.

very good point.

> for example, the message's headers field

Strange, I was fixing it in the schema in the past and it was working as expected -> https://github.com/asyncapi/asyncapi-node/pull/10

Yeah,  just fix the schema here https://github.com/asyncapi/asyncapi-node/blob/master/schemas/2.1.0.json#L644-L657 😄

magicmatatjahu:
@derberg Sorry, you're right, if you define another type that `object` inside the Message Object then you will have error, but if you define in trait it's ok -> https://github.com/asyncapi/asyncapi-node/blob/master/schemas/2.1.0.json#L885-L892 Traits are merged after validation so then I can see in playground: 

image

I will create an issue in the `asyncapi-node` repo.

fmvilas:
> but as you mentioned it's still possible in the Parameter Object, so we should standardize it

Yeah, I agree we should standardize it. I just don't agree to extend server variables but "cut" the parameters object instead. It's weird in both places. And regarding headers, you're right, it's another example. The reason it is like that is that I couldn't find a better way to do it. In this case (including the Parameter Object case), I think our best choice is to actually remove options to keep it simple. My two cents.

magicmatatjahu:
Still valid. We will need this in the next major release.

magicmatatjahu:
@smoya but in ParameterObject you have also `location` field, so `schema` have to be inside ParemeterObject, but it should allows only `string`, `number`, `boolean` types with their constraints.

smoya:
> @smoya but in ParameterObject you have also `location` field, so `schema` have to be inside ParemeterObject, but it should allows only `string`, `number`, `boolean` types with their constraints.

What I understood (maybe wrongly) with that suggestion is that we drop support completely.

magicmatatjahu:
@smoya 

> What I understood (maybe wrongly) with that suggestion is that we drop support completely for using Schema Object (and schema field obviously).

And how people will define constraints for parameter? 😅  Looking at this idea, removing the `schema` field will not be a good idea. We should just limit it to 3 types, `string`, `number` and `boolean` with all their constraints.

smoya:
> @smoya
> 
> > What I understood (maybe wrongly) with that suggestion is that we drop support completely for using Schema Object (and schema field obviously).
> 
> And how people will define constraints for parameter? 😅 Looking at this idea, removing the `schema` field will not be a good idea. We should just limit it to 3 types, `string`, `number` and `boolean` with all their constraints.

No idea. I thought this was part of what it should be done. But as I mentioned, I think I did a wrong assumption so please discard it :)

magicmatatjahu:
At the moment, I don't see the point of keeping that PR open. If I find another way to use Schema Object in Server Variable, I will make an another proposal.

PR #598, opened 2021-07-28
ci: fix update-spec workflow

magicmatatjahu:
As in title: treat injected value from action context as string.


derberg (approved):
🤣

magicmatatjahu:
@derberg Could you check again? I had to update the `github-scripts` action to `v4` version, because that version only read the local scripts from repo.

derberg (approved):

Issue #604, opened 2021-08-04
Inconsistency between Message.examples and MessageTrait.examples


char0n:
**Message.examples** description

> An array of key/value pairs where keys MUST be either headers and/or payload. Values MUST contain examples that validate against the headers or payload fields, respectively. Example MAY also have the name and summary additional keys to provide respectively a machine-friendly name and a short summary of what the example is about.

**Message.trait** description

> An array with examples of valid message objects.

I propose we unify these description and use `Message.examples` description in `Message.trait` description. 

If agreed I can issue a PR.

magicmatatjahu:
@char0n Good catch! `Message Trait`s `examples` should have this same shape as `Message.examples`.  PR will be more than welcome :)

char0n:
PR issued https://github.com/asyncapi/spec/pull/608

char0n:
PR https://github.com/asyncapi/spec/pull/608 has been accepted and merged and became part of 2.2.0.

PR #608, opened 2021-08-10
fix: unify Message and MessageTrait examples description

char0n:
Refs #604

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -1220,7 +1220,7 @@ Field Name | Type | Description
 <a name="messageTraitObjectTags"></a>tags | [Tags Object](#tagsObject) | A list of tags for API documentation control. Tags can be used for logical grouping of messages.
 <a name="messageTraitObjectExternalDocs"></a>externalDocs | [External Documentation Object](#externalDocumentationObject) | Additional external documentation for this message.
 <a name="messageTraitObjectBindings"></a>bindings | [Message Bindings Object](#messageBindingsObject) \| [Reference Object](#referenceObject) | A map where the keys describe the name of the protocol and the values describe protocol-specific definitions for the message.
-<a name="messageTraitObjectExamples"></a>examples | [Map[`string`, `any`]] | An array with examples of valid message objects.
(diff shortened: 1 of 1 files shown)


magicmatatjahu (approved):
LGTM @derberg @fmvilas do you want to add something?

derberg:
it touches the specification, bug fix imho and therefore `fix: ` that will have to trigger 2.1.1 release. We have the  same issue in json schema -> https://github.com/asyncapi/spec-json-schemas/blob/master/schemas/2.1.0.json#L934 instead of https://github.com/asyncapi/spec-json-schemas/blob/master/schemas/2.1.0.json#L699

magicmatatjahu:
@derberg I think that it is editorial change, because we "missed" changing the description of the examples in the Message Trait Object, so `docs:` is enough for me - it other words, Message Trait Object inheritances the shape from Message Object, so we don't change it's shape, but only fix it.

I also agree that json-schemas should be updated.

derberg:
> but only fix it

exactly, we fix it 😄 and fix is so significant that even json schema needs an update. This would be our first patch release so better to wait for @fmvilas opinion too. For now, it is 2.1.1 for me, pure bug fix in the specification, not simple editorial change.

@char0n is it ok for you to provide a fix in a JSON Schema too? Let us discuss there on PR but I guess to avoid "forgetting" things in the future, `examples` should be reused from `definitions`

magicmatatjahu:
> but only fix it

I shouldn't write it 😆  For me it's editorial change.

char0n:
@derberg yep I'll amend the metaschema as well

fmvilas (approved):

char0n:
@derberg schema fixed in https://github.com/asyncapi/spec-json-schemas/pull/85

fmvilas (approved):
LGTM 👍

magicmatatjahu (approved):
LGTM

derberg:
@fmvilas you need to share your opinion on PR title as for me it is `fix:` and for @magicmatatjahu  it is `docs:`

if we agree with `fix: ` we should rather push it to release branch and release in a week together with `2.2` release

fmvilas:
It's indeed a fix. The previous meaning was different than the current meaning.

derberg:
@char0n can you please change base branch to release branch for this one, the one that points to September release?

char0n:
I've changed the base but I'd have to redo the PR so that unrelated commits are not visible here.

derberg:
rebasing with release branch should do the trick, no?

char0n:
PR force-pushed and rebased on top of 2021-09-release

Issue #606, opened 2021-08-04
Proposal to create new Message Example Object


char0n:
Current situation is that `Message.examples` and `MessageTrait.examples` are using generic type of `[Map[string, any]]`. Then the description of `examples` field provide more structure:

> An array of key/value pairs where keys MUST be either headers and/or payload. Values MUST contain examples that validate against the headers or payload fields, respectively. Example MAY also have the name and summary additional keys to provide respectively a machine-friendly name and a short summary of what the example is about.

IMHO this description warrants a new `Message Example Object` with fixed fields, instead of defining the structure with generic type and description. I propose to introduce new `Message Example Object` with following definition:

#### <a name="messageExampleObject"></a>Message Example Object

`<description placeholder>`

##### Fixed Fields

Field Name | Type | Description
---|:---:|---
<a name="messageExampleObjectHeaders"></a>headers | `Map[string, any]` | `<description placeholder>`
<a name="messageExampleObjectPayload"></a>payload | `any` | `<description placeholder>`
<a name="messageExampleObjectName"></a>name | `string` | A machine-friendly name.
<a name="messageExampleObjectSummary"></a>summary | `string` |  A short summary of what the example is about.

---

`Message.examples` and `MessageTrait.examples` type will subsequently change to `[Message Example Object]`

fmvilas:
Yeah, I think it would be a great idea 👍 

Just a comment on the process side of things. For the proposal to be Stage 1 (RFC 1) it has to be a PR, not an issue: https://github.com/asyncapi/spec/blob/master/CONTRIBUTING.md#stage-1-proposal. It's ok as I don't want to be picky but thought you'd like to know 😊

char0n:
> Just a comment on the process side of things. For the proposal to be Stage 1 (RFC 1) it has to be a PR, not an issue: https://github.com/asyncapi/spec/blob/master/CONTRIBUTING.md#stage-1-proposal. It's ok as I don't want to be picky but thought you'd like to know

You're absolute right ;] I'll read the document and align my efforts with the established processes.

char0n:
Formal proposal has been issued via PR: https://github.com/asyncapi/spec/pull/609

char0n:
We can close this. The proposal has been accepted, [merged](https://github.com/asyncapi/spec/pull/609) and became part of 2.2.0.

PR #609, opened 2021-08-10
docs(spec): add RFC for Message Example Object

char0n:
This is formal proposal for introducing new Message Example Object as originally proposed by a strawman in #606. Champion of this RFC proposal is me (@char0n).


spec/asyncapi.md
```diff
@@ -1234,6 +1235,55 @@ schemaFormat: 'application/vnd.apache.avro+yaml;version=1.9.0'
 contentType: application/json
 ```
 
+#### <a name="messageExampleObject"></a> Message Example Object
+
+Message Example Object represent example of a [Message Object](#messageObject) and MUST contain either **headers** and/or **payload** fields.
```
fmvilas, line 1240:
```suggestion
Message Example Object represents an example of a [Message Object](#messageObject) and MUST contain either **headers** and/or **payload** fields. 
```

spec/asyncapi.md
```diff
@@ -1234,6 +1235,55 @@ schemaFormat: 'application/vnd.apache.avro+yaml;version=1.9.0'
...
 
+#### <a name="messageExampleObject"></a> Message Example Object
+
+Message Example Object represent example of a [Message Object](#messageObject) and MUST contain either **headers** and/or **payload** fields. 
+
+##### Fixed Fields
+
+Field Name | Type | Description
+---|:---:|---
+<a name="messageExampleObjectHeaders"></a>headers | `Map[string, any]` | The value of this field MUST validate against the [Message Object](#messageObject) [headers](#messageObjectHeaders) field.
```
fmvilas, line 1246:
```suggestion
<a name="messageExampleObjectHeaders"></a>headers | `Map[string, any]` | The value of this field MUST validate against the [Message Object's headers](#messageObjectHeaders) field. 
```

spec/asyncapi.md
```diff
@@ -1234,6 +1235,55 @@ schemaFormat: 'application/vnd.apache.avro+yaml;version=1.9.0'
...
+#### <a name="messageExampleObject"></a> Message Example Object
+
+Message Example Object represent example of a [Message Object](#messageObject) and MUST contain either **headers** and/or **payload** fields. 
+
+##### Fixed Fields
+
+Field Name | Type | Description
+---|:---:|---
+<a name="messageExampleObjectHeaders"></a>headers | `Map[string, any]` | The value of this field MUST validate against the [Message Object](#messageObject) [headers](#messageObjectHeaders) field. 
+<a name="messageExampleObjectPayload"></a>payload | `any` | The value of this field MUST validate against the [Message Object](#messageObject) [payload](#messageObjectPayload) field.
```
fmvilas, line 1247:
```suggestion
<a name="messageExampleObjectPayload"></a>payload | `any` | The value of this field MUST validate against the [Message Object's payload](#messageObjectPayload) field.
```

fmvilas (changes requested):

spec/asyncapi.md
char0n: Sure, look better and `Message Object` string is still searchable by native browser capabilities (CTRL+F)

fmvilas: Actually, the link takes you to the Message Object. Ok, to the `headers` field but still the Message Object 😄 That's why I did it.

fmvilas (approved):
👍 LGTM

jonaslagoni (approved):
LGTM 👍

smoya (approved):
LGTM!! 🚀 🌔

derberg:
@fmvilas linked proposal is at stage 1. Is this really `docs:` like change? There is also no implementation in the Parser atm (https://github.com/asyncapi/parser-js/issues/321). This change would introduce Parser 2.0 release.

I'm all in with pushing it to the Parser and having 2.0 release, my feeling is that we have here a kind of unofficial fear of major releases, but I say I do not care that 2.0 will only have one new model, we should do it and improve UX and not look back (Jest.js had already 27 releases, React 17 and this is one of their strong points).

anyway, before we merge it, the proposal needs to reach stage 3 imho 🤷🏼

char0n:
@derberg how can I help to push this to stage 3?

fmvilas:
@derberg I don't think this proposal should be linked to the one on the parser. To be honest, this is just a docs reorganization. I understand they're related but IMHO parser implementation of types shouldn't be a blocker for this one. Am I missing something?

derberg:
@fmvilas I think you are right 👍🏼

derberg (approved):

Issue #619, opened 2021-09-09
Invalid Operation References in Example Snippets


balogal:
Hi all,

both v2.1.0 and v2.0.0 of the specification document contain a couple of invalid example snippets. It's always the following snippet or a slight variation of it:

```yaml
user/signedup:
  subscribe:
    $ref: "#/components/messages/userSignUp"
```

This is not valid according to the JSON schema and the specification itself since no reference object may be used in place of an `Operation` object. Let alone a reference to a `Message` object.

The snippet appears in the following sections:
- v2.1.0:
    - https://www.asyncapi.com/docs/specifications/v2.1.0#introduction
    - https://www.asyncapi.com/docs/specifications/v2.1.0#channelsObject
    - https://www.asyncapi.com/docs/specifications/v2.1.0#parametersObject
- v2.0.0:
    - https://www.asyncapi.com/docs/specifications/v2.0.0#introduction
    - https://www.asyncapi.com/docs/specifications/v2.0.0#channelsObject
    - https://www.asyncapi.com/docs/specifications/v2.0.0#parametersObject

fmvilas:
Good catch @balogal. Would you be open to submitting a PR?

balogal:
Hi @fmvilas, there's already a [PR](https://github.com/asyncapi/spec/pull/620) opened alongside this issue.

fmvilas:
Oh sorry, my bad. Thanks a lot 🙏 Left comments in the PR.

PR #620, opened 2021-09-09
docs: fix invalid operation references in example snippets

balogal:
---
title: Fix invalid operation references in example snippets
---

---

**Related issue(s):**
#619 
---

Replaces invalid example snippets with valid ones. I decided to no longer include the message in the channel and parameter object section because I felt like it didn't add much to the understanding. But if you prefer, let me know and I add them as in the intro section.


spec/asyncapi.md
```diff
@@ -537,17 +538,14 @@ Field Pattern | Type | Description
 ```json
 {
   "user/signedup": {
-    "subscribe": {
-      "$ref": "#/components/messages/userSignedUp"
-    }
+    "subscribe": { }
```
fmvilas, line 541: Why not the following?

```suggestion
    "subscribe": {
      "$ref": "#/components/messages/userSignedUp"
    }
```

spec/asyncapi.md
```diff
@@ -537,17 +538,14 @@ Field Pattern | Type | Description
...
   }
 }
 ```
 
 ```yaml
 user/signedup:
-  subscribe:
-    $ref: "#/components/messages/userSignedUp"
+  subscribe: { }
```
fmvilas, line 548: Same as above

spec/asyncapi.md
```diff
@@ -814,9 +812,7 @@ Field Pattern | Type | Description
         }
       }
     },
-    "subscribe": {
-      "$ref": "#/components/messages/userSignedUp"
-    }
+    "subscribe": { }
```
fmvilas, line 815: Same as above

spec/asyncapi.md
```diff
@@ -828,8 +824,7 @@ user/{userId}/signup:
       description: Id of the user.
       schema:
         type: string
-  subscribe:
-    $ref: "#/components/messages/userSignedUp"
+  subscribe: { }
```
fmvilas, line 827: Same as above

fmvilas (changes requested):

spec/asyncapi.md
balogal: You mean:
```json
"subscribe": {
    "message": {
        "$ref": "#/components/messages/userSignedUp"
    }
}
```
right?

I dropped the message here and in the other places you commented because I felt that it didn't add much to the understanding in these sections (explaining the channel item and parameter object).

But definitely not a strong opinion on that one. I changed it to your suggestion.

spec/asyncapi.md
```diff
@@ -538,7 +539,9 @@ Field Pattern | Type | Description
 {
   "user/signedup": {
     "subscribe": {
-      "$ref": "#/components/messages/userSignedUp"
+      "message": {
+        "$ref": "#/components/messages/userSignUp"
```
balogal, line 543: Please note, that I also changed `userSignedUp` to `userSignUp` here and in other locations. Before, both variations where used inconsistently in the spec.

I decided for `userSignUp` because it was used more frequently and it's also the spelling used in the [Message Object](https://www.asyncapi.com/docs/specifications/v2.1.0#messageObject) section.

spec/asyncapi.md
fmvilas: Oh yeah, I forgot the message key again 😅 Yeah, I think it's great to add some context. What you did is perfect IMHO 👍

spec/asyncapi.md
fmvilas: Yeah, thanks for the heads up. IMHO it should be `userSignedUp` (past tense) but I don't think it matters a lot. Thanks for taking care anyway.

fmvilas (approved):

spec/asyncapi.md
smoya: +1 to rename it to use the simple past tense as it is common in events (Domain events, event-sourcing, cqrs, etc)

smoya (dismissed):
LGTM, only one minor about the name of the event. I agree @fmvilas event names are usually written in the simple past tense.

spec/asyncapi.md
balogal: I think originally, `userSignUp` was used, because the example message given in the [Message Object](https://www.asyncapi.com/docs/specifications/v2.1.0#messageObject) section is intended as a command. See for example this part: `"summary": "Action to sign a user up.",`.

In this case, past tense wouldn't be appropriate imho. So if we want to change to past tense, we should adapt a couple of other wordings in the spec. What do you think?

fmvilas: We use both (command and event) interchangeably. I don't think we have to change anything here or in any other place. It's correct as it is, it was just a suggestion because most of the time we're speaking about events instead of commands. That's why we explicitly describe it when it's a command. For the rest of the occurrences, it should be an event. But really, not a blocker. It's ok as it is, at least for me.

balogal: Alright, definitely not a hard opinion on the naming on my part. Even if it is not a blocker, since both of you have a preference towards the past tense names, I reverted the re-naming. So now, everything is as it was before naming wise.

fmvilas (approved):

balogal:
Hi @fmvilas, @smoya, is there currently anything needed from my side to advance this PR? This is my first contribution to the repo so I'm not 100% sure about the processes in place.

derberg:
@all-contributors please add @balogal for doc and bug

Issue #628, opened 2021-09-20
The many meanings of an AsyncAPI file


fmvilas:
As of version 2.x, an AsyncAPI file represents an application. However, as we work on #618, other proposals have arisen to:

1. Make the AsyncAPI file also become a collection of reusable objects that would be referenced from other applications.
2. Make it possible to have a single AsyncAPI file to describe multiple applications. This is especially useful when defining a client and server together.

fmvilas:
I think option 2 would also allow for people to define their whole architecture in a single file. Which might or might not be a good thing. Definitely, something to take into account.

jmenning-solace:
My feeling on this is that:

-  _Async_ encompasses a wide range of use cases, from 1:1 interactions using websockets to n:n solutions using event brokers.  
-  If AsyncAPI aims to be a generic async spec, we need to accommodate all async usecases
-  Because n:n architectures are complex, particularly at scale, it is sometimes desirable to describe the entire architecture, in order to describe the function of a particular component of that architecture

The implications of the above for the spec would be:

- Allow--_but do not mandate_--flexibility on how an AsyncAPI spec is formatted . This would include "library" files, databases, event registries, etc.  However, if the best representation is a single file, the spec should allow it to collapse to that.
- Strictly separate logical relationships from their physical implementations.  That is, use bindings to prevent protocol specific language from entering the main spec. 
- Plan for the complex (brokered) use case, collapse to the simple (1:1) use case, not the other way around (credit to @lornajane for this)

nictownsend:
My thoughts are:

- As we achieve point 2 - we should ensure that it's bidirectional such that tooling can combine and split definitions.
- I should be able to look at the combined document and easily answer the following:
  - How many separate applications are described?
  - Which channels are used by which applications?
  - What messages are read/sent onto the channels by each application?
  - Where are those channels located?
  - If I want to send/receive a message from a specific application, which protocol/server/channel do I connect to?

Perhaps it's as simple as adding an `applications` stanza whereby individual application's operations are listed (I've also included the proposal in #618 for assigning messages to operations):

```
asyncapi: 3.0.0
info:
  title: EDA
  version: 1
  description: The whole application
servers:
  kafka-server: 
    url:
    protocol:
 mq-server:
   url:
   protocol:
channels:
  kakfa-alerts:
    servers:
       - kafka-server
     messages:
       oneOf:
           - message1
           - message2
  mq-messages:
    servers:
       - mq-server
    messages:
      message2
applications:
[...]

Issue #377, opened 2020-04-21
[FEATURE REQUEST] allow $schema


ripienaar:
**Is your feature request related to a problem? Please describe.**

When editing AsyncAPI documents it would be useful to allow $schema since modern IDEs will do code completion etc based on whats valid in the schema.

**Can't it be tackled using specification extensions?**

nope

**Describe the solution you'd like**

Allow $schema key in the top level.

**Describe alternatives you've considered**

Manual editing of IDEs, not nice.

**Additional context**

Publishing the schemas on well known URLs would be good if not already.

I522722:
Hi @ripienaar ,

thanks for bringing this up! This is excactly the problem I'm currently facing as well.

I've prepared this commit as an example how this could be solved at the JSON Schema document level: https://github.com/I522722/asyncapi/commit/dede4318a3425e5541a47832bf2cd87147508f1f

Probably we can make the following distinctions for solving this problem:
* Just allowing the $schema property without any further restrictions or explanations to open up the JSON Schema usecase of the [$schema attribute](https://json-schema.org/understanding-json-schema/reference/schema.html) 
* Further making $schema an enum and therefore a discriminator. Adding `$id` to the Schema Document. See my commit example. This would be still optional to not break the spec. In theory, the existence of a mandatory $schema would make the `asyncapi` attribute obsolete as the link (should) describe both the identity and the version of the document.
* Additionally we could consider publishing it to http://schemastore.org/json/. OpenAPI e.g. is already published there.

derberg:
For others interested in this topic. There is a lot of clarification here https://asyncapi.slack.com/archives/C34F2JV0U/p1588926494498200

I522722:
For the simplest intermediary solution, we could just add `$schema` without a `const` or `enum` restriction to the top-level properties. Since it's just a technical attribute imho it does not need to be added to the human readable documentation.

```json
  "$schema": {
      "type": "string",
      "format": "uri",
      "description": "The $schema keyword is used to declare that this JSON document is validated against the JSON Schema of the AsyncAPI 2.0.0 Specification. Adding this allows for automated tool support and code intelligence. See https://json-schema.org/understanding-json-schema/reference/schema.html"
  } 
```

This would be a very quick fix to enable the use case. There could be a follow-up issue whether this should be taken up to be more explicitly supported and e.g. the Schema officially published at a schema registry.

[...]

PR #651, opened 2021-11-11
docs: small social media system example

### LINKED ISSUE #646: Uniform example with different use-cases
As the discussions become more and more complex I think it would make sense to uniform an example "system" to use as a reference for changes and suggestions.

I will suggest we use the example @fmvilas included in the proposal for [solving publish/subscribe confusion](https://github.com/asyncapi/spec/issues/618). 

This example takes into consideration of the different perspectives and use-cases we see AsyncAPI being used in. This does of course not mean any changes to the spec need to provide changes to this example, however, if it makes sense, people only need to know about one specific example to understand the changes that are being suggested, which I think makes it easier to understand the issue and the solutions proposed.

### The proposed example 
Say we have a social network, a very basic one, where all you can do is upvoting comments.

The individual services that need to be defined are:
 - A frontend application where users interact through the website, where they can _like_ comments. 
 - A backend WebSocket server that sends and receives events for the UI to update in real-time.
 - A comment service which processes all events related to comments through a message broker.
 - A public-facing API which allows others to get notified about updates.

This ensures that we try to describe: 
- A public API
- A WebSocket server and client
- A broker setup 

### Suggestions
What do you think?
- Do you see this as being useful to have?
- Is there any changes to the system you would like to see?

### Related resources
- https://github.com/asyncapi/glee/pull/174
- https://github.com/asyncapi/spec/issues/618
- https://github.com/asyncapi/spec/issues/628

jonaslagoni:
This PR introduces the basic social media system that we can use as a baseline for features and improvement discussions as it introduces a lot of different perspectives and how AsyncAPI can be used.

It is based on @fmvilas example in https://github.com/asyncapi/spec/issues/618, however, adapted to 2.2.0, as there is no need to wait to introduce it.

Fixes #646


examples/social-media/readme.md
```diff
@@ -0,0 +1,24 @@
+## Small social media example
+
+Say we have a social network, a very basic one, where all you can do is upvoting comments.
+
+The individual services that need to be defined are:
+
+- A frontend application where users interact through the website, where they can like comments.
```
smoya, line 7: Linking to the documents will be awesome

examples/social-media/readme.md
```diff
@@ -0,0 +1,24 @@
...
+- A backend WebSocket server that sends and receives events for the UI to update in real-time.
+- A comment service which processes all events related to comments through a message broker.
+- A public-facing API which allows others to get notified about updates.
```
smoya, line 10: It is not clear enought for me what this public facing API would do. Here you mention `allows others to get notified about updates`, however in the `public.asyncapi.yaml` file, there is only one operation which is `publish`. Shall we add a `Subscribe` Operation so users can be notified?

jonaslagoni: The public API needs to be seen as an external application, that anyone can have and interact with the system. We still use `publish` because that application need to consume messages, where others publish to it 🙂

examples/social-media/readme.md
jonaslagoni: Added 🙂

examples/social-media/readme.md
```diff
@@ -0,0 +1,25 @@
...
+
+<p align = "center">
+Fig.1 - Mermaid diagram for the system, partly generated using Cupid.
```
derberg, line 17:
```suggestion
Fig.1 - Mermaid diagram for the system, partly generated using [Cupid](https://github.com/asyncapi/cupid).
```

examples/social-media/dictionary.asyncapi.yaml
```diff
@@ -0,0 +1,59 @@
+asyncapi: 2.2.0
+
+info:
+  title: Dictionary for organization-wide definitions
```
derberg, line 4: This one is super confusing for me 😄 
Kinda reverse the "official" recommendations we always had that align with the spec where `app` is in the center.
I would never advise users to do it this way. I know people do it. I was involved in PRs where we had to make it possible to read info from `components` that is not referenced in `channels`. I would advise users to store schemas as separate files, not related to asyncapi and use `$ref` to them. 
And yeah, it is not addressed in the readme 😄 

Nothing is black or white, I know, but I have a feeling that in official example it should be clearly described in the readme, why it is not this way or another event though the spec say this and that

derberg (changes requested):
First great complex example!
Would make sense to adopt https://github.com/amadeus4dev/amadeus-async-flight-status some day too

examples/social-media/dictionary.asyncapi.yaml
jonaslagoni: Hmm... 

I see your point 😆 However, for simplicity, having everything in one file, just makes it easier. I could change the file to not follow AsyncAPI structure?


`dictionary.asyncapi.yaml` -> `dictionary.yaml`

```yaml
components:
  messages:
    commentLiked: 
      payload: 
        $ref: '#/components/schemas/commentLikedPayload'
    likeComment: 
      payload: 
        $ref: '#/components/schemas/likeCommentPayload'
    commentChanged: 
      payload: 
        $ref: '#/components/schemas/commentChangedPayload'
    updateCommentLikes: 
      payload: 
        $ref: '#/components/schemas/updateCommentLikesPayload'
  schemas: 
    commentLikedPayload: 
      type: object
      title: commentLikedPayload
      additionalProperties: false
      properties:
        commentId: 
          allOf: 
            - $ref: '#/components/schemas/commentId'
            - description: Id of the comment that was liked
    likeCommentPayload: 
      type: object
      title: likeCommentPayload
      additionalProperties: false
      properties:
        commentId: 
          allOf: 
            - $ref: '#/components/schemas/commentId'
            - description: Id of the comment that should be liked
    commentChangedPayload: 
      type: object
      title: commentChangedPayload
      additionalProperties: false
      properties:
        commentId: 
          allOf: 
            - $ref: '#/components/schemas/commentId'
            - description: Id of the comment that was changed, such as when someone liked it.
    updateCommentLikesPayload:
      type: object
      title: updateCommentLikesPayload
      additionalProperties: false
      properties:
        commentId: 
          allOf: 
            - $ref: '#/components/schemas/commentId'
            - description: Id of the comment that was changed, such as when someone liked it.
    commentId:
      type: string

```

examples/social-media/readme.md
jonaslagoni: This won't work in an HTML wrapped tag (`p`), gotta use `a`. Changed.

examples/social-media/dictionary.asyncapi.yaml
derberg: still don't get the name like `dictionary` but as long as it is not `asyncapi` file, that is fine. It only still follows `asyncapi` naming and this can confuse, how about 2 files, one called `shared-messages.yaml` and the other `shared-schemas.yaml` and you do not need `components.messages` and stuff like that

derberg:
as for the filenames, can we promote here `asyncapi.yaml ` name everywhere?

jonaslagoni:
> as for the filenames, can we promote here `asyncapi.yaml ` name everywhere?

@derberg do you mean switching from `backend.asyncapi.yaml` -> `backend/asyncapi.yaml` 🤔?

examples/social-media/dictionary.asyncapi.yaml
jonaslagoni: You dont? 🤔 I thought it would be pretty clear of a file where you lookup definitions, i.e. looking up words in a dictionary to find their meaning 😄 

I see your point... Not sure what would make the most sense for others 🤔

derberg: for me dictionary is dictionary 😄 notice that when people talk about schema registry solutions they do not call them schema dictionaries. Just my point of view

jonaslagoni: @derberg do you mind writing your comments in https://github.com/asyncapi/spec/issues/628#issuecomment-968050476 🙂

jonaslagoni:
> as for the filenames, can we promote here `asyncapi.yaml ` name everywhere?

I only think that this is appropriate if we plan to include implementation as well. Otherwise, I think it makes it less readable, and annoying that they are located in separate directories.

derberg:
> I only think that this is appropriate if we plan to include implementation as well. Otherwise, I think it makes it less readable, and annoying that they are located in separate directories.

yeah, my opinion is that people are not so perceptive when learning to notice your example is done this way and not another because there is no code. Imho example should always push best practices. In the real project, this is how it should look like, asyncapi file per folder, not all in one. Also in tools we promote `asyncapi` as file name, in CLI it is supported and auto-discovered and also in the future it is important for https://github.com/asyncapi/spec/issues/377 (I still plan to complete it this year as I wrote that I will do it, December last year 😆 )

jonaslagoni:
Alright, changing it 👍

jonaslagoni:
> Imho example should always push best practices. In the real project, this is how it should look like, asyncapi file per folder, not all in one

That's all good fine until you want to introduce reusability. If we assume each project is a git repository (very likely). If you do `asyncapi file per folder`, you would have to maintain something like git-submodules, or remote hosting of references for the common definitions. I have not used `asyncapi file per folder` since I first started using references.

So about the common file, where we look up stuff, should that be located under `./common/messages.yaml`, etc?

derberg:
Last time I did microservices in production, it was indeed a separate repo per project @jonaslagoni 
In monorepo you would also get each project in a separate folder (submodules is not a must)

`./common/messages.yaml` sound good

@boyney123 you might want to have a look at this one, it adds more complex structure to examples, not sure how it affects your concept of "starting new file from an existing example"

examples/social-media/dictionary.asyncapi.yaml
smoya: I heard the word `catalog` in the past as an alternative somewhere. I think I agree @derberg the name `dictionary` is not the best either.

jonaslagoni: It has been changed now to two separate files, and the dictionary has been removed.

boyney123:
> @boyney123 you might want to have a look at this one, it adds more complex structure to examples, not sure how it affects your concept of "starting new file from an existing example"

Yeah at the moment it just renders things on a 121 basis, but no real reason why this cannot be supported would just need thinking about, or for now we just don't pull it over in the examples when the GitHub action triggers, but yeah would need thinking about.

At the moment nothing would break 👍 , so good to think about but don't let it block it

boyney123:
> sorry you mean that you basically read entire dir and get asyncapi files only? or that you read only from dir root?

Yeah the idea is to copy over the `examples` directory into any project that uses them (using GitHub Actions), the assumption would of been that any file in that directory was an AsyncAPI file, which means we could just read them and assume.

With the new pattern that would not be the case, so would just need to rethink how that would work, but it's possible.

examples/social-media/common/schemas.yaml
```diff
@@ -0,0 +1,38 @@
+commentLikedPayload: 
+  type: object
+  title: commentLikedPayload
+  additionalProperties: false
+  properties:
+    commentId: 
+      allOf: 
+        - $ref: '#commentId'
```
ceich, line 8:
```suggestion
        - $ref: '#/commentId'
```

examples/social-media/common/schemas.yaml
```diff
@@ -0,0 +1,38 @@
...
+    commentId: 
+      allOf: 
+        - $ref: '#/commentId'
+        - description: Id of the comment that was changed, such as when someone liked it.
+updateCommentLikesPayload:
+  type: object
+  title: updateCommentLikesPayload
+  additionalProperties: false
+  properties:
+    commentId:
```
ceich, line 33: I think a `likeCount` should be included as well.

examples/social-media/common/schemas.yaml
jonaslagoni: Good catch @ceich, thanks!

examples/social-media/common/schemas.yaml
jonaslagoni: There are definitely multiple things that could be added to make the system more believable. However, we gotta stop somewhere, and I just decided to take a minimal approach.

I guess my question is: would this minimal payload not show what is needed? 🤔

ceich: It requires a separate call to update the count, and there are race conditions. I don't feel strongly as it's just an example.

derberg: ideal example would be actually something that we can use in code generators to generate real project

jonaslagoni: @derberg so you also think we should add more details to the messages?

derberg: hahaha, when I clicked on `Add single comment` I had a quick thought in my head "this one can be confusing" 😄 

basically, when I generate code with these examples, in frontend app when I add logic to react on event about comments count, I will not have access to actually the count information. Thus from a code generation perspective, this is not a full example.

Does it make sense?

jonaslagoni: Hmm, that is a good point...

jonaslagoni: Alright, I have added a `likedBy` property to the `likeCommentPayload`, and `likeCount` to the two schemas `commentChangedPayload`, `updateCommentLikesPayload`.

Not sure if `likedBy` makes sense 🤔 Maybe better to be left out. What do you think?

examples/social-media/common/schemas.yaml
```diff
@@ -16,6 +16,10 @@ likeCommentPayload:
       allOf: 
         - $ref: '#/commentId'
         - description: Id of the comment that should be liked
+    likedBy:
+      allOf: 
+        - $ref: '#/commentId'
```
ceich, line 21:
```suggestion
        - $ref: '#/userId'
```

ceich (changes requested):

examples/social-media/common/schemas.yaml
jonaslagoni: 🤦

ceich (approved):

derberg (approved):

derberg:
@jonaslagoni are you merging this one?

jonaslagoni:
@derberg right...

For some reason, I still have the thought that code owners do the ready for merge because they are satisfied with the state. I guess this lies rather with the PR owners instead then 😄

derberg:
actually you are right 🤦🏼 I could merge too, I somehow mixed stuff up, / r t m is there to help but still owners can merge 😆

derberg:
@allcontributors please add jonaslagoni for example

Issue #649, opened 2021-11-10
Unify reference and bundling behavior with JSON Schema


smoya:
## Context

Current AsyncAPI [Reference Object](https://www.asyncapi.com/docs/specifications/v2.2.0#referenceObject) documentation mentions:

> For this specification, reference resolution is done as defined by the [JSON Reference](https://tools.ietf.org/id/draft-pbryan-zyp-json-ref-03.html) specification and not by the JSON Schema specification.

On the other hand, [JSON Schema Draft-07](https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-01), the version current AsyncAPI spec relies on, **does not** follow JSON Reference but rather its own implementation.

[JSON Schema Draft-07](https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-01) includes the `$id` keyword which modifies the behavior of `$ref`. It allows defining a base URI for the `$ref` keyword, where referenced URIs will be taken as relative to the one in `$id`.

On the contrary, `$id` property is not allowed on any AsyncAPI object.

## Why is this a problem?

We **do not** have a unified reference object behaviour between AsyncAPI [Reference Object](https://www.asyncapi.com/docs/specifications/v2.2.0#referenceObject) and [JSON Schema Draft-07](https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-01). That impacts directly the user experience.

Tooling **must** support two different dereferencing and bundling behaviors:
  1. One for AsyncAPI [Reference Object](https://www.asyncapi.com/docs/specifications/v2.2.0#referenceObject).
  2. Another one for [JSON Schema Draft-07](https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-01).

The following issues have been created:

- https://github.com/asyncapi/parser-js/issues/403
- https://github.com/asyncapi/parser-js/issues/404

Tooling availability is a problem as well. As there is no standard spec we all (AsyncAPI, OpenAPI, etc) can follow, tooling will need to be customized.

### 1. Porting the same dereference + bundling resolution behavior as JSON Schema does
For the case of [JSON Schema Draft-07](https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-01), will be adding support for `$id` keyword.

In the case we decide to eventally update to [JSON Schema Draft 2020-12](https://github.com/asyncapi/spec/issues/596), some other keywords and behaviors will be supported, such as allowing extra properties when using `$ref`.

[...]

PR #653, opened 2021-11-12
docs: clarify bundling behavior on Schema Object

smoya:
This PR clarifies the behavior of bundling and (de)referencing of Schema Object, introducing a clear statement and reminder that the `$ref` behavior does not follow JSON Schema one, therefore things like `$id` won't have any effect on it.

**Related issue(s):**

https://github.com/asyncapi/spec/issues/649


spec/asyncapi.md
```diff
@@ -1670,6 +1669,7 @@ The following properties are taken from the JSON Schema definition but their def
 - default - The default value represents what would be assumed by the consumer of the input as the value of the schema if one is not provided. Unlike JSON Schema, the value MUST conform to the defined type for the Schema Object defined at the same level. For example, of `type` is `string`, then `default` can be `"foo"` but cannot be `1`.
 
 Alternatively, any time a Schema Object can be used, a [Reference Object](#referenceObject) can be used in its place. This allows referencing definitions in place of defining them inline.
+It is appropriate to clarify that the `$ref` keyword will follow the behavior described by [Reference Object](#referenceObject) **and not** the one in JSON Schema definition. Therefore, `$id` keyword won't modify the behavior of `$ref`.
```
jonaslagoni, line 1672: To further clarify this problem people will need to wrap their head around what `$id` is and do (we don't describe any of this in the schema object), what JSON Schema is, how it relates to the default super-set of JSON Schema, and this is not relevant to users if they aren't using JSON Schema at all.

Not sure if we can really clarify this, on the other hand, it is kinda needed... Idk 😆 

Did that make sense? 😅

smoya: I think I see your point. However, I would like to clarify some things first:

> we don't describe any of this in the schema object

In the "Schema Object" section, we mention it is a superset of JSON Schema Draft-07, including a link to the core and validation documents. Also, we have this statement `As such, any keyword available for those vocabularies is by definition available in AsyncAPI, and will work the exact same way, including but not limited to`

So IMHO it is clear there is a relation and the fact we directly depend on JSON Schema Draft-07 spec.  

Is it enough with just having those lines? Do you suggest we expand the information further?

jonaslagoni: But it also states further down:

> Alternatively, any time a Schema Object can be used, a Reference Object can be used in its place. This allows referencing definitions in place of defining them inline.

Which is even more confusing, when they have nothing to do with each other 😆

smoya: I removed the reference to `$id` plus removed the linebreak so it stick all together. Also added a link to the JSON Schema reference docs. Hope that makes more sense now @jonaslagoni !

smoya:
cc @derberg @fmvilas you might be interested in this one! Thank you!

spec/asyncapi.md
```diff
@@ -1669,7 +1668,7 @@ The following properties are taken from the JSON Schema definition but their def
 - format - See [Data Type Formats](#dataTypeFormat) for further details. While relying on JSON Schema's defined formats, the AsyncAPI Specification offers a few additional predefined formats.
 - default - The default value represents what would be assumed by the consumer of the input as the value of the schema if one is not provided. Unlike JSON Schema, the value MUST conform to the defined type for the Schema Object defined at the same level. For example, of `type` is `string`, then `default` can be `"foo"` but cannot be `1`.
 
-Alternatively, any time a Schema Object can be used, a [Reference Object](#referenceObject) can be used in its place. This allows referencing definitions in place of defining them inline.
+Alternatively, any time a Schema Object can be used, a [Reference Object](#referenceObject) can be used in its place. This allows referencing definitions in place of defining them inline. It is appropriate to clarify that the `$ref` keyword will follow the behavior described by [Reference Object](#referenceObject) **and not** the one in [JSON Schema definition](https://json-schema.org/understanding-json-schema/structuring.html#ref).
```
fmvilas, line 1671: A suggestion to use a more formal language:

```suggestion
Alternatively, any time a Schema Object can be used, a [Reference Object](#referenceObject) can be used in its place. This allows referencing definitions in place of defining them inline. It is appropriate to clarify that the `$ref` keyword MUST follow the behavior described by [Reference Object](#referenceObject) instead of the one in [JSON Schema definition](https://json-schema.org/understanding-json-schema/structuring.html#ref).
```

fmvilas (changes requested):

spec/asyncapi.md
smoya: Good suggestion. Applied!

jonaslagoni (approved):

fmvilas (approved):

PR #666, opened 2021-12-06
feat: add Solace bindings to the list of supported bindings in the spec.

damaru-inc:
---
title: "Add Solace bindings to the list of supported bindings in the spec."
---
This is a request to add the Solace bindings to the lists of bindings in the spec.

The PR for the json schema repo corresponding to this PR is here: https://github.com/asyncapi/spec-json-schemas/pull/136

### DIFF
--- a/.github/workflows/if-nodejs-release.yml
+++ b/.github/workflows/if-nodejs-release.yml
@@ -21,6 +21,10 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
     steps:
+      - name: Set git to use LF #to once and for all finish neverending fight between Unix and Windows
+        run: |
+          git config --global core.autocrlf false
+          git config --global core.eol lf
       - name: Checkout repository
         uses: actions/checkout@v2
       - name: Check if Node.js project and has package.json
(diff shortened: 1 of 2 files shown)


jessemenning:
Looks good to me, excited to have Solace join the bindings.

derberg (changes requested):
@MichaelDavisSolace please also update the sections I shared in below screen shots

(1)
Screenshot 2021-12-21 at 16 52 27
(2)
Screenshot 2021-12-21 at 16 52 38
 
@fmvilas IMHO we can move it to `Draft` stage as PR for schema is also there. There are some fixes needed for this PR but this is why it is stage 2 not 3

fmvilas:
Yeah, I think this can be marked as Draft already. @MichaelDavisSolace I think the next step would be to get the JSON Schema version into the JS Parser and we'd be done. In the parser, you can point to your branch temporarily so it can be tested.

damaru-inc:
> Yeah, I think this can be marked as Draft already. @MichaelDavisSolace I think the next step would be to get the JSON Schema version into the JS Parser and we'd be done. In the parser, you can point to your branch temporarily so it can be tested.

Okay, will do.

damaru-inc:
I see Dale Lane is the one who got the 2.3.0 schema into the spec-json-schemas repo. Does that mean that he'll be the one to update parser-js?

derberg (approved):
@fmvilas I approved this one and promoted to `accepted`

Imho https://github.com/asyncapi/spec-json-schemas/pull/143 is enough for now and agree with one of the comments Michael did in Slack that in case of this feature, where there is just small change in the schema, it doesn't make sense to open up a PR before https://github.com/asyncapi/spec-json-schemas/pull/143 is actually merged and the release candidate is ready.

So @fmvilas once you approve this one and release it, then release PR for JSON Schema, please just open a bump PR in the Parser to the release branch...or just merge together with https://github.com/asyncapi/parser-js/pull/430 and @smoya will bump it

fmvilas (approved):
Yeah, makes sense to me. Approving 👍

PR #670, opened 2021-12-08
docs: extend example for kafka with x509 security

derberg:
this is driven by https://github.com/asyncapi/nodejs-template/pull/103/files PR where I had to add customized asyncapi file just with one tiny change, could be better to extend official example and use it in test instead of custom version

I think we still have this unwritten rule that best if templates work with official examples, if only possible

### DIFF
--- a/examples/streetlights-kafka.yml
+++ b/examples/streetlights-kafka.yml
@@ -21,6 +21,7 @@ servers:
     description: Test broker
     security:
       - saslScram: []
+      - certs: []
 
 defaultContentType: application/json
 
@@ -139,6 +140,9 @@ components:
     saslScram:
       type: scramSha256
       description: Provide your username and password for SASL/SCRAM authentication
+    certs:
+      type: X509
+      description: Download the certificate files from service provider
 
   parameters:
     streetlightId:


dalelane (commented):
Putting this in the spec would say that client applications could either use mutual TLS or SASL/SCRAM to authenticate with the Kafka cluster - using the same address for either.

More typically when using different auth approaches with Kafka, you'd have different URLs (in Kafka jargon, "listeners") for each security method.

Which would make the spec more like:

```yaml
servers:
  scram-connections:
    url: test.mykafkacluster.org:18092
    protocol: kafka-secure
     description: Test broker
     security:
       - saslScram: []
  mtls-connections:
    url: test.mykafkacluster.org:28092
    protocol: kafka-secure
     description: Test broker
     security:
       - certs: []
components:
  securitySchemes:
     saslScram:
       type: scramSha256
       description: Provide your username and password for SASL/SCRAM authentication
     certs:
       type: X509
       description: Download the certificate files from service provider
```

I touched base with @nictownsend as well, and we don't actually know how you could configure a single Kafka listener to support both mutual TLS and SASL/SCRAM. We've not tried, so we're not 100% certain, but we suspect it wouldn't be possible to setup Kafka to support the spec as it is in the PR. (We know you can get a single listener to support different SASL mechanisms, so you could support both SASL/PLAIN and SASL/SCRAM on a single URL for example. But not different protocols like SASL and mTLS).

derberg:
@dalelane thanks for the review, I update example by following your advice

dalelane (approved):

derberg:
@fmvilas need your approval mate
@dalelane 🙇🏼

nictownsend (approved):

fmvilas (approved):

fmvilas:

derberg:

PR #672, opened 2021-12-08
chore: fix codeowners file to enable people to also decide and not just bot :smile:

derberg:
I have to admit that my initial CODEOWNER configuration was not the best, like future-proof, especially now for the permissions cleanup


fmvilas:
Sorry for the noise, was just playing with slash commands 😄

fmvilas (approved):

Issue #660, opened 2021-11-30
Allow `servers` and `channels` to be defined inside `components`


smoya:
## Context

This is part of the RFC https://github.com/asyncapi/spec/issues/618#issuecomment-980093487. 
In an effort around reducing the scope of such RFC, it has been split into several proposals so they can move forward independently.

## The problem

Users can't define `servers` and `channels` as components. Meaning this has an impact on reusability. 
In the scenario of https://github.com/asyncapi/spec/issues/654 and https://github.com/asyncapi/spec/issues/658, reusing components is more than a requirement.
This will allow users to reference both `servers` and `channels` so the following will be now valid:

```yaml
servers:
  production:
    $ref: '#/components/servers/myserver'
channels:
  some/events:
    $ref: '#/components/channels/myChannel'
components:
  servers:
    myserver:
      url: "http://localhost:5000/ws"
      protocol: ws
  channels:
    myChannel:
      description: "mychannel"
```

## Suggestion

To allow `servers` and `channels` to be defined inside `components`.
As far as I see, this won't be considered as breaking change.

Depends on https://github.com/asyncapi/spec/issues/650

smoya:
I'm championing this one if no one is against it.

jonaslagoni:
I think you have to include part of #650 so that servers and channels can be referenced, otherwise there is no need to define it in components 🙂

smoya:
Some work has been done now:

- PR changing Spec docs: https://github.com/asyncapi/spec/pull/665
- PR changing Spec JSON Schema schemas: https://github.com/asyncapi/spec-json-schemas/pull/133

Pending: work on Parser-js

smoya:
I'm gonna ping some people I think might be interested in this. Please do not feel you have to review this, just to raise some awareness :) 

cc @fmvilas @GeraldLoeffler @damaru-inc @jessemenning @magicmatatjahu

fmvilas:
Can we clarify that this is depending on https://github.com/asyncapi/spec/issues/650? Otherwise, as Jonas said, it's useless, right?

smoya:
> Can we clarify that this is depending on #650? Otherwise, as Jonas said, it's useless, right?

The current changes in both opened PRs (as per today) consider https://github.com/asyncapi/spec/issues/650. For example, https://github.com/asyncapi/spec-json-schemas/pull/133 adds `$ref` field to the server object.
I will add a link to the description of this Issue as this is what I guess you are missing.

[...]

Issue #588, opened 2021-07-26
Ambiguity in `Operation.message` field


char0n:
Hi everybody,

I've intercepted a possible ambiguity and un-clarity within the specification (2.1.0) regarding `Operation.message` fixed field.

Type of `Operation.message` fixed field is defined as `[Message Object | Reference Object]` which means that the field is going to be a list and either Message Object or Reference Object is allowed within it.

### Inconsistent description

First sentence in the description then says `A definition of the message that will be published or received on this channel.` Other fields descriptions containing list of union types start with following sentence: `A list of ...`. So I propose we change the first sentence of description to `A list of Message Objects that will be published or received on this channel.`

### Allowing additional type in union not specified in Type field

Second sentence in description says: `oneOf is allowed here to specify multiple messages`. This has very serious implications. In order for this to be true we have to amend the Type of `Operation.message` to `[Message Object | Reference Object | Schema Object]`. From the perspective of usage this doesn't really make sense as we already have this mechanism in `[Message Object | Reference Object]`. I suggest to eliminate this sentence entirely to reduce complexity. The alternative ca be defining the type as `[Message Object | Reference Object | Schema Object]`, but as argued there is no additional value in this.

### Invalid example

[This example](https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md#operation-object-example) contains invalid usage of `Operation.message` field. The value of the field in example is defined as Message Object. This shouldn't be possible according to the current spec.

```json
{
  "operationId": "registerUser",
  "summary": "Action to sign a user up.",
  "description": "A longer description",
  "tags": [
    { "name": "user" },
    { "name": "signup" },
    { "name": "register" }
  ],
  "message": {
    "headers": {
      "type": "object",
      "properties": {
        "applicationInstanceId": {
          "description": "Unique identifier for a given instance of the publishing application",
          "type": "string"
        }
      }
    },
    "payload": {
      "type": "object",
      "properties": {
        "user": {
          "$ref": "#/components/schemas/userCreate"
        },
        "signup": {
          "$ref": "#/components/schemas/signup"
        }
      }
    }
  },
  "bindings": {
    "amqp": {
      "ack": false
    },
  },
  "traits": [
    { "$ref": "#/components/operationTraits/kafka" }
  ]
}
``` 

I can issue a PR with the changeset if we agree on these points.

magicmatatjahu:
Exactly, we have a bug with type of `Operation.message` field, but

[...]

PR #681, opened 2021-12-27
feat: release version 2.3.0

derberg:
This PR is created upfront to reflect on a daily basis what things are included in the release.

Also, the reason to create this PR a long time before the release is to enable automation (bot keeps upstream branches always up to date with master) that we have in place to regularly update the release branch with whatever is changed in the master branch. So nobody has to do it manually. 

Resolves https://github.com/asyncapi/spec/issues/660
Fixes https://github.com/asyncapi/spec/issues/588


fmvilas (approved):
🚀

Issue #661, opened 2021-11-30
Make `channels` field optional within an AsyncAPI file.


smoya:
## Context

This is part of the RFC https://github.com/asyncapi/spec/issues/618#issuecomment-980093487. 
In an effort around reducing the scope of such RFC, it has been split into several proposals so they can move forward independently.

## The problem

From https://github.com/asyncapi/spec/issues/628#issuecomment-968050476: 

> One way we have seen the specification being used is in a Dictionary type of AsyncAPI document, which is a type of pattern for defining definitions for the common information, shared by each separate application.

However, users have been forced to declare an empty object as a value for the `channels` field since this is not optional.

## Suggestion

To make `channels` field optional. 
It depends on https://github.com/asyncapi/spec/issues/660, because without it perhaps it won't make a lot of sense.

derberg:
I think this one is pretty easy to push for January 2022 release. Change is small, and I know people use AsyncAPI just to list messages in components, and they do not need `channels` object and just leave it empty. 

@Fannon not sure if you are the one that uses AsyncAPI this way or someone else from SAP? don't remember 😅

Fannon:
@derberg : What do you mean with "this way" excactly?
In the case of using CloudEvents, having channels with topic names may not  bean ideal fit because with CloudEvents you would subscribe via "type" or "source" properties. A topic / channel then becomes an implementation detail or may not fit very well.

However, channels still make sense to indicate which events from messages are actually subscribable / actively sent (as opposed to events which are technically supported, but not "active").

If you need any more background, I can get in contact with my colleagues. I'm not working on this directory, but I'm in contact with the colleagues who do.

derberg:
oh sorry, so I meant that someone from SAP uses AsyncAPI in a way that channels are just `channels: {}` and then just messages are listed in components. But yeah, I forgot how big SAP is and that probably it was someone not related to the work that you are doing 😆 

this is the case I was talking about https://github.com/asyncapi/parser-js/issues/210

smoya:
PR's has been created:

- https://github.com/asyncapi/spec/pull/682
- https://github.com/asyncapi/spec-json-schemas/pull/146

smoya:
As per [asyncapi/spec-json-schemas#146 (comment)](https://github.com/asyncapi/spec-json-schemas/pull/146#issuecomment-1016290582), we won't include this change in `2.3.0` but rather in `3.0.0`.

smoya:
[...]

PR #682, opened 2022-01-04
feat: make channels optional

smoya:
**Description**

Part of https://github.com/asyncapi/spec/issues/661.

This PR removes the `required` prefix from the `channels` field at root level of the document.

**Related issue(s)**
https://github.com/asyncapi/spec/issues/661

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -156,7 +156,7 @@ Field Name | Type | Description
 <a name="A2SInfo"></a>info | [Info Object](#infoObject) | **Required.** Provides metadata about the API. The metadata can be used by the clients if needed.
 <a name="A2SServers"></a>servers | [Servers Object](#serversObject) | Provides connection details of servers.
 <a name="A2SDefaultContentType"></a>defaultContentType | [Default Content Type](#defaultContentTypeString) | Default content type to use when encoding/decoding a message's payload.
-<a name="A2SChannels"></a>channels | [Channels Object](#channelsObject) | **Required** The available channels and messages for the API.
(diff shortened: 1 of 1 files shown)


magicmatatjahu (approved):

smoya:
cc @derberg @fmvilas as you both are code owners.

jonaslagoni:
Is this not a breaking change? 🤔

magicmatatjahu:
@jonaslagoni 

> Is this not a breaking change? 🤔

If we switch optional to required then yes, but switching required -> optional won't break any AsyncAPI document.

fmvilas:
> Is this not a breaking change? 🤔

It will not be backward compatible but it will be forward compatible. I think we should also make it clear in the spec, that we aim for forward compatibility.

fmvilas (approved):
Looks good to me. Let's wait for @derberg.

jonaslagoni:
> Looks good to me. Let's wait for @derberg.

see https://github.com/asyncapi/spec/issues/688

fmvilas (changes requested):
I thought twice about it and think this one is a breaking change and should be postponed to v3.

dalelane:
> I thought twice about it and think this one is a breaking change and should be postponed to v3.

I'll have a look to see if there is anything else that needs to be pulled out of 2.3 release branches

smoya:
As per https://github.com/asyncapi/spec-json-schemas/pull/146#issuecomment-1016290582, we won't include this change in `2.3.0` but rather in `3.0.0`.

Issue #520, opened 2021-03-19
Confusions with the Publish and Subscribe meaning/perspective


fmvilas:
### Description

I'm creating this issue to hold discussions about the problem with the publish/subscribe perspective. As it is right now, the perspective is from the "client". Let's put it a different way:

* A1 is the owner of one application.
* A2 is the owner of another application.
* If on the A1 AsyncAPI file it says `publish`, it means the A1 application allows or expects that A2 (and others) can publish messages to A1.
* If on the A1 AsyncAPI file it says `subscribe`, it means the A1 application allows or expects that A2 (and others) can subscribe to receive messages from A1.
* Same applies if you see an A2 AsyncAPI file.

Now, this is highly confusing especially for those working on private broker-based architectures. It's not for those working with public asynchronous APIs.

**This is not an opinion. This is how it works right now.** Switching the meaning of the verbs would not work either, as we saw in the past. Happy to listen for alternatives :)

### Related issue(s)

Proposal from @damaru-inc (Solace) on how to solve it: #390

jonaslagoni:
To me it is not so much about `private broker-based architectures` (since you might not have a broker) but rather those who use AsyncAPI to define the behavior of the specific application. i.e. you want to define what resources A1 provides to others not what others may consume from A1.

GeraldLoeffler:
(Paraphrasing from https://github.com/asyncapi/spec/issues/390#issuecomment-800422743:)
The AsyncAPI document for A1 should use
- `applicationSent` when A1 sends messages to a channel (to which A2 might be subscribed)
- `applicationReceived` when A1 receives messages from a channel (to which A2 might publish).
This means that the AsyncAPI document for A2 (which is equally valid!) should use
- `applicationReceived` when A2 receives messages from a channel (such as those that were sent by A1)
- `applicationSent` when A2 sends messages to a channel (such as those to be received by A1)

GeraldLoeffler:
i've submitted PR #521 that adds wording to the current spec simply to clarify the current meaning of `publish` and `subscribe`.

nictownsend:
Continuing with thread in #390 - I do like the idea of splitting channels away from operations.

The ability to define what a channel looks like, then applications state the operations they perform on a channel.

However, it means either applications state the payload as part of their operation (which could lead to many discrepancies across spec files) - or the channel states its single payload/`allOf` (but different applications/operations can't have differing payloads). Which I'm guessing for NATS (request/reply) becomes a bigger problem.

[...]

PR #687, opened 2022-01-14
docs: Switch description of Subscribe and Publish ✏️ 

aaronkorver:
I may be interpreting this incorrectly, but when I see SUBSCRIBE, the application is going to subscribe to the channel and consume messages.  When I see PUBLISH, the application is going to produce a message and send it to a channel.  I think the definitions were switched in the docs.  This PR is to swap the definitions.

---
title: "Switch description of Subscribe and Publish"
---

---

**Related issue(s):**

---


dalelane:
hi @aaronkorver !

I think it'd be worth having a look at https://www.asyncapi.com/blog/publish-subscribe-semantics - it might help make the intention of subscribe/publish a bit clearer.

aaronkorver:
https://github.com/asyncapi/spec/issues/520 this also talks about it.

The description in the Summary of the blog is actually the best part perhaps.

"publish means publish an event to the channel and this application will receive it"
"subscribe means subscribe to this channel to receive events published by this application"

Let me see if I can instead of switching them (which I now understand is wrong) if instead I can clarify it better in the spec description

Issue #607, opened 2021-08-04
ChannelItem.$ref field deprecation + proposal how to replace it with simpler and more flexible solution


char0n:
Hi all,

I was wondering if specification would be open to proposal of deprecating `ChannelItem.$ref` field and instead introduce a more flexible solution based on following spec changes:

### Introduce new fixed field called `channelItems` to `Component Object` 

Field Name | Type | Description
---|:---:|---
channelItems | Map[`string`, `Channel Item Object` \| `Reference Object`] | An object to hold reusable Channel Item Objects.

### Change type of `Channels Object`

The type of `Channels Object` patterned field would change from `Channel Item Object` to `Channel Item Object | Reference Object`

### Deprecating the `Channel Item Object` `$ref` field.

---

This proposal makes specification more flexible (introducing reusable `Channel Item Objects`) and simpler as there will be only one mechanism for referencing other objects (Reference Object) instead of two (Reference Object, Channel Item Object).

Similar course of action is already in progress in OAS: https://github.com/OAI/OpenAPI-Specification/issues/2635

fmvilas:
That's a good catch 👍 I'll put this on hold while we see how to solve #520 because most probably the Channels object will be reworked too. I'm listing this issue there so we don't forget. Thanks!

derberg:
@char0n if you want to push this one further into the spec in 3.0, next week we start meetings related to 3.0 where all folks can sync the async work 😄 

Just check this calendar https://calendar.google.com/calendar/u/0/embed?src=tbrbfq4de5bcngt8okvev4lstk@group.calendar.google.com or join https://groups.google.com/u/1/g/asyncapi-users to get invites

char0n:
@derberg sure, I'd like to be the champion of this RFC. 

![image](https://user-images.githubusercontent.com/193286/149735231-7ac8db32-6caf-4e74-a7bf-89cc811d5be9.png)

![image](https://user-images.githubusercontent.com/193286/149735552-ab370ae6-3838-4b1b-bc79-90c55eff25ac.png)

I was a little bit confused by the calendar, looks like the meeting starts at 9:00 CET Tuesday but ends at 10:00 next day.
I assume this is mistake and the meeting will start at 9:00 CET and will last hour or two, right?

char0n:
@derberg I have an overlapping meeting in my work today. Will there be another chance to challenge this issue for 3.0 release?

derberg:
[...]

PR #696, opened 2022-01-21
docs(spec): add RFC for safe Channel Item referencing

char0n:
Refs #607


spec/asyncapi.md
```diff
@@ -565,7 +565,7 @@ Describes the operations available on a single channel.
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="channelItemObjectRef"></a>$ref | `string` | Allows for an external definition of this channel item. The referenced structure MUST be in the format of a [Channel Item Object](#channelItemObject). If there are conflicts between the referenced definition and this Channel Item's definition, the behavior is *undefined*.
+<a name="channelItemObjectRef"></a>$ref | `string` | Allows for an external definition of this channel item. The referenced structure MUST be in the format of a [Channel Item Object](#channelItemObject). If there are conflicts between the referenced definition and this Channel Item's definition, the behavior is *undefined*. <br><br>**Deprecated:** Usage of the `$ref` property has been deprecated.
```
smoya, line 568: I think we should give a better explanation on the reason for deprecating. Like mentioning we deprecate it in favor of using `Reference Object` from the `Channels Object` instead of `Channel Item Object`. WDYT?

smoya:
@char0n I think this should be a `fix:` rather than `docs:` since we are fixing the spec and semantic-release would need to know.

spec/asyncapi.md
char0n: If we say we deprecate in favor of `Reference Object` it may indicate that future versions will use `Reference Object`, but later versions of the spec might handle this quite differently. Best way to probably handle this is to have wording self-contained withing the version of the spec this RFC targets - and the sentence I put there was the only one I could come up with. I'm really open to suggestions here ;]

char0n:
@smoya I've force-pushed to change the commit msg type

smoya:
> @smoya I've force-pushed to change the commit msg type

FYI, for next time you just only need to change the PR title instead of the commit, because the squashed merged commit uses the PR title as commit message.

fmvilas (approved):

spec/asyncapi.md
smoya: I'm fine if others see it make sense 👍

smoya:
@char0n Additionally, would you like to provide some words regarding this change for the `v2.3.0` release notes @dalelane is handling in https://github.com/asyncapi/website/pull/512? 🙏

char0n:
@smoya yes would be a great medium where to describe the deprecation plan of the `$ref` field. Added a PR comment there with suggested addition: https://github.com/asyncapi/website/pull/512/files#r789842794

dalelane:
@char0n @smoya Should we also update the JSON schema at https://github.com/asyncapi/spec-json-schemas/pull/142/files#diff-c42eed206b44fe4fa90f1635ff156f9804ae2450b901771b9c4d6db5188d640cR498-R500 to mark `$ref` as deprecated? 

e.g. 
```json
    "channelItem": {
       ...
       "properties": {
         "$ref": {
           "$ref": "#/definitions/ReferenceObject",
           "deprecated": true
         },
```

dalelane:
@smoya Do you think anything else is needed to be able to approve and merge this?

smoya (approved):
LGTM (not an owner anyway)

derberg:
this one was pretty interesting and education read 😄 

@char0n please have a look at the change I did in PR title, I think most important is to highlight deprecation in the release notes

also can you please extend description of this issue so reader of release notes, that will end up in this PR doesn't have to go through a long discussion from the issue but have a summary here in PR description?

char0n:
> @char0n please have a look at the change I did in PR title, I think most important is to highlight deprecation in the release notes

Yeah, change in PR title looks good. @dalelane did a great job in verbalizing the proper deprecation message in https://github.com/asyncapi/website/pull/512/files#diff-a43f6e629535a43413950aefe222b7d7e591e50515314573ff440e91b8aaed8aR66

> also can you please extend description of this issue so reader of release notes, that will end up in this PR doesn't have to go through a long discussion from the issue but have a summary here in PR description?

I'll do my best ;]

derberg (approved):

Issue #705, opened 2022-02-04
Unable to reuse `Server Objects` defined in `Components.servers` field


char0n:
Hello everybody,

While implementing 2.3 support into tooling I've noticed couple of inconsistencies/errors in the specification.
As AsyncAPI 2.3.0 introduced new `Components.servers` field, there is no way how `Server Objects` defined in this map can be used later in the definition.

### Usage in `Servers Object`

I assume first intended usecase was to reference `Server Object` defined in `Components.servers` field in `Servers Object`. This intention is clear from the release article:

![image](https://user-images.githubusercontent.com/193286/152542621-9673fc78-e6ca-4078-8a79-736fdb672ae2.png)

What the article is showing is not possible according to the spec as the `Servers Object` pattern field only allows `Server Object` and not `Server Object | Reference Object`.

### Usage in `Channel Item Object`.`servers` field

With this I'm not 100% sure of the intended usage. The specification says following:

![image](https://user-images.githubusercontent.com/193286/152543308-738af474-25da-48ff-aad3-a4ddab295cda.png)

This limits using referencing `Server Object` to ones defined `Servers Object` (by name). Should we allow referencing the `Server Object` even from `Components.servers` field as well? 

I'll provide a PR in a couple of minutes.

jonaslagoni:
Good catch! 

>This limits using referencing Server Object to ones defined Servers Object (by name). Should we allow referencing the Server Object even from Components.servers field as well?

That would be a new feature, as regardless of what you define in your components sections, servers are not "active" until referenced under `servers`. They are nothing but reusable chunks. So `string` still makes sense to use for `Channel Item Object.servers`. 🙂

char0n:
> That would be a new feature, as regardless of what you define in your components sections, servers are not "active" until referenced under `servers`. They are nothing but reusable chunks. So `string` still makes sense to use for `Channel Item Object.servers`. slightly_smiling_face

Right, I assumed it was like you explained but I'd rather asked ;] Having said that, maybe it warrants clarification in the spec? But maybe not...not sure.

jonaslagoni:
> Right, I assumed it was like you explained but I'd rather asked ;] Having said that, maybe it warrants clarification in the spec? But maybe not...not sure.

Could definitely be now that you can define servers in multiple places 🤔 Maybe another issue for further discussing that?

char0n:
Yeah I think it's worth discussing further. If agreed, I'll extract part of this issue regarding servers to a new one.

[...]

Issue #688, opened 2022-01-18
What do we define as a breaking change?


jonaslagoni:
This question emerged when @smoya wanted to make the channels field optional - https://github.com/asyncapi/spec/issues/661

This change, might not be a breaking change to the spec itself, however, for tooling that relies on the channels being required and always present, it would.

So when should a change in the spec be considered a breaking change?

I see two ways we can do this (and that is assuming we want to stay with semantic releases).

## Only focusing on the spec
Agreeing that as long as the spec is backward compatible, it is not a breaking change.

## Agreeing what is breaking changes
Agreeing on special cases would probably be possible to minimize breaking changes in tooling and spec.

My immediate thoughts are the following that would be questionable:

_Non-breaking changes:_
- Adding a property

_Breaking changes:_
- Making a property required
- Making a property optional
- Removing a property
- Changing the type of a property in any way

Can you think of another possible solution?

smoya:
You made a really good point here. I agree having in mind breaking changes also at tooling level is important.
I like the idea of treating breaking changes in different ways, depending on the context: A non-breaking change for the spec doesn't mean is not a breaking change for tooling and vice versa. 

If we go in that direction, we should set obvious boundaries and consider **only** the tools under AsyncAPI umbrella (Code hosted on github.com/asyncapi) as implementations to check for breaking changes (unless the breaking change is clearly detected earlier).

Also, it would be nice if we build some quick tool we can run on CI to check for usages and, at least, list them on the PRs, like possible affected code. Github search is [not very powerful](https://docs.github.com/en/search-github/searching-on-github/searching-code) for it.

fmvilas:
> Making a property required

@jonaslagoni making a property required is also a breaking change. If I use an AsyncAPI 2.2.0 document as the input for an AsyncAPI 2.3.0 tool, and we have made a property required in v2.3.0, it will be a breaking change. I mean, changing `asyncapi: 2.2.0` to `asyncapi: 2.3.0` would not be enough.

Regarding #682, if we're not sure about it, we can always postpone it to v3 now that we're working on it.

jonaslagoni:
> @jonaslagoni making a property required is also a breaking change. If I use an AsyncAPI 2.2.0 document as the input for an AsyncAPI 2.3.0 tool, and we have made a property required in v2.3.0, it will be a breaking change. I mean, changing `asyncapi: 2.2.0` to `asyncapi: 2.3.0` would not be enough.

You right, adapted suggestion!

[...]

PR #706, opened 2022-02-04
fix(Servers): allow to use union type of Server and Reference objects

char0n:
Refs #705

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -319,7 +319,7 @@ The Servers Object is a map of [Server Objects](#serverObject).
 
 Field Pattern | Type | Description
 ---|:---:|---
-<a name="serversObjectServer"></a>`^[A-Za-z0-9_\-]+$` | [Server Object](#serverObject) | The definition of a server this application MAY connect to.
+<a name="serversObjectServer"></a>`^[A-Za-z0-9_\-]+$` | [Server Object](#serverObject) \| [Reference Object](#referenceObject) | The definition of a server this application MAY connect to.
 
 ##### Servers Object Example
 


smoya:
> If it is a new feature, we could include it in 2.4 release (expected to be released in April 2022).

The reality is that our parser supports this from the very beginning, so we could consider this change just as it is, a fix.
However, It could be considered even as a breaking change, since another tooling may expect only a Server Object but now it can be both Server Object or Reference Object.
This is another example for https://github.com/asyncapi/spec/issues/688.

I'm up for adding this fix into v2.4.0 release, what do others think @dalelane @derberg @fmvilas ?

derberg:
I see it as a fix, but fix to the specification should be released as a patch. Since 2.4 is pretty close, best to put it into 2.4 branch than now trigger 2.3.1 as this is not critical. How doesn't it look like in spec's JSON Schema, all good there?

smoya:
> I see it as a fix, but fix to the specification should be released as a patch. Since 2.4 is pretty close, best to put it into 2.4 branch than now trigger 2.3.1 as this is not critical. How doesn't it look like in spec's JSON Schema, all good there?

We are covered as `Server` object has the property `$ref` that is a `ReferenceObject`. See https://github.com/asyncapi/spec-json-schemas/blob/master/schemas/2.3.0.json#L173-L175.

Is that the best approach? It is consistent with the rest of the objects? This is a topic @jonaslagoni brought in https://github.com/asyncapi/spec-json-schemas/issues/167., but out of the scope of this PR.

smoya:
@char0n would you mind changing the target base branch to `2022-04-release` so we can include this fix in the 2.4 release happening next month? Thanks!

smoya:
> @char0n would you mind changing the target base branch to `2022-04-release` so we can include this fix in the 2.4 release happening next month? Thanks!

Maybe any code owner can do this? cc @fmvilas @derberg @dalelane 🙏

char0n:
@smoya sorry for delay - was a little sick. Doing it now.

char0n:
base changed

smoya (approved):
LGTM! 🚀 🌔 

However, you will need approval from at least one code owner. cc @dalelane @derberg @fmvilas

dalelane (approved):
👍

PR #710, opened 2022-02-07
chore: add 2.3.0 version to the list in readme

char0n:
I've noticed that the version list in main README file is missing 2.3.0 version of the spec. This PR provides remediation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,8 @@
 
 The latest draft specification can be found at [spec/asyncapi.md](./spec/asyncapi.md) which tracks the latest commit to the master branch in this repository.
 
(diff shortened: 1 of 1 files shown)


jonaslagoni:
@derberg @fmvilas I saw a few PR's now that missed changes needed to be done for a new version. Are we missing anything in the [release process](https://github.com/asyncapi/spec/blob/master/RELEASE_PROCESS.md) so this don't turn up going forward?

dalelane:
@jonaslagoni yeah, I'm getting a PR ready with updates to the process to cover the things that we caught post-release (as well as the things we noticed earlier too)

fmvilas (approved):

derberg:
/rtm

I changed to `chore` as it is not spec related docs change and no need to trigger update on website

Issue #612, opened 2021-08-24
Channel Item Object `$ref` field resolution rules


char0n:
Description of the `$ref` field explicitly says: 

> Allows for an external definition of this channel item. The referenced structure MUST be in the format of a Channel Item Object. If there are conflicts between the referenced definition and this Channel Item's definition, the behavior is undefined.

As a tooling author I have couple of questions here:

### Allowing only external definition

Why are we allowing only `external definition` and not an internal as well? By changing wording to `Allows for a referenced definition of this channel item.` we open a possibility to reference other Channel Items defined in the same document. This constraint (`external definition`) can be easily bypassed by defining a `$ref` field as URL+JSON Pointer fragment, where URL has the same baseURL as the root document.

### Rules of resolving

Rules of resolving are missing entirely. It is not clear how to do the actual resolution. I can only assume what the rules might be by reading one of the paragraph in `Reference Object` section:

> For this specification, reference resolution is done as defined by the JSON Reference specification and not by the JSON Schema specification.

So if resolution for `Reference Object` is done using JSON Reference spec, I'd assume the same for Channel Item Object. This assumption can be wrong thought. In order to remove un-clarity I propose to mention following sentence in `Channel Item Object`.`$ref` field description: `Resolution is done as defined by the JSON Reference specification.`

I can open PR with proposed changes if/when we agree on them.

bali182:
This is a very good question, I was wondering about the $ref field too. OpenAPI has the same issue with Operation Objects. Could we either 

- Remove the $ref field from Channel Item Object, this way forcing all Channel Item Objects to be defined in the same document
- Make it union type (either a Channel Item Object or a Reference Object) and make it behave like all other referenceable pieces of the schema?

I think either would be

char0n:
> This is a very good question, I was wondering about the $ref field too. OpenAPI has the same issue with Operation Objects. Could we either

I guess you mean Path Item Object in OpenAPI. I am already driving effort to eliminate it in v3.2: https://github.com/OAI/OpenAPI-Specification/issues/2635
> 
> * Remove the $ref field from Channel Item Object, this way forcing all Channel Item Objects to be defined in the same document
> * Make it union type (either a Channel Item Object or a Reference Object) and make it behave like all other referenceable pieces of the schema?
> 
> I think either would be better than the mixture of both

For this I’ve already submitted a proposal to remove $ref field and replace it with Reference Object: https://github.com/asyncapi/spec/issues/607

[...]

PR #714, opened 2022-02-07
docs(spec): indicate in detail how Channel Item Object  field resolves

char0n:
Refs #612 

This PR targets targets two things:

1. As we now have reusable `Channel Item Object`s stored in `Components.channels`, there is no reason why the description of the `$ref` field should explicitly only allow referencing external definition of the `Channel Item Object` (in different document) and not the internal one

2. Second change targets the unspecified rules of resolution for the `$ref` field. I've added explicit mention of JSON Reference specification which the resolution of the `$ref` field is based on.


char0n:
Succeeded by https://github.com/asyncapi/spec/pull/779

PR #718, opened 2022-02-09
docs: fix URN examples

mike-schenk:
---
title: "Use valid URIs or URNs in Identifier examples"
---

Changes invalid URN examples to either valid URIs or to valid URNs according to the linked RFC 8141 and its associated IANA URN namespaces registry.

Reverse DNS names do not form valid URNs. The registry of defined [URN namespace identifiers](https://www.iana.org/assignments/urn-namespaces/urn-namespaces.xhtml) is quite limited. Because of that, it seems unlikely that a valid URN will actually be useful in the context of an AsyncAPI document, but to make the minimal change in the spec document, I used the `example` namespace identifier.

Instead of a URN, a URI with the registered [`tag` scheme](https://www.rfc-editor.org/rfc/rfc4151.html) seems to apply better. Since the semantics of the tag URI scheme appear to match the desired intent of a universally-unique and long-lived identifier.

This is an editorial change, but I apparently can't set that label myself.

---

**Related issue(s):**

---


derberg:
@mike-schenk hi, I'm really sorry for getting so late here. Can you solve merge conflicts please 🙏🏼 

@fmvilas @dalelane anything from your side?

fmvilas (approved):
Looks good to me 👍

dalelane (approved):
looks good

@mike-schenk - would you like some help with the merge conflicts?

mike-schenk:
Merge conflicts resolved.

dalelane (approved):
👍

fmvilas:

derberg:

Issue #663, opened 2021-12-01
Let channels be identified by an ID rather than their address.


smoya:
## Context

This is part of the RFC https://github.com/asyncapi/spec/issues/618#issuecomment-980093487. 
In an effort around reducing the scope of such RFC, it has been split into several proposals so they can move forward independently.

## The problem

Channels are currently identified by their address. E.g.:

```yaml
channels:
  user/signedup: # this is both the address and the identifier of the channel
    description: ...
```

With the current approach, and once https://github.com/asyncapi/spec/issues/660 happens, we will have few issues:

- When referencing a channel, the channel identifier should match with the refereced one, otherwise the address will be different than in the referenced channel. E.g.:
  ```yaml
  # common.yaml
  channels:
    user/signedup:
        description: ...
        
  ```

  ```yaml
  # asyncapi.yaml
  channels:
    user/signedup: # here we have to use the same identifier as in common.yaml, otherwise the address will be different
      $ref: 'common.yaml#/components/channels/user~signedup'
  ```
- That introduces a problem when we need to change the address of a shared channel. Let's say Kafka topic name changes. With the current approach, we need to change the address on all files that reference the `user/signedup` channel from `common.yaml` (as JSON Reference will fail as well).
- As you may noticed, the JSON Pointer for `user/signedup` channel is `user~1signedup`. It is unexpected (at least to me) and you have to have some JSON Pointer knowledge or find examples to find out such format. It is very error-prone.

## Suggestion

To add a new `address` field to the channel definition that represents the real address and not the channel identifier.
With that change, the previous example will look like:

  ```yaml
  # common.yaml
  channels:
    userSignedUp:
        address: user/signedup
        description: ...
        
  ```

  ```yaml
  # asyncapi.yaml
  channels:
    myUserSignedUp: # here we can add whatever ID we want. If address changes in `userSignedUp` from common.yaml, nothing will need to change in this file
      $ref: 'common.yaml#/components/channels/userSignedUp' # no slash transformation is needed in the JSON Pointer :)
  ```

jonaslagoni:
I love the move to a property, whether `address` is the right keyword, not so sure, but I don't have other alternatives at the moment.

What exactly is the reason we need an ID rather than making channels an array?

Especially now that https://github.com/asyncapi/spec/pull/665 is added.

smoya:
> I love the move to a property, whether `address` is the right keyword, not so sure, but I don't have other alternatives at the moment.
> 
[...]

PR #719, opened 2022-02-09
feat: let channels be identified by ID. Address is now moved into a new field

smoya:
This PR allows channels to be identified by an ID rather than by their address. Instead, the address has now been moved into a new field called `address`.

This is a breaking change so it should be considered as a candidate for `v3.0.0`.

**Related issue(s):**

See https://github.com/asyncapi/spec/issues/663 for knowing about the reason behind this RFC.

CC @char0n @magicmatatjahu @derberg

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -20,13 +20,14 @@ These files can then be used to create utilities, such as documentation, integra
 The file(s) MUST describe the operations an [application](#definitionsApplication) accepts. For instance, consider the following AsyncAPI definition snippet:
 
 ```yaml
-user/signedup:
+userSignedUp:
+  address: "user/signedup"
   subscribe:
     message:
       $ref: "#/components/messages/userSignUp"
 ```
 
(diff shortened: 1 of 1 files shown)


spec/asyncapi.md
```diff
@@ -20,13 +20,14 @@ These files can then be used to create utilities, such as documentation, integra
 The file(s) MUST describe the operations an [application](#definitionsApplication) accepts. For instance, consider the following AsyncAPI definition snippet:
 
 ```yaml
-user/signedup:
+userSignedUp:
+  address: "user/signedup"
```
smoya, line 24: Perhaps we want to keep calling this `path` instead of `address`? I would love to know others opinion. I'm fine with `address` anyway.

smoya:
@fmvilas @derberg @dalelane would you mind removing `stale` label? thanks

smoya:
Closing in favour of https://github.com/asyncapi/spec/pull/827

Issue #654, opened 2021-11-12
Allow grouping servers


smoya:
## Context

Current [Server Object](https://www.asyncapi.com/docs/specifications/v2.2.0#serversObject) is identified by the key on `servers` map. E.g.:

```yaml
servers:
  production: # 'production' is the name of the server.
    url: localhost:8080
```

In the previous example, the word `production` identifies the server. We could also call it by it's server machine name or service name, e.g. `yavin04.yavin.systems`.
No matters what name do we use for the server as far as are unique in the `servers` map.

One common pattern (See https://github.com/asyncapi/spec/issues/628) is to declare reusable servers in common AsyncAPI files, then reference them in the specific AsyncAPI files. E.g.

```yaml
# common.asyncapi.yaml
components:
  servers:
    yavin04.yavin.systems:
      url: https://yavin04.yavin.systems:8080
      protocol: http
```

```yaml
# app.asyncapi.yaml
servers:
  production:
    $ref: 'common.asyncapi.yaml#/components/servers/yavin04.yavin.systems'
```

What if a server exposes multiple ports with different protocols? Current Spec forces to create a different [Server Object](https://www.asyncapi.com/docs/specifications/v2.2.0#serversObject) for this matter, since only one protocol is allowed per server. E.g.:

```yaml
# common.asyncapi.yaml
components:
  servers:
    yavin04.yavin.systems-http:
      url: https://yavin04.yavin.systems:8080
      protocol: http
    yavin04.yavin.systems-websocket:
      url: https://yavin04.yavin.systems:5000
      protocol: ws
```

```yaml
# app.asyncapi.yaml
servers:
  productionHTTP:
    $ref: 'common.asyncapi.yaml#/components/servers/yavin04.yavin.systems-http'
  productionWebsocket:
    $ref: 'common.asyncapi.yaml#/components/servers/yavin04.yavin.systems-websocket'
```

> Just as side note, I wouldn't say no to evaluating the idea of allowing several protocols to be declared (linked to ports). But this is for another issue :)

## The problem

How can a user answer the question: "How can I identify the list of real/physical servers or service names in my system?", where `system` refers to the infrastructure around this or all my applications.

Also from the point of view of code generators, the same answer could be formulated. For example. "Generate all servers from service name `yavin04.yavin.systems`". This will end up generating one for `http` protocol, and another one for `ws`.

[...]

PR #744, opened 2022-03-28
feat: add 'values' field to Tag Object

smoya:
Part of https://github.com/asyncapi/spec/issues/654

This PR adds a new `values` field to  `Tag Object` to declare a list of plain strings as values.

**Related issue(s):**

https://github.com/asyncapi/spec/issues/654

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -1351,6 +1351,7 @@ Field Name | Type | Description
 <a name="tagObjectName"></a>name | `string` | **Required.** The name of the tag.
 <a name="tagObjectDescription"></a>description | `string` | A short description for the tag. [CommonMark syntax](https://spec.commonmark.org/) can be used for rich text representation.
 <a name="tagObjectExternalDocs"></a>externalDocs | [External Documentation Object](#externalDocumentationObject) | Additional external documentation for this tag.
+<a name="tagObjectValues"></a>values | [`string`] | The value(s) for the tag.
 
 This object can be extended with [Specification Extensions](#specificationExtensions).
 


smoya:
Examples added.

spec/asyncapi.md
```diff
@@ -1351,15 +1351,18 @@ Field Name | Type | Description
...
 
-##### Tag Object Example
+##### Tag Object Examples
+
+###### Simple example
 
 ```json
 {
-	"name": "user",
-	"description": "User-related messages"
```
smoya, line 1362: For some reason, there were tabs here.

spec/asyncapi.md
```diff
@@ -1351,15 +1351,18 @@ Field Name | Type | Description
 <a name="tagObjectName"></a>name | `string` | **Required.** The name of the tag.
 <a name="tagObjectDescription"></a>description | `string` | A short description for the tag. [CommonMark syntax](https://spec.commonmark.org/) can be used for rich text representation.
 <a name="tagObjectExternalDocs"></a>externalDocs | [External Documentation Object](#externalDocumentationObject) | Additional external documentation for this tag.
+<a name="tagObjectValues"></a>values | [`string`] | The value(s) for the tag.
```
fmvilas, line 1354: 🤔 I'm not sure I like having `values` inside the global Tag object. We only verified this has a use case on servers but how is it going to affect the rest of the spec where Tag is used? Should we have another kind of Tag object just for servers?

magicmatatjahu: For other objects it will work the same way as for servers, just the tags will be treated as "metadata", with which you will be able to filter certain elements in the UI or via the CLI (of course you will have to implement this). Currently a tag only has a key (name field) but you can't define certain more specific values for that (we don't want use delimeter or something like). This can be compared to `labels` in Kubernetes where labels are used to filter/create relationships between resources - https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/

dalelane:
I'm not sure what I think about this. I think my hesitation mainly comes from the fact that [OpenAPI doesn't do this](https://spec.openapis.org/oas/v3.1.0#tag-object). 

I know we're not beholden to what OAI do, and consistency for the sake of consistency isn't smart. But... tagging isn't a uniquely-event-centric thing, so I'm hesitant about diverging on things where it isn't specifically in service of asynchronous issues. 

Like I say I'm not sure. If OAI didn't exist, purely looking at this idea in isolation, I'd probably be more in favour.

smoya:
> I'm not sure what I think about this. I think my hesitation mainly comes from the fact that [OpenAPI doesn't do this](https://spec.openapis.org/oas/v3.1.0#tag-object).
> 
> I know we're not beholden to what OAI do, and consistency for the sake of consistency isn't smart. But... tagging isn't a uniquely-event-centric thing, so I'm hesitant about diverging on things where it isn't specifically in service of asynchronous issues.
> 
> Like I say I'm not sure. If OAI didn't exist, purely looking at this idea in isolation, I'd probably be more in favour.



> I'm not sure what I think about this. I think my hesitation mainly comes from the fact that [OpenAPI doesn't do this](https://spec.openapis.org/oas/v3.1.0#tag-object).
> 
> I know we're not beholden to what OAI do, and consistency for the sake of consistency isn't smart. But... tagging isn't a uniquely-event-centric thing, so I'm hesitant about diverging on things where it isn't specifically in service of asynchronous issues.
> 
> Like I say I'm not sure. If OAI didn't exist, purely looking at this idea in isolation, I'd probably be more in favour.

I understand your point. In fact, I didn't pay attention to how OAI does (now I realize we just ported it literally).
I like to keep things simple, and I agree this can introduce an extra layer of complexity. However, I see it as a nice UX improvement over what we have, especially for those composable tags. 
Considering the examples in the Issue description and more in particular, the one @fmvilas showed in https://github.com/asyncapi/spec/issues/654#issuecomment-1074341626, I see composable tags are a needed thing.
Otherwise, as exposed in https://github.com/asyncapi/spec/issues/654#issuecomment-1076293457, some filters might be hard to implement on tooling (or too tricky).

As @magicmatatjahu mentioned, this pattern of key/value for tags is common in other solutions such as Kubernetes, where you can implement selectors that solve most of the issues of grouping and filtering objects (workloads in their case).

I see it as a UX improvement that allows more extensibility, even though not mandatory for any new feature yet afaik.

fmvilas:
Yeah, that's a great point @dalelane. @smoya what about introducing the concept of `labels` and making them exactly the same as k8s labels? We can then recommend that labels are meant to be used when you want to filter data (like k8s do with selectors). And tags are still relevant when you simply want to "categorize" stuff. Not sure it makes sense, what do y'all think?

smoya:
> Yeah, that's a great point @dalelane. @smoya what about introducing the concept of `labels` and making them exactly the same as k8s labels? We can then recommend that labels are meant to be used when you want to filter data (like k8s do with selectors). And tags are still relevant when you simply want to "categorize" stuff. Not sure it makes sense, what do y'all think?

I wanted to make the same suggestion yesterday, however, those are the points that stopped me to do it:

1. The term `label` and `tag` seems mostly the same.
2. In consequence, It introduces ambiguity. When should I use a tag? When a label instead? What's the difference?
3. Tooling will have to support filters or selectors for both types. I.e. in generators, the command for generating code for **production** servers would contain a selector/filter like `env:production`. Would the command expect the tag or the label? or both? we could fallback to the first thing we find (ie. label with key `env` and value `production` or fallback to tag with name `env:production`. 

For the 1st and 2nd points, I would love to ask @dalelane about the terminology `label` vs `tag` (english is not my mother tongue).
Regarding the 3rd, makes me ask myself if we really need a new object (Label) or instead, use the one we have now, with or without the upgrade to have values.




However, after writing this again here, It doesn't look that bad to me. Especially the third point, which

spec/asyncapi.md
smoya: Btw, I'm not sure if we should call it `values` or rather `value` in singular, even we allow having multiple. It is just a feeling I guess but the plural sounds weird to me in this case.

fmvilas:
> 1. The term label and tag seems mostly the same.
> 2. In consequence, It introduces ambiguity. When should I use a tag? When a label instead? What's the difference?

Yeah, I actually searched for the difference between label and tag when replying above so fair point.

> Are we at the starting point again now?😅

Pretty much yeah 😅 But that's fine 😄 

So, my concern with changing the Tag Object is that people would not be able to reuse their definitions between OpenAPI and AsyncAPI. Actually, if we make the `values` field optional —and I think we should anyways—, the Tag Object would be portable from OpenAPI to AsyncAPI but not the other way around. But "portable" is not the same as "reusable", meaning that if someone has tags extracted into a separate file or registry, they would not be able to reference them from both places **if we add a values field**.

So my proposal is that we ping the OpenAPI folks and see if they would be happy to add the `values` field too. At least as a first step. If they don't want to add it and we don't agree with the reasons, we can always move forward from our side but we should always try to collaborate with the rest of specs as much as possible IMHO.

derberg:
added comment here https://github.com/asyncapi/spec/issues/654#issuecomment-1088414968

PR #749, opened 2022-04-11
feat: add operations to the components object

magicmatatjahu:
---
title: "feat: add operations to the components object"
---

---

This PR add ability to define reusable OperationObject inside ComponentsObject.

---


dalelane (approved):

fmvilas:
What's the use case driving this? 🤔 I see no issue discussing this feature.

magicmatatjahu:
There is no discussion in any issue (only in slack discussion - https://asyncapi.slack.com/archives/C0230UAM6R3/p1649669700248579), because I was wondering why we have channels/servers/messages in components and no operations. Simply use case is a possibility to use reusable operations between several AsyncAPI documents - you define in one file where you define only components (you treat one document as registry) and then you make references where you want.

fmvilas:
> Simply use case is a possibility to use reusable operations between several AsyncAPI documents - you define in one file where you define only components (you treat one document as registry) and then you make references where you want.

Yeah, understand the **potential** use case but I'm more curious if this is actually that has been needed at some point by someone (even if just you while doing an experiment or something). One of the guiding principles of the spec is [Favor no change](https://github.com/asyncapi/spec/blob/master/CONTRIBUTING.md#guiding-principles). I know we're not always respecting it but I think we should be more cautious by adding stuff that may not be necessary ever.

magicmatatjahu:
In version 3.0 we will have to add this. Is it better sooner or later?

fmvilas:
You're not answering my question :)

And if we're adding it in v3.0, why bother adding it to v2.x? It's never been there and it's been fine. It can wait until v3 perfectly.

magicmatatjahu:
@fmvilas You didn't ask a question in the previous comment, and on the previous question you nodded to my answer, although you didn't fully agree 😄  - and no, there was no discussion and I have no real, production use case.

When we were adding channels and servers to the components in 2.3 it was also "not" a real use case (of course people complain about the lack of reusability of channels, but due to the fact that operations are strictly defined with channels the reusability of channels in >=2.3 still causes a big problem). We agreed then, that it is better to add things that we plan to add in 3.0 to 2.x, those that do not cause breaking changes. In my opinion, reusing operations themselves makes more sense than channels or servers, because they can be reused in several channels, and channels and servers can be used only once.

fmvilas:
> We agreed then, that it is better to add things that we plan to add in 3.0 to 2.x, those that do not cause breaking changes.

It's going to be a breaking change anyway because the operation object shape will change.

> In my opinion, reusing operations themselves makes more sense than channels or servers, because they can be reused in several channels, and channels and servers can be used only once.

It does. And it introduces more complexity in the current spec major version when nobody asked for that. That's my point. We did not ask when introducing channels or servers and that's probably our fault. We should have questioned those features too given that v3 is in the works. In other words, I'd just continue evolving v2 if it's strictly necessary and useful for a real use case. I mean, it should always be like that but now more than ever.

magicmatatjahu:
Then I close the PR, if someone wants it in `2.5` I will reopen it.

derberg:
```
You didn't ask a question in the previous comment, and on the previous question you nodded to my answer, although you didn't fully agree 😄 
```

say what? 🤣

PR #769, opened 2022-04-25
docs: fix links in Markdown files

KhudaDad414:
### Context
this PR is supposed to fix all the [broken links](https://github.com/asyncapi/spec/runs/6143352833?check_suite_focus=true) in markdown files across the repo.


mlc_config.json
```diff
@@ -0,0 +1,16 @@
+{
+  "ignorePatterns": [
```
smoya, line 2: Would you mind adding `^https://github.com/asyncapi/spec/blob/.*.md$` (or similar)?
Just to avoid errors on links that are going to exist only once the particular PR is merged, like during releases of the spec. I.e. https://github.com/asyncapi/spec/runs/6190017382?check_suite_focus=true

smoya: MM I think its better to don't add it. As the action is not mandatory for merging prs, and considering this is just an exception, I would just not add this into the ignorePatterns.

KhudaDad414: yeah, it makes more sense.

@derberg I guess because of cases like this, the decision of making the "link checking workflow" `required` should be taken per repo basis or we should not make it `required` at all?

derberg: to have it useful it should be required everywhere, we just did not make it required as we know there is a bazillion of links broken and we need few weeks until we get into the stage that "we know we will not cause unexpected CI fails on PR, so we won't annoy contributors" - like we will not cause the problem that developers do not like, like fix thinks not related to PR, not caused by them

KhudaDad414: @smoya  Then, I guess adding it to the ignorePatterns is the only way forward since this link will cause the weekly check and maybe the other PRs to fail after we make it mandatory.

smoya: I'm happy to make it required, but then yes, this should be added into the ignorePatterns, otherwise PR's for new releases might fail as links could not exist yet.

fmvilas (approved):
LGTM 👍

derberg (approved):

fmvilas:

derberg:

Issue #775, opened 2022-05-03
Extending `Server.variables` field type


char0n:
As of version **2.4.0**, `Server Variable Object` can now be defined in `Components.serverVariables` fixed field, which indicates re-usability. Here is an example that I've took from [AsyncAPI 2.4.0 release article](https://www.asyncapi.com/blog/release-notes-2.4.0).

![image](https://user-images.githubusercontent.com/193286/166421534-d92ff2fa-08f6-461a-a6a6-485ac888ca0f.png)

Unfortunately according to the **2.4.0** version of the specification, `Server.variables` field doesn't allow to use `Reference Object`:

![image](https://user-images.githubusercontent.com/193286/166421745-18766566-e78d-4494-8c73-00f2b02523ba.png)

In order to remedy this issue, we have to extend the `Server.variables` type to `Map[string, [Server Variable Object | Reference Object]]`. 

---

Given we can now define re-usable `Server Variable Objects` in `Components.serverVariables` field, we cannot reuse them in any way in **2.4.0** version of the spec.

Ref https://github.com/asyncapi/spec/issues/707

char0n:
I've issued a [PR](https://github.com/asyncapi/spec/pull/776) that remedies the situation and would probably suggest to go for **2.4.1** as soon as possible.

derberg:
Recent comments about the release from the bot were added by mistake. More details in https://github.com/asyncapi/spec/issues/899

PR #776, opened 2022-05-03
feat: allow re-usability of Server Variable Objects

char0n:
Refs #775


fmvilas:
Answering your comment in the #775 issue:

> I've issued a https://github.com/asyncapi/spec/pull/776 that remedies the situation and would probably suggest to go for 2.4.1 as soon as possible.

Yeah, I think it's a pretty good use case for a patch version.

fmvilas (approved):
Looks good to me. Let's wait for the other code owners to review it.

Would also love @smoya and @danielkocot to have a look since they've been involved in the last release.

magicmatatjahu:
Do we really have to do `2.4.1`? In my opinion, it's more of an oversight and that doesn't change the structure of the specification at all. When OpenAPI 3.1.0 came out, people were still making small improvements to the spec without releasing 3.1.1.

char0n:
> Do we really have to do `2.4.1`? In my opinion

No, it's not really necessary, I've just suggested it given the quick cadence of release process for the spec. 

> it's more of an oversight and that doesn't change the structure of the specification at all. 

IMHO it does change the capabilities of the specification; strictly speaking it extends the capabilities, so `feat` would be more appropriate, but given the context and the fact that this is oversight I went with `fix`.

> When OpenAPI 3.1.0 came out, people were still making small improvements to the spec without releasing 3.1.1.

Not sure this is correct observation (but I might be mistaken). OAS is considered immutable and the last commit to the 3.1.0 version is actually the `release` commit. No commits after that.

Commit history for the 3.1.0 spec: https://github.com/OAI/OpenAPI-Specification/commits/main/versions/3.1.0.md

danielkocot:
For me it looks also good. Thanks @char0n for extending the idea of reusability.

smoya (approved):
LGTM!

Even though I don't like the idea of releasing a patch, I agree @char0n it is appropriate in this case. The spec is defined in this repository, and not in the json schema definitions (those are supporting the spec, not defining them IMHO), meaning this should be released as what it is, a fix.
I would not release it as a minor since, even though this introduces a new feature technically speaking, the intention on 2.4.0 was to support this and we missed this change, so It can be considered a bug (ergo this fix).

smoya:
@dalelane @fmvilas your review is needed here. Thanks!

dalelane (approved):

char0n:
Given we've discontinued effort with [2.4.1](https://github.com/asyncapi/spec/issues/821) release and are going directly for [2.5.0](https://github.com/asyncapi/spec/issues/830), this PR needs to be send against next-spec and the title needs to changeto `feat: allow re-usability of Server Variable Objects`

derberg:
@char0n yup, I think you should be able to change the base branch

char0n:
Right, I'll do it right after https://github.com/asyncapi/spec/pull/831 is merged.

char0n:
Base branch changed for `next-spec`.

char0n:
Changed title from `fix: allow re-usability of Server Variable Objects` to `feat: allow re-usability of Server Variable Objects`. Can I ask somebody to to remove `do-not-merge` label or merge this PR? Thanks!

Issue #699, opened 2022-01-25
Remove `$ref` field from `Channel Item Object` in next breaking change version (3.0.0)


char0n:
This is continuation of https://github.com/asyncapi/spec/issues/607. Changes for AsyncAPI 2.3 has already been issued.

**AsyncAPI 2.3**

The aim here is to limit complexity and bring more flexibility.

- [x]  Deprecate the use of `$ref` fixed field in `Channel Item Object`
- [x]  Introduce new fixed field called `channels` in `Components Object` with type of Map[`string`, `Channel Item Object`]

**AsyncAPI 3.0.0**

The aim here is to have single referencing mechanism within AsyncAPI 3.x spec (reducing complexity) - `Reference Object` 

- [x] Remove `$ref` from `Channel Item Object` fixed fields
- [x] Change `Channels Object` patterned field type to `Reference Object | Channel Item Object`
- [x] `Components Object`.`channels` field type will change to Map[`string`, `Channel Item Object` \| `Reference Object`]

char0n:
[PR](https://github.com/asyncapi/spec/pull/777) has been issued for [next-major-spec](https://github.com/asyncapi/spec/tree/next-major-spec) branch.

jonaslagoni:
As I can see we can close this issue as it's completed both for version 2 and 3.

cc @char0n @fmvilas @derberg @dalelane

char0n:
@jonaslagoni sure, fine with me.

derberg:
I don't think we should close as long as 3.0 is not yet merged and released.
We should rather make sure it is linked in description of https://github.com/asyncapi/spec/pull/759 with info that it will be solved after the merge

char0n:
Be patient @github-actions ;]

char0n:
Reopening due to https://github.com/asyncapi/spec/issues/699#issuecomment-1273650238

PR #777, opened 2022-05-04
feat: strip custom referencing mechanism from Channel Item

char0n:
This is attempt to get RFC(https://github.com/asyncapi/spec/issues/699) from **Stage 1** to **Stage 2**. It has already been stipulated in https://github.com/asyncapi/spec/issues/607 that this is reasonable course of action which avoids future ambiguity and simplifies the specification.

Refs #699


char0n:
Link checker is producing errors in CI. Seems to be unrelated with my changeset.

fmvilas (approved):
Good catch 👏

derberg (approved):

PR #779, opened 2022-05-04
docs: clarify the resolution rules of Channel Item `$ref` field

char0n:
Refs #612


fmvilas:
I think this PR can be removed in favor of #777?

char0n:
@fmvilas this change is for 2.x branch, the #777 is for 3.x branch. Assuming we don't know exactly how soon the 2.x branch of the spec goes away and it will clarify for 2.x consumers what the resolutions rules are.

derberg:
is it really needed? taking into account that the field is anyway deprecated, which means "do not use it"

char0n:
@derberg it's up to you to decide. I can only provide reasons why it would be maybe appropriate:

The only reason I created the PR is that I work on tooling implementation and I needed to implement resolution mechanism for this. I made an assumption that resolution here adheres to JSON Reference spec. This was not clear from the specification for me. Even though we say "do not use it" the mechanism is there and it should be crystal clear how it works, even though it's deprecated.

This doesn't change the specification in any way, it just clarifies something that is currently up to implement-or interpretation.

fmvilas (approved):
Looks good to me 👍

char0n:
@derberg @dalelane can I ask you to look into this PR? Thanks!

dalelane (approved):

derberg (approved):
sorry for keeping you waiting 😞 

this one doesn't have to wait for 2.5 as this is an editorial change that points to `master`

/rtm

fmvilas:

derberg:

Issue #674, opened 2021-12-14
Clarification on URL and it's resolution rules


char0n:
Hello,

In the AsyncAPI 2.2.0 spec, there are multiple fields that have description containing: `MUST be in the format of a URL`. These fields include:

#### All URL fields

- Info.termsOfService
- Contact.url
- License.url
- ExternalDocumentation.url
- OAuthFlow.authorizationUrl
- OAuthFlow.tokenUrl
- OAuthFlow.refreshUrl

Does `MUST be in the format of a URL` mean that these fields allow only absolute URL as defined by [rfc3986 section 4.3](https://datatracker.ietf.org/doc/html/rfc3986#section-4.3)? It looks like that from all the examples in spec and given that the only field inside spec that explicitly mentions relative URLs is `Server.url` field. So I'm assuming only absolute URLs are allowed here, am I right?

#### Server.url field

`Server.url` field looks like an exception to URL handling within the spec and claims that `URL MAY be relative`.  I assume that being relative complies with [rfc3986 section 4.2](https://datatracker.ietf.org/doc/html/rfc3986#section-4.2)? 

---

I'm doing a lot of assumptions here. It would be great to have this all specified explicitly so that there is not place for assumptions and different interpretations. I'm willing to issue a clarification PR after we discuss this further. 

PS: why not allow all URL fields to be relative? If the URL is relative it will be resolved using one of the `Server.url` (after it has been resolved itself against location where the AsyncAPI document is being served) as a Base URL. If no `Server Object` is defined, then the Base URL is automatically the location where the AsyncAPI document is being served.

Thanks!

derberg:
Hey, you did not come to community meeting and somehow it got lost, sorry. Stale bot to the rescue 🚀 😄 

More details in JSON Schema -> URL shoudl be of `uri` format so RFC3986. You want it explicit in the spec markdown file too, right? IMHO it definitely makes sense, users should not be forced to check JSON Schema

char0n:
> More details in JSON Schema -> URL shoudl be of uri format so RFC3986. You want it explicit in the spec markdown file too, right?

Yep, I expect the markdown file to be the single source of truth. I've created https://github.com/asyncapi/spec/issues/780 to make it explicit.

> IMHO it definitely makes sense, users should not be forced to check JSON Schema

Cool, I've created a [PR](https://github.com/asyncapi/spec/pull/782) that remedies the unclarity.

char0n:
[...]

PR #782, opened 2022-05-04
docs: clarify how URL form is resolved

char0n:
Refs #674

Along with the clarification of URL resolution, I've made the sentence that describes the fact that the value is URL consistent within the spec. This makes it possible to identify all `URL forms` within the specification using native browser search capability (CTRL+F).


dalelane (dismissed):

spec/asyncapi.md
```diff
@@ -144,6 +145,11 @@ connected parts at the discretion of the author. In the latter case, [Reference
 
 By convention, the AsyncAPI Specification (A2S) file is named `asyncapi.json` or `asyncapi.yaml`.
 
+### <a name="relativeReferencesURL"></a>Relative References in URLs
+
+Unless specified otherwise, all properties that are URLs MAY be relative references as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-4.2).
+Unless specified otherwise, relative references are resolved using the URLs defined in the [`Server Object`](#serverObject) as a Base URL. Note that these themselves MAY be relative to the referring document.
```
derberg, line 151: do you mean that if there is URL pointing to license or external docs and it is a relative URL, it should be resolved with Server URL as base URL?

char0n: @derberg yes, this is how relative references in URLs are handled in OpenAPI 3.1. The reasoning is that assets like external documentation, licenses or authentication needs to live on a server as they are concrete assets. Without server being involved, the only other `baseURI` that relative URLs can be resolved against is the location of the definition document itself, which is in most cases is not what you want.

fmvilas: I don't think we can apply the same principle here. Note that in many (most?) cases these servers are not using HTTP but a messaging protocol. How is this supposed to be served then? I understand it makes sense in OpenAPI, where everything is HTTP, but it's not clear to me it works here.

char0n: @fmvilas good point. You can technically still resolve relative URLs against not HTTP protocol based URLs, but as you pointed out it just might not make sense. But isn't it a responsibility of the definition author so that data within the definition make sense?

I can see two options:

1.) Extending the explanation of `HTTP` protocol

> Unless specified otherwise, relative references are resolved using the HTTP URLs defined in the [`Server Object`](#serverObject) as a Base URL. Note that these themselves MAY be relative to the referring document.

But one can still use use other protocols that browsers can handle like `ftp://` or `file://` or others. That goes to my point of responsibility of definition author.

2.) Provide clarification that only absolute URLs are supported

We will specify what `MUST be in the format of a URL` actually means - URLs must comply to [rfc3986 section 4.3](https://datatracker.ietf.org/doc/html/rfc3986#section-4.3)

fmvilas: I'd vote for absolute URLs. License, docs, etc. should always point to production (or final stage) environments IMHO.

char0n: > I'd vote for absolute URLs.

Well if we agree to specify absolute URLs only, it will clarify the specification and simplify the tooling implementation. So I tend to agree with this as well.

> License, docs, etc. should always point to production (or final stage) environments IMHO.

Yep, if we assume that relative URLs of license, docs, etc.. should successfully resolve against any of the `Server Object` - that will never be true, as many `Server Objects` (as you pointed out) will be defined for protocols other then HTTP or FTP or FILE.

---

I'll redo the PR. And thanks for the input.

char0n:
PR has been redo into absolute URLs only support.

spec/asyncapi.md
fmvilas: Thanks for all the work you're doing mate!

char0n:
@fmvilas @derberg @dalelane can you please have a final look at this? Thanks!

fmvilas (approved):
👍 Makes sense

dalelane (approved):

derberg (approved):
This editorial change definitely makes sense. Our JSON Schema is also aligned. We have `uri` format everywhere it is needed.

fmvilas:

derberg:

PR #794, opened 2022-05-10
feat: move root tags and externalDocs to the info object

magicmatatjahu:
---
title: "move root tags and externalDocs to the info object"
---

This PR moves `tags` and `externalDocs` (describing applications) from root to the `Info Object`. From an application description point of view, this makes much more sense, because all the metadata is in one place. This is more of a stylistic change, however it introduces a breaking change.

PR for spec-json-schemas repo https://github.com/asyncapi/spec-json-schemas/pull/244

### DIFF
--- a/examples/social-media/backend/asyncapi.yaml
+++ b/examples/social-media/backend/asyncapi.yaml
@@ -1,4 +1,4 @@
-asyncapi: 2.2.0
+asyncapi: 3.0.0
 
 info:
   title: Website Backend
--- a/examples/social-media/backend/asyncapi.yaml
+++ b/examples/social-media/backend/asyncapi.yaml
@@ -1,4 +1,4 @@
-asyncapi: 3.0.0
+asyncapi: 2.2.0
 
 info:
   title: Website Backend
--- a/examples/social-media/comments-service/asyncapi.yaml
+++ b/examples/social-media/comments-service/asyncapi.yaml
@@ -1,4 +1,4 @@
-asyncapi: 3.0.0
+asyncapi: 2.2.0
 
 info:
   title: Comments Service
--- a/examples/social-media/comments-service/asyncapi.yaml
+++ b/examples/social-media/comments-service/asyncapi.yaml
@@ -1,4 +1,4 @@
-asyncapi: 2.2.0
+asyncapi: 3.0.0
 
 info:
   title: Comments Service
--- a/examples/social-media/frontend/asyncapi.yaml
+++ b/examples/social-media/frontend/asyncapi.yaml
@@ -1,4 +1,4 @@
-asyncapi: 2.2.0
+asyncapi: 3.0.0
 
 info:
   title: Website WebSocket Client
--- a/examples/social-media/frontend/asyncapi.yaml
+++ b/examples/social-media/frontend/asyncapi.yaml
@@ -1,4 +1,4 @@
-asyncapi: 3.0.0
+asyncapi: 2.2.0
 
 info:
   title: Website WebSocket Client
--- a/examples/social-media/notification-service/asyncapi.yaml
+++ b/examples/social-media/notification-service/asyncapi.yaml
@@ -1,4 +1,4 @@
-asyncapi: 2.2.0
+asyncapi: 3.0.0
 
 info:
   title: Notifications Service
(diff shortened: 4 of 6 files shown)


dalelane (commented):
I think this makes sense. 

I know I'm normally the one who is pointing out changes that would make AsyncAPI inconsistent with OpenAPI for things that aren't specific to event driven apps, but I think v3 is going to be different enough from OAI that maybe we just do what we think makes the most sense.

magicmatatjahu:
@dalelane Thanks for your comment! I don't think it causes incompatibility with OpenAPI, because e.g. from the beginning tags in AsyncAPI are used in a different way than in OpenAPI. In OpenAPI you define tags as objects on the root, and in operations you refer to them only by name. In AsyncAPI tags as objects can be used in all places where tags may occur. The only problem may arise if someone wants to refer an Info Object from AsyncAPI in OpenAPI, then yes, this will be a problem, because the validation should inform about the use of additional fields that are not supported (`tags` and `externalDocs`).

magicmatatjahu:
@KhudaDad414 Could you check the failing `Check Markdown links` job? I guess that after merging it https://github.com/asyncapi/spec/commit/f269234668a63653616ff7b3fc9a6b249c73ee5b it should be fixed, but it's not. Please correct me if I'm wrong.

smoya (approved):
LGTM! 

As a side note, I was close to suggesting we leave the `tags` field at root level as well + add it to Info Object, but I think this will generate more confusion than benefit.

jonaslagoni:
Is anything stopping us from merging this? 🤔

dalelane (approved):

magicmatatjahu:
@jonaslagoni 

> Is anything stopping us from merging this? 🤔

yes, failing `Check Markdown links` job 😄 But perhaps one of the maintainers can merge it with admin privileges.

dalelane:
I'll see if we have more luck with re-running the job quickly first

magicmatatjahu:
I thought that problem was resolved by this PR https://github.com/asyncapi/spec/commit/f269234668a63653616ff7b3fc9a6b249c73ee5b but unfortunately no.

dalelane:
I don't think the problem is actually with the links being broken - the errors are all HTTP-429 - essentially GitHub complaining about too many requests coming too quickly.

smoya:
> Details

Ideally, the action would accept a header with a valid GH token, so request rate can be higher than actually needed. However there is no such option afaik. We can fix it by follow their recommendations at: https://github.com/gaurav-nelson/github-action-markdown-link-check#too-many-requests

derberg (approved):
LGTM

I recommend ignoring failing links. This CI check is not mandatory as still in testing phase. Pr's can be merged if it gets red. Also Khuda that is keeping an eye on it is not available at the moment. And also these errors are super strange as they never happened to us before, might be some random stuff.

derberg:
/dnm

we need changes in JSON Schema and the Parser

magicmatatjahu:
@derberg 

> we need changes in JSON Schema and the Parser

I will create PR for json-schemas. For parser the current "changes" in spec is reflected in the next parserJS version - https://github.com/asyncapi/parser-js/blob/next-major/src/models/v2/info.ts#L64 You can see that it's for 2.X.X, but due to fact that we don't know how 3.0.0 finally will look like the models are only made for version 2.X.X.

magicmatatjahu:
Changes in json-schemas https://github.com/asyncapi/spec-json-schemas/pull/244

magicmatatjahu:
Rebased with current `next-major-spec`. Could you check again? @derberg @fmvilas @dalelane Thanks!

fmvilas (approved):

magicmatatjahu:
I had to rebase with upstream `next-major-spec`. Could you check again and merge? Thanks! @derberg @fmvilas @dalelane @smoya @char0n

fmvilas (approved):
Still looking good to me 👍

fmvilas:
@magicmatatjahu do you want to merge using `/rtm` or want us to avoid Squash and Merge?

magicmatatjahu:
@fmvilas by `/rtm` but I will wait for another accepts.

magicmatatjahu:
@derberg `/rtm` doesn't work because you added `dnm` label. Could you merge it?

magicmatatjahu:
@derberg Thanks!

Issue #795, opened 2022-05-10
[3.0.0] Introduce `Info Item Object`


magicmatatjahu:
I noticed that we have some inconsistencies in describing the metadata for the 4 main elements of the AsyncAPI specification: servers, channels, operations, and messages (we can also treat schemas as 5 element). In `Server Object` we can only define `description`, in the `Channel Object` also only `description`, in the `Operation Object` we can define `description`, `operationId `, `summary`, `tags` and `externalDocs`, in the `Message Object` these same fields (except `operationId`) and also `name` and `messageId`. I wonder if we should standardise the item description with a single object type (I propose the name `Info Item Object`) that would have the following shape:

Field Name | Type | Description
---|:---:|---
name | `string` | A machine-friendly name for the item.
title | `string` | A human-friendly title for the item.
summary | `string` | A short summary of what the item is about.
description | `string` | A verbose explanation of the item. [CommonMark syntax](https://spec.commonmark.org/) can be used for rich text representation.
tags | Tags Object | A list of tags for API documentation control. Tags can be used for logical grouping of items.
externalDocs | External Documentation Object | Additional external documentation for described item.

Some example

```yaml
# server
info:
  title: Kafka server
protocol: kafka
protocolVersion: ...
``` 

```yaml
# message
info:
  name: UserSignup
  title: User signup
  summary: Action to sign a user up.
  description: ...
  tags:
    - name: user
    - name: signup
    - name: register
payload:
  ...
``` 

Benefits:
- A single info object for item will allow for easier familiarization with AsyncAPI. Currently each of the above mentioned objects in AsyncAPI has a different set of metadata, and I wonder why we have such a big discrepancy. As `html-template` and `markdown-template` are the only ones (probably) using whole metadata, it's very strange for me and every time I had to look which metadata I can render.
- Easier support in tooling. One object shape can be reused in several places, which makes the implementation of e.g. parserJS simpler and easier to understand (it is mainly about the number of methods in classes).

**NOTE**: I am also considering adding a `contact` field to the new object to describe the main "maintainers" of the objects. This would be used mainly to indicate who is responsible for a given object in an organization/company, but I don't know if that makes sense. What do you think?

**NOTE**: Should we also add `metadata` field as Łukasz suggested in https://github.com/asyncapi/spec/issues/654#issuecomment-1088414968 comment?

murari-goswami:
I like the idea here. This standardization will give more structure to the metadata for the core objects.

[...]

PR #796, opened 2022-05-12
chore: add metadata object

magicmatatjahu:
---
title: "add metadata object"
---

---

**Related issue(s):** 
Resolves https://github.com/asyncapi/spec/issues/795

---

This PR introduces new `Metadata Object` that user can use on `Server Object`, `Channel Object`, `Operation Object` and `Message Object` (and corresponding traits). `Metadata Object` contains `name`, `title`, `description`, `summary`, `tags` and `externalDocs` fields. It can contain previously non defined fields and also extensions - but I have a problem with that, because I don't know if we should allow only defining extensions and not non defined field. With this possibility there will be problems with validation on tooling side, by extension we can support it. 

The idea was discussed in https://www.youtube.com/watch?v=-qUGjylYLJg

cc @fmvilas @derberg @jonaslagoni @smoya @dalelane


spec/asyncapi.md
```diff
@@ -1401,6 +1490,64 @@ description: Find more info here
 url: https://example.com
 ```
 
+
+#### <a name="metadataObject"></a>Metadata Object
+
+The object provides metadata about the described object. May contain previously non defined fields.
```
magicmatatjahu, line 1496: I have a problem with `May contain previously non defined fields.` sentence, because I don't know if we should allow only defining extensions and not non defined field. With this possibility there will be problems with validation on tooling side, by extension we can support it, by non defined fields no.

magicmatatjahu:
@fmvilas @jonaslagoni @jessemenning New way of applying Metadata Object was introduces alongside with Server and Channel Traits in latest commits. Feedback are more that welcome :)

smoya:
I agree with some folks from https://www.youtube.com/watch?v=uoKEet0jcSc to **not** having a Metadata Object but have all those fields inline on those objects that don't have them. 
Regarding implementation on the spec, I'm up for duplicating the fields on the objects rather than creating the new "Metadata Object in there and talking about inheritance. I think this will overcomplicate the spec and its readability, adding no real value to the final user.

What I'm concerned about right now is the fact `name` is not clear to me once we have `id`s on each Object. Ok, Channel doesn't have ID but could have it, so why would we want to keep `name` field anymore? Could we then deprecate it?

magicmatatjahu:
@smoya 

> What I'm concerned about right now is the fact name is not clear to me once we have ids on each Object. Ok, Channel doesn't have ID but could have it, so why would we want to keep name field anymore? Could we then deprecate it?

We don't have id on each object, please don't look on mentioned issue :)

smoya:
> We don't have id on each object, please don't look on mentioned issue :)

By `id` I'm referring to `operationId` and `messageId`

magicmatatjahu:
@smoya 

But `operationId` and `messageId` is unique id for object, `name` is a name, you can have multiple these same names in spec and it should be still valid. I don't see any problems with that.

magicmatatjahu:
@smoya 

> Now let's think, Isn't the messageId right made for that purpose?

I don't think so. As I wrote, id should be one in spec, names you can have several. A good analogy would probably be a dictionary. You can only have unique keys but the given keys can have the same values. That's how I've always understood it.

smoya:
> @smoya I don't know. I'm not the creator of this spec :trollface:

Let's ping them cc @fmvilas.

magicmatatjahu:
@fmvilas thanks! So tl;dr; 😄 

- create separate PR for new traits,
- add all fields from Metadata Object to the destination Objects?

fmvilas:
> @fmvilas thanks! So tl;dr; 😄
> 
>     * create separate PR for new traits,
> 
>     * add all fields from Metadata Object to the destination Objects?

Yes, but only the fields that make sense or are providing real value.

spec/asyncapi.md
```diff
@@ -371,32 +391,54 @@ Field Name | Type | Description
 <a name="serverObjectUrl"></a>url | `string` | **REQUIRED**. A URL to the target host.  This URL supports Server Variables and MAY be relative, to indicate that the host location is relative to the location where the AsyncAPI document is being served. Variable substitutions will be made when a variable is named in `{`braces`}`.
 <a name="serverObjectProtocol"></a>protocol | `string` | **REQUIRED**. The protocol this URL supports for connection. Supported protocol include, but are not limited to: `amqp`, `amqps`, `http`, `https`, `ibmmq`, `jms`, `kafka`, `kafka-secure`, `anypointmq`, `mqtt`, `secure-mqtt`, `solace`, `stomp`, `stomps`, `ws`, `wss`, `mercure`, `googlepubsub`.
 <a name="serverObjectProtocolVersion"></a>protocolVersion | `string` | The version of the protocol used for connection. For instance: AMQP `0.9.1`, HTTP `2.0`, Kafka `1.0.0`, etc.
+<a name="serverObjectName"></a>name | `string` | A machine-friendly name for the server.
```
fmvilas, line 394: Isn't the server id (the key in `servers` object) already a machine-friendly name for the server? Why do we need one more property here? 🤔

spec/asyncapi.md
```diff
@@ -612,12 +654,15 @@ Field Name | Type | Description
 ---|:---:|---
 <a name="channelObjectAddress"></a>address | `string` \| `null` | An optional string representation of this channel's address. The address is typically the "topic name", "routing key", "event type", or "path". When `null` or absent, it MUST be interpreted as unknown. This is useful when the address is generated dynamically at runtime or can't be known upfront. It MAY contain [Channel Address Expressions](#channelAddressExpressions).
 <a name="channelObjectMessages"></a>messages | [Messages Object](#messagesObject) | A map of the messages that will be sent to this channel by any application at any time. **Every message sent to this channel MUST be valid against one, and only one, of the [message objects](#messageObject) defined in this map.**
+<a name="channelObjectName"></a>name | `string` | A machine-friendly name for the server.
```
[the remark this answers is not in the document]
fmvilas: If we end up leaving it, let's not forget to fix the typo:

```suggestion
<a name="channelObjectName"></a>name | `string` | A machine-friendly name for the channel.
```

spec/asyncapi.md
```diff
@@ -808,6 +855,8 @@ Field Name | Type | Description
 ---|:---:|---
 <a name="operationObjectAction"></a>action | `string` | **Required**. Allowed values are `send` and `receive`. Use `send` when it's expected that the application will send a message to the given [`channel`](#operationObjectChannel), and `receive` when the application should expect receiving messages from the given [`channel`](#operationObjectChannel).
 <a name="operationObjectChannel"></a>channel | [Reference Object](#referenceObject) | **Required**. A `$ref` pointer to the definition of the channel in which this operation is performed. Please note the `channel` property value MUST be a [Reference Object](#referenceObject) and, therefore, MUST NOT contain a [Channel Object](#channelObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience.
+<a name="operationObjectName"></a>name | `string` | A machine-friendly name for the operation.
```
fmvilas, line 858: Same as with server and channel above. Isn't it enough with the operation id?

fmvilas (changes requested):

spec/asyncapi.md
magicmatatjahu: Thanks for review! This is one field I wonder if it makes sense. Of course, we can treat the key in servers/channels/operations as an id/name of object, but in v3, servers/channels/operations/messages will be easily referenced, hence this id between references can change, and someone, for example, needs to give a "static" name for an object that will not change between refs. This makes a lot of sense for model generation and reusability of models.

If, on the other hand, the `name` doesn't make sense, we should probably remove it also from the `Message Object` - the `messageId` could be such a name for message.. What do you think?

cc @derberg @jonaslagoni @smoya @dalelane

magicmatatjahu:
As we discussed, I removed `name` property from objects (I leave `name` in Message Object - we have another discussion for it https://github.com/asyncapi/spec/issues/862). Could someone check it? :)

cc @derberg @fmvilas @dalelane @smoya @char0n @jonaslagoni

fmvilas (commented):
LGTM but you got conflicts.

magicmatatjahu:
@fmvilas Thanks for review! Done :)

fmvilas (approved):

spec/asyncapi.md
```diff
@@ -2316,7 +2362,7 @@ type: userPassword
 ```
 
 ```yaml
-type: apiKey,
+type: apiKey
```
dalelane, line 2365: nice catch :-)

dalelane (approved):
LGTM 👍

magicmatatjahu:
@derberg @smoya @char0n And I need one more approval :)

magicmatatjahu:
PR in the `parser-js` https://github.com/asyncapi/parser-js/pull/677

fmvilas:
@derberg @smoya @char0n can we help unlock this PR?

char0n (approved):
LGTM. Nice work!

magicmatatjahu:
@derberg @smoya @char0n @fmvilas @dalelane Ok, I have 3 approvals. Someone from you, feel free to merge when you have time :) Thanks!

magicmatatjahu:
Sorry! I had forgotten that automerge works here 😅

PR #803, opened 2022-06-08
docs: clarify the meaning of components

fmvilas:
This PR clarifies the meaning of the Components Object and, more importantly, the meaning of everything outside the Components Object.

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -138,6 +138,8 @@ In order to preserve the ability to round-trip between YAML and JSON formats, YA
 An AsyncAPI document MAY be made up of a single document or be divided into multiple,
 connected parts at the discretion of the author. In the latter case, [Reference Objects](#referenceObject) are used.
 
+It is important to note that everything that is defined in an AsyncAPI document MUST be used by the implemented [Application](#definitionsApplication), with the exception of the [Components Object](#componentsObject). Everything that is defined inside the Components Object represents a resource that MAY or MAY NOT be used in the rest of the document and MAY or MAY NOT be used by the implemented [Application](#definitionsApplication).
+
 By convention, the AsyncAPI Specification (A2S) file is named `asyncapi.json` or `asyncapi.yaml`.
 
 ### <a name="schema"></a>Schema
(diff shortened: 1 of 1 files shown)


spec/asyncapi.md
```diff
@@ -138,6 +138,8 @@ In order to preserve the ability to round-trip between YAML and JSON formats, YA
 An AsyncAPI document MAY be made up of a single document or be divided into multiple,
 connected parts at the discretion of the author. In the latter case, [Reference Objects](#referenceObject) are used.
 
+It is important to note that everything that is defined in an AsyncAPI document MUST be used by the implemented [Application](#definitionsApplication), with the exception of the [Components Object](#componentsObject). Everything that is defined inside the Components Object represents a resource that MAY or MAY NOT be used in the rest of the document and MAY or MAY NOT be used by the implemented [Application](#definitionsApplication).
```
fmvilas, line 141: Not super happy with this description. I'm strongly accepting suggestions to improve it.

Also, not sure if this belongs here but I think it makes sense to put it under the File Structure section.

fmvilas: I see what you mean now. It's actually not necessary to have that last part 👍

smoya:
Is this really introducing a breaking change? I'm not very sure and I would say it doesn't. If you think it doesn't, would you like to include this into 2.4.1 release? This is ready to be merged: https://github.com/asyncapi/spec/pull/776 but we could pack both PR's into the same release.

jonaslagoni:
What can we do to move this forward, is it only reviews we need @fmvilas ? 🙂

fmvilas:
Yup, reviews are needed. Not just from @dalelane and @derberg but also y'all.

dalelane (approved):

derberg (approved):

spec/asyncapi.md
```diff
@@ -157,7 +159,7 @@ Field Name | Type | Description
 <a name="A2SServers"></a>servers | [Servers Object](#serversObject) | Provides connection details of servers.
 <a name="A2SDefaultContentType"></a>defaultContentType | [Default Content Type](#defaultContentTypeString) | Default content type to use when encoding/decoding a message's payload.
 <a name="A2SChannels"></a>channels | [Channels Object](#channelsObject) | **Required** The available channels and messages for the API.
-<a name="A2SComponents"></a>components | [Components Object](#componentsObject) | An element to hold various schemas for the specification.
+<a name="A2SComponents"></a>components | [Components Object](#componentsObject) | An element to hold various reusable objects for the specification. Everything that is defined inside this object represents a resource that MAY or MAY NOT be used in the rest of the document and MAY or MAY NOT be used by the implemented [Application](#definitionsApplication).
```
smoya, line 162: I think makes sense to also add this clarification under `Components Object` description.

fmvilas: Yup, agree. Would you mind opening the PR?

PR #759, opened 2022-04-19
feat!: release for version 3.0.0 of the spec

jonaslagoni:
This PR is created upfront to reflect on a daily basis what things are included in the release.

Also, the reason to create this PR a long time before the release is to enable automation (bot keeps upstream branches always up to date with master, see this [action](https://github.com/asyncapi/spec/blob/master/.github/workflows/autoupdate.yml)) that we have in place to regularly update the release branch with whatever is changed in the master branch. So nobody has to do it manually.

#### TO DO:

* [ ] Cherry-pick #753 commit right before merging this PR. There's a backup branch where you'll be able to find this commit: https://github.com/asyncapi/spec/tree/next-major-spec-backup.

PR #806, opened 2022-06-13
feat!: move operations to its own root object

fmvilas:
---
title: "Move operations to its own root object"
---

---

**Related issue(s):**

#618 #663

---

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -768,7 +768,7 @@ onUserSignUp:
 
 #### <a name="operationObject"></a>Operation Object
 
-Describes a `send` or a `receive` operation.
+Describes a specific operation.
 
 ##### Fixed Fields
 
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -169,7 +169,7 @@ Field Name | Type | Description
 <a name="A2SServers"></a>servers | [Servers Object](#serversObject) | Provides connection details of servers.
 <a name="A2SDefaultContentType"></a>defaultContentType | [Default Content Type](#defaultContentTypeString) | Default content type to use when encoding/decoding a message's payload.
 <a name="A2SChannels"></a>channels | [Channels Object](#channelsObject) | The channels used by this [application](#definitionsApplication).
-<a name="A2SOperations"></a>operations | [Operations Object](#operationsObject) | The operations this [application](#definitionApplication) MUST implement.
+<a name="A2SOperations"></a>operations | [Operations Object](#operationsObject) | The operations this [application](#definitionsApplication) MUST implement.
 <a name="A2SComponents"></a>components | [Components Object](#componentsObject) | An element to hold various reusable objects for the specification. Everything that is defined inside this object represents a resource that MAY or MAY NOT be used in the rest of the document and MAY or MAY NOT be used by the implemented [Application](#definitionsApplication).
 <a name="A2STags"></a>tags | [Tags Object](#tagsObject) | A list of tags used by the specification with additional metadata. Each tag name in the list MUST be unique.
 <a name="A2SExternalDocs"></a>externalDocs | [External Documentation Object](#externalDocumentationObject) | Additional external documentation.
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -713,7 +713,7 @@ userCompletedOrder:
 
 Holds a dictionary with all the [operations](#operationObject) this application MUST implement.
 
-Note: If you're looking for a place to define operations that MAY or MAY NOT be implemented by the application, consider defining them in [`components/operations`](#componentsOperations).
+> If you're looking for a place to define operations that MAY or MAY NOT be implemented by the application, consider defining them in [`components/operations`](#componentsOperations).
 
 ##### Patterned Fields
 
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -1,6 +1,6 @@
 # ATTENTION: Work in progress
 
-This version is not yet ready to be used. We're currently working on it. If you want to join the effort and participate in the development of the next major version of AsyncAPI, head over to https://github.com/asyncapi/spec/issues/691.
+This version is not yet ready to be used. We're currently working on it. If you want to join the effort and participate in the development of the next major version of AsyncAPI, head over to [GitHub Issue that we use for tracking 3.0 development progress](https://github.com/asyncapi/spec/issues/691).
 
 # AsyncAPI Specification
 
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -719,7 +719,7 @@ Holds a dictionary with all the [operations](#operationObject) this application 
 
 Field Pattern | Type | Description
 ---|:---:|---
-<a name="operationsObjectOperation"></a>{operationId} | [Operation Object](#channelItemObject) \| [Reference Object](#referenceObject) | The operation this application MUST implement. The field name (`operationId`) MUST be a string used to identify the operation in the document where it is defined, and its value is **case-sensitive**. Tools and libraries MAY use the `operationId` to uniquely identify an operation, therefore, it is RECOMMENDED to follow common programming naming conventions.
+<a name="operationsObjectOperation"></a>{operationId} | [Operation Object](#channelObject) \| [Reference Object](#referenceObject) | The operation this application MUST implement. The field name (`operationId`) MUST be a string used to identify the operation in the document where it is defined, and its value is **case-sensitive**. Tools and libraries MAY use the `operationId` to uniquely identify an operation, therefore, it is RECOMMENDED to follow common programming naming conventions.
 
 ##### Operations Object Example
 
@@ -779,7 +779,7 @@ Describes a specific operation.
 Field Name | Type | Description
 ---|:---:|---
 <a name="operationObjectAction"></a>action | `string` | **Required**. Allowed values are `send` and `receive`. Use `send` when it's expected that the application will send a message to the given [`channel`](#operationObjectChannel), and `receive` when the application should expect receiving messages from the given [`channel`](#operationObjectChannel).
-<a name="operationObjectChannel"></a>channel | [Reference Object](#referenceObject) | **Required**. A `$ref` pointer to the definition of the channel in which this operation is performed. Please note the `channel` property value MUST be a [Reference Object](#referenceObject) and, therefore, MUST NOT contain a [Channel Object](#channelItemObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience.
+<a name="operationObjectChannel"></a>channel | [Reference Object](#referenceObject) | **Required**. A `$ref` pointer to the definition of the channel in which this operation is performed. Please note the `channel` property value MUST be a [Reference Object](#referenceObject) and, therefore, MUST NOT contain a [Channel Object](#channelObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience.
 <a name="operationObjectSummary"></a>summary | `string` | A short summary of what the operation is about.
 <a name="operationObjectDescription"></a>description | `string` | A verbose explanation of the operation. [CommonMark syntax](http://spec.commonmark.org/) can be used for rich text representation.
 <a name="operationObjectSecurity"></a>security | [[Security Requirement Object](#securityRequirementObject)]| A declaration of which security mechanisms are associated with this operation. Only one of the security requirement objects MUST be satisfied to authorize an operation. In cases where Server Security also applies, it MUST also be satisfied.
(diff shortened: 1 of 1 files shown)


spec/asyncapi.md
```diff
@@ -689,63 +689,101 @@ bindings:
...
+<a name="operationObjectAction"></a>action | `string` | **Required**. Allowed values are `send` and `receive`. Use `send` when it's expected that the application will send a message to the given [`channel`](#operationObjectChannel), and `receive` when the application should expect receiving messages from the given [`channel`](#operationObjectChannel).
+<a name="operationObjectChannel"></a>channel | [Reference Object](#referenceObject) | **Required**. A `$ref` pointer to the definition of the channel in which this operation is performed. Please note the `channel` property value MUST be a [Reference Object](#referenceObject) and, therefore, MUST NOT contain a [Channel Object](#channelItemObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience.
```
smoya, line 762: As per the whole discussion in https://github.com/asyncapi/spec/issues/618, the `channel` field would expect a plain string, being the id of a channel. I guess by ID we would refer to https://github.com/asyncapi/spec/pull/719.

However, in the same issue linked to that PR, https://github.com/asyncapi/spec/issues/663#issuecomment-1036133392, we do have a discussion around using `$ref` as you added here.

Just to be more aligned, Is there any other discussion on this?

spec/asyncapi.md
```diff
@@ -808,15 +834,16 @@ traits:
...
-<a name="operationTraitObjectOperationId"></a>operationId | `string` | Unique string used to identify the operation. The id MUST be unique among all operations described in the API. The operationId value is **case-sensitive**. Tools and libraries MAY use the operationId to uniquely identify an operation, therefore, it is RECOMMENDED to follow common programming naming conventions.
+<a name="operationTraitObjectAction"></a>action | `string` | Allowed values are `send` and `receive`. Use `send` when it's expected that the application will send a message to the given [`channel`](#operationObjectChannel), and `receive` when the application should expect receiving messages from the given [`channel`](#operationObjectChannel).
+<a name="operationTraitObjectChannel"></a>channel | [Reference Object](#referenceObject) | A `$ref` pointer to the definition of the channel in which this operation is performed. Please note the `channel` property value MUST be a [Reference Object](#referenceObject) and, therefore, MUST NOT contain a [Channel Object](#channelItemObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience.
```
smoya, line 846: In the hypothetical case we finally do this, then it will make sense to apply the same principle to the `servers` field for the **Channel Object** https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md#channel-item-object**

fmvilas: Yup, makes sense.

spec/asyncapi.md
fmvilas: @smoya @derberg @dalelane Nobody argued against https://github.com/asyncapi/spec/issues/663#issuecomment-1158094050. Does it mean are you ok if we move forward with this solution?

derberg: added comment in linked issue

fmvilas:
/progress 60

New `operations` object is defined but there's still a pending discussion about how to reference the channel, i.e., via ID or `$ref` object. Read more on #663.

derberg:
/progress 20

did it work :trollface:? or did you make sure that only your "progress comments" are taken into account? 😄 

and regarding PR:
- please add new object to ToC of the spec
- it is a new root object, so should be added to `AsyncAPI Object`
- I'm definitely up with using `$ref` for channel reference, let's be consistent everywhere though
- I'm guessing the message goes away from operation as it stays in the channel, right? and it is handled here https://github.com/asyncapi/spec/pull/827/ so all good from my side

> New operations object is defined but there's still a pending discussion about how to reference the channel, i.e., via ID or $ref object. Read more on https://github.com/asyncapi/spec/issues/663.

@fmvilas I think we just need @dalelane view on that, or are you waiting for someone else?

fmvilas:
/progress 60

> did it work :trollface:? or did you make sure that only your "progress comments" are taken into account? 😄

It did work because it's meant to be used as a team so I just have to rely on the hope that you would not troll me so much 😅  (damn, as I'm writing this I'm already realizing I'm dead 😂)

> I think we just need @dalelane view on that, or are you waiting for someone else?

Dale's point of view for sure. Would also love to have the opinions of —at least— @smoya, @char0n, and @magicmatatjahu. And of course, anyone else reading this is also welcome. I know you know but just so everyone knows :)

---

Regarding the review, thanks! I'll address all those points 👍

fmvilas:
Alright, I think I addressed all the points @derberg. Would you please review it again?

spec/asyncapi.md
```diff
@@ -694,37 +696,97 @@ bindings:
...
 
-Describes a publish or a subscribe operation. This provides a place to document how and why messages are sent and received.
+##### Operations Object Example
+
+```json
+{
+  "onUserSignUp": {
+    "summary": "Action to sign a user up.",
+    "description": "A longer description",
+    "channel": {
```
smoya, line 718: I'm thinking that it might make sense to convert this into an array of channels (of references to channel) instead.
The use case that came to my mind was the **fanout** pattern,  sending the same message to different channels.

fmvilas: AFAIK, the fanout pattern is usually implemented by the broker, not the application. I mean, for instance, RabbitMQ has the fanout exchange that lets you connect various queues to a specific routingKey. Kafka, NATS, and others have consumer groups, with which you can implement the fanout pattern. The application sends a message to a particular channel and then it gets distributed to many other queues or consumers. Am I missing something?

Now, I see the value it can have by allowing us to group multiple operations into one. Batch operations if you want. We just need to be very clear that if a `send` operation is performed against 1 out of 3 channels and it fails, it shouldn't stop the code from sending the other 2. Aren't we getting into the land of workflow definition already? 🤔

smoya: > AFAIK, the fanout pattern is usually implemented by the broker, not the application.

This statement doesn't reflect the reality out there. The fanout could be perfectly implemented by the application; writing the same message to different Kafka topics is a completely valid use case. For logistics, you might want to set up different retention policies, permissions, etc to each topic.

Also, we can still use an AsyncAPI document to describe a message broker (wich turns to be the application), where this pattern is way more common.
At the end, a client should know where to connect to and to which channels a particular message will be sent so it can subscribe to one of those (depending on each scenario).

> Aren't we getting into the land of workflow definition already?

TBH I don't think we are that far. This is still referring to an application definition.

fmvilas: Yeah, that's my point. I wouldn't say it's to support fanout but to batch operations. That's a pretty valid and common use case.

@derberg @dalelane @char0n @magicmatatjahu what do you folks think?

smoya: I don't get why it's not to support the [fanout pattern](https://en.wikipedia.org/wiki/Fan-out_(software)). Messages are sent in parallel to different channels.

From Wikipedia:

>In message-oriented middleware solutions, fan-out is a messaging pattern used to model an information exchange that implies the delivery (or spreading) of a message to one or multiple destinations possibly in parallel, and not halting the process that executes the messaging to wait for any response to that message.

fmvilas: Sorry, my point was that you can still do fanout without making it an array. In any case, check that definition mate, it clearly says "to model an **information exchange**" 😄 Exchanges are always part of the broker. If your application is sending the same message three times to three different destinations, it's not fanout, it's just a single publish three times. The advantage of fanout is that you only publish once and the message is spread to multiple destinations. I only have to succeed to send the message once and I only need to know about one destination. If your application is sending a message 3 times:

1. It needs to know the 3 destinations. Not a constraint for fanout but definitely misses one of the strongest points of fanout.
2. You need to make sure your application doesn't halt in the middle of the process, which may not be always possible, making the "fanout" unreliable.

That said, I don't care. We seem to understand each other, we just don't agree on what fanout means. Again, I think the proposal still has a point because it lets you batch/group operations. Whether we call it "fanout" or "batch operations" here, is not relevant for the discussion IMHO.

smoya: Indeed we ended up talking about the same thing, each one believing strongly in its own point 😆. 
As you shared, let's don't care about how to call it but focus then on this possible use case that we should consider. 

Let's see what others say.

spec/asyncapi.md
fmvilas: FYI https://github.com/asyncapi/spec/pull/827/commits/58e9d0725a2883d267a4132ce8fefc251d85fd09

spec/asyncapi.md
derberg: it was indeed an interesting conversation,

I learned a lot.

spec/asyncapi.md
```diff
@@ -694,37 +696,97 @@ bindings:
 
 
 
+#### <a name="operationsObject"></a>Operations Object
 
+Holds a dictionary with all the [operations](#operationObject) this application MUST implement.
 
-#### <a name="operationObject"></a>Operation Object
+Note: If you're looking for a place to define operations that MAY or MAY NOT be implemented by the application, consider defining them in [`components/operations`](#componentsOperations).
```
derberg, line 703:
```suggestion
> If you're looking for a place to define operations that MAY or MAY NOT be implemented by the application, consider defining them in [`components/operations`](#componentsOperations).
```

yeah, whenever I see "Node:" I say "let us use markdown" - but just a small suggestion

derberg (approved):
just to be clear, after we merge it, a release candidate will be released and published on the website

spec/asyncapi.md
fmvilas: And what do you think about making it an array?

fmvilas:
> just to be clear, after we merge it, a release candidate will be released and published on the website

🤔 Should we avoid it for now? I just want to make progress towards 3.0.0 but we're still far. Should I use `chore` in this case?

derberg:
tbh, I'm pro-pre-releases 😄  as they raise awareness. They will be published to asyncapi.com and will make it pretty clear to users going to "docs" that 3.0 is on the horizon. Docs is the most visited part of the website. We could add some big note at the beginning of the spec file, with some essential links on how to engage in works towards 3.0, just a thought.

fmvilas:
> We could add some big note at the beginning of the spec file, with some essential links on how to engage in works towards 3.0, just a thought.

That's a good idea. Will add it.

fmvilas:
@derberg done. Let me know what you think.

spec/asyncapi.md
derberg: > For example, a user sign up action on a server might result in the server sending the same message to several queues on an EDA. Let's say those queues are owned by different departments, teams, or whatever.,

I can't imagine such a situation, really. Knowing how things work, I bet that devs asked to make sure they send the same message to another dedicated channel for some team, will just tell people to subscribe to the existing channel 🤷🏼 

delivering one message to 2 channels is basically duplicating messages, right? cause they are different as they will have to have different tracing id?

fmvilas: The only use case that comes to my mind right now is auditing. Some messages will be sent to their corresponding channel + the auditing channel. But I've never seen this done in the application code. It's always configured as a rule in the broker so you don't forget to audit anything.

spec/asyncapi.md
```diff
@@ -1,3 +1,7 @@
+# ATTENTION: Work in progress
+
+This version is not yet ready to be used. We're currently working on it. If you want to join the effort and participate in the development of the next major version of AsyncAPI, head over to https://github.com/asyncapi/spec/issues/691.
```
derberg, line 3:
```suggestion
This version is not yet ready to be used. We're currently working on it. If you want to join the effort and participate in the development of the next major version of AsyncAPI, head over to [GitHub Issue that we use for tracking 3.0 development progress](https://github.com/asyncapi/spec/issues/691).
```

to improve accessibility of the link

spec/asyncapi.md
smoya: The use case I had in my mind is illustrated in the following pic grabbed from Salesforce blog. In particular, I'm talking about **Point-to-Point** architecture (even though the name is not correct IMHO) and not entering into which one is more scalable or less because it's not the main topic being discussed here.

derberg: but point to point is actually in contrast to EDA, right?

smoya: > but point to point is actually in contrast to EDA, right?

The name on the picture is not correct as I mentioned in my previous comment. The right name I'm looking for is **EventDriven Integration**, and this article from Solace explains what it is: https://solace.com/blog/event-driven-architecture-difference-event-driven-integration

jonaslagoni (approved):
Niiice 🔥

spec/asyncapi.md
fmvilas: @smoya It looks like this requires its own discussion. Would you mind creating a follow-up issue? It would be great to find actual cases where this is required. Salesforce is great but I don't think it's a common use case.

derberg:
@fmvilas 

- please remember not to merge this PR until https://github.com/asyncapi/spec/pull/759#issuecomment-1253666349 is solved and the release workflow, otherwise we are not able to release RC
- update PR title to `feat!:` as we are actually suppose to release a major release candidate

fmvilas:
@dalelane mind having a look?

dalelane (approved):
LGTM 👍

PR #807, opened 2022-06-13
chore: merge next-major-spec with latest master

fmvilas:
Merge `next-major-spec` with latest master


smoya (approved):

fmvilas:
Yup. Moving this convo to #759.

fmvilas:
Yeah, it's like every link is failing 😅 Or I'm failing to read the checks properly 😄

derberg:
@fmvilas yeah you need to update with latest master again, where @KhudaDad414 made fixes

fmvilas:
You're right 👍

fmvilas:
I wished 😄

fmvilas:
@derberg CI still broken in this PR. Any idea why?

derberg:
We need to investigate. It is not yet a blocking CI check so go ahead and merge. Some of fails are because there are already future links for 3.0. should be ignored in config. But I don't think it is the only issue

Issue #465, opened 2020-11-25
Support for multiple Kafka clusters within a spec


dalelane:
**Is your feature request related to a problem? Please describe.**

A Kafka cluster is typically made of a group of Kafka brokers. The brokers act as peers, so a client is able to make an initial connection to any broker in the cluster, and a metadata exchange takes place to inform the client which broker it should connect to.

To enable the cluster to be highly-available to connections from clients, Kafka clients are typically configured with the address of every broker in the Kafka cluster - so that it can try each broker in turn in the event that the first broker it attempts to connect to is unavailable.

Currently, this is being modelled in AsyncAPI by identifying each broker as a separate `server` in the list of servers. For the purposes of code generation, an assumption is made that all the Kafka servers listed in the `servers` section of the spec belong to the same cluster, and so all server URLs can be combined to provide a single bootstrap servers list.

**Describe the solution you'd like**

I'd like a way to identify multiple separate Kafka clusters within a single AsyncAPI spec.

For example, imagine I have a development/test Kafka cluster composed of three brokers A, B, C
and a production Kafka cluster composed of three brokers D, E, F

It's not safe to put all six broker addresses as six separate server objects in the spec, as I don't have a way to identify the grouping, and current code-generation would treat it as a single Kafka cluster with six brokers in.

**Additional context**

- [Slack discussion](https://asyncapi.slack.com/archives/C34F2JV0U/p1606297904403300)
- [Current Java template behaviour](https://github.com/asyncapi/java-spring-template/blob/661b7c387af76f8d02c2f850a925c9a66e213efb/template/src/main/resources/application.yml#L61-L64)
- [Related discussion](https://github.com/asyncapi/java-spring-template/pull/55#issuecomment-635824053)

derberg:
Copying over discussion from Slack as we don't know when it will dissapear cause of the free plan

```

Dale Lane Yesterday at 10:51 AM
It's customary in Kafka to provide client applications with a comma-separated list of brokers in the cluster (so the client is able to attempt connection to any of the broker addresses in that list)
Are there any issues with putting a comma-separated list in the url field of a Server object for this purpose?
https://www.asyncapi.com/docs/specifications/2.0.0#fixed-fields-4
(or are there better ways I should represent this? like having each broker as a separate server object?) (edited) 




9 replies

Lukasz Gornicki  22 hours ago
[...]

PR #809, opened 2022-06-17
feat: add tags field to Server Object

smoya:
**Description**

This PR adds support for defining Tags at Server Object. They can be used for providing metadata, including the environment such as `production`, `development`, etc.

**Related issue(s)**
Fixes https://github.com/asyncapi/spec/issues/654
Fixes https://github.com/asyncapi/spec/issues/465

### DIFF
--- a/examples/social-media/backend/asyncapi.yaml
+++ b/examples/social-media/backend/asyncapi.yaml
@@ -7,12 +7,26 @@ info:
 servers:
   websiteWebSocketServer:
     $ref: '../common/servers.yaml#/websiteWebSocketServer'
+  tags:
+    - name: "env:production"
+      description: "This environment is meant for production use case"
+    - name: "kind:local"
+      description: "This server is a local server. It is exposed by the application"
+    - name: "visibility:public"
+      description: "This resource is public and available to everyone"
   mosquitto:
     url: mqtt://test.mosquitto.org
     protocol: mqtt
     bindings:
       mqtt:
         clientId: websocketServer
+    tags:
+      - name: "env:production"
+        description: "This environment is meant for production use case"
+      - name: "kind:remote"
+        description: "This server is a remote server. Not exposed by the application"
+      - name: "visibility:public"
+        description: "This resource is public and available to everyone"
 
 channels:
   comment/liked:
--- a/examples/social-media/backend/asyncapi.yaml
+++ b/examples/social-media/backend/asyncapi.yaml
@@ -7,13 +7,6 @@ info:
 servers:
   websiteWebSocketServer:
     $ref: '../common/servers.yaml#/websiteWebSocketServer'
-  tags:
-    - name: "env:production"
-      description: "This environment is meant for production use case"
-    - name: "kind:local"
-      description: "This server is a local server. It is exposed by the application"
-    - name: "visibility:public"
-      description: "This resource is public and available to everyone"
   mosquitto:
     url: mqtt://test.mosquitto.org
     protocol: mqtt
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -397,7 +397,10 @@ The following shows how multiple servers can be described, for example, at the A
       "protocol": "amqp",
       "protocolVersion": "0.9.1",
       "tags": [
-        { "name": "env:development" }
+        { 
+          "name": "env:development",
+          "description": "This environment is meant for developers to run their own tests"
+        }
       ]
     },
     "staging": {
@@ -406,7 +409,10 @@ The following shows how multiple servers can be described, for example, at the A
       "protocol": "amqp",
       "protocolVersion": "0.9.1",
       "tags": [
-        { "name": "env:staging" }
+        { 
+          "name": "env:staging",
+          "description": "This environment is a replica of the production environment"
+        }
       ]
     },
     "production": {
@@ -415,7 +421,10 @@ The following shows how multiple servers can be described, for example, at the A
       "protocol": "amqp",
       "protocolVersion": "0.9.1",
       "tags": [
-        { "name": "env:production" }
+        { 
+          "name": "env:production",
+          "description": "This environment is the live environment available for final users"
+        }
       ]
     }
   }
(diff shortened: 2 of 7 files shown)


spec/asyncapi.md
```diff
@@ -361,6 +361,7 @@ Field Name | Type | Description
 <a name="serverObjectDescription"></a>description | `string` | An optional string describing the host designated by the URL. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
 <a name="serverObjectVariables"></a>variables | Map[`string`, [Server Variable Object](#serverVariableObject)] | A map between a variable name and its value.  The value is used for substitution in the server's URL template.
 <a name="serverObjectSecurity"></a>security | [[Security Requirement Object](#securityRequirementObject)] | A declaration of which security mechanisms can be used with this server. The list of values includes alternative security requirement objects that can be used. Only one of the security requirement objects need to be satisfied to authorize a connection or operation.
+<a name="serverObjectTags"></a>tags | [Tags Object](#tagsObject) | A list of tags for API documentation control. Tags can be used for logical grouping of servers.
```
fmvilas, line 364: Let's avoid using API and documentation. These are just specific use cases IMHO. What about this:

```suggestion
<a name="serverObjectTags"></a>tags | [Tags Object](#tagsObject) | A list of tags for logical grouping and categorization of servers.
```

fmvilas (changes requested):

spec/asyncapi.md
fmvilas: That would be awesome yes 👍

smoya: Okey changes applied.

However, cc @alequetzalli in case you want to take this opportunity to add, remove, fix, and or rephrase this description.

dalelane (approved):

derberg (commented):
I would be good to extend some of the official examples, like streetlights for kafka with this feature, as this use case is most common, that Kafka has multiple production servers. Could make sense to also add example `description` in `Tags` object.

separate question:
I think we need to mention at least in this PR that we are aware there is a feature that allows you too specify that some channel is available only on a given server (`CHANNEL_NAME.servers`).

I somehow feel that people will immediately assume, they can refer not only to a server by server name but the tag too 😄

smoya:
> I would be good to extend some of the official examples, like streetlights for kafka with this feature, as this use case is most common, that Kafka has multiple production servers. 

I don't fully get what you mean by `Kafka has multiple production servers`. Would you mind sharing an example or clarifying it? 🙏 

> Could make sense to also add example `description` in `Tags` object.

Good suggestion! I've pushed a new commit with the requested change.

> separate question: I think we need to mention at least in this PR that we are aware there is a feature that allows you too specify that some channel is available only on a given server (`CHANNEL_NAME.servers`).
> 
> I somehow feel that people will immediately assume, they can refer not only to a server by server name but the tag too 😄

TBH I don't see the relation between `CHANNEL_NAME.servers` and the tags. Those are not selectors nor similar.

dalelane (approved):

derberg:
> I don't fully get what you mean by Kafka has multiple production servers. Would you mind sharing an example or clarifying it? 🙏

so with Kafka you can have a cluster of brokers. Related discussion https://github.com/asyncapi/spec/issues/465

I'm pretty sure people will use `tag` to address above.

> TBH I don't see the relation between CHANNEL_NAME.servers and the tags. Those are not selectors nor similar.

Am I really the only one (with Kafka cluster use case in mind) that thinks that people will like to do `CHANNEL_NAME.servers=['env:production']` 😄

fmvilas:
> Am I really the only one (with Kafka cluster use case in mind) that thinks that people will like to do CHANNEL_NAME.servers=['env:production'] 😄

This is a really good point 🤔 Maybe something to consider for 3.0.0?

smoya:
@fmvilas @dalelane @derberg a last review so we can include this into 2.5 finally?

fmvilas (approved):

derberg:
is this PR resolving #654? I think yes, so please update description

do we agree that this PR also solves #465? 👀 @dalelane that wanted to champion the other proposal

---

@smoya 👇🏼 was not yet addressed

> I would be good to extend some of the official examples, like streetlights for kafka with this feature, as this use case is most common, that Kafka has multiple production servers

and regarding 👇🏼 

> TBH I don't see the relation between CHANNEL_NAME.servers and the tags. Those are not selectors nor similar.

I think we have an agreement that for 3.0 we want to go `$ref`-everywhere direction, including `CHANNEL_NAME.servers`, so I guess we can just leave it as it is without any concerns.

smoya:
> @smoya 👇🏼 was not yet addressed

@derberg I updated the `socialmedia`, `streetlights-kafka` and `streetlights-mqtt` examples. Can you take a look to those? 🙏

examples/social-media/backend/asyncapi.yaml
```diff
@@ -7,12 +7,26 @@ info:
 servers:
   websiteWebSocketServer:
     $ref: '../common/servers.yaml#/websiteWebSocketServer'
+  tags:
```
derberg, line 10: I don't think this one is correct, right?

derberg:
@dalelane can you have a final look please, ref -> https://github.com/asyncapi/spec/pull/809#issuecomment-1252499990

dalelane (approved):
This looks really good - I like the examples, as they are super helpful prompts for the ways that tags can be used, instead of just mechanically showing how to add a tag 👍

derberg:
@dalelane thanks, does it also solve https://github.com/asyncapi/spec/issues/465 from your perspective?

derberg (approved):

dalelane:
> @dalelane thanks, does it also solve #465 from your perspective?

yeah I think so - I'll add a comment there, too

derberg:
@smoya can you update the description accordingly so the other issue is also resolved after merge, by default?

and then merge 💪🏼

derberg:
@magicmatatjahu something we might want to support in react component out of the box after release 😉

smoya:
> @smoya can you update the description accordingly so the other issue is also resolved after merge, by default?
> 
> and then merge 💪🏼

Done at the same time you posted this comment ⚡ 🏃

magicmatatjahu:
@derberg I will remember :)

char0n:
Looks all good here and ready for 2.5.0, going for a merge. Thanks @smoya!

PR #815, opened 2022-06-27
docs: end OAuth Flows Object field descriptions always with dot

char0n:
Some fixed fields description have dots and some don't. This PR provides remediation and introduces a little bit more consistency.

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -2346,8 +2346,8 @@ Allows configuration of the supported OAuth Flows.
 ##### Fixed Fields
 Field Name | Type | Description
 ---|:---:|---
-<a name="oauthFlowsImplicit"></a>implicit| [OAuth Flow Object](#oauthFlowObject) | Configuration for the OAuth Implicit flow
-<a name="oauthFlowsPassword"></a>password| [OAuth Flow Object](#oauthFlowObject) | Configuration for the OAuth Resource Owner Protected Credentials flow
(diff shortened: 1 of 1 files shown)


smoya (changes requested):
LGTM, however I think `docs:` commit prefix fits better than `refactor:`.

char0n:
> LGTM, however I think `docs:` commit prefix fits better than `refactor:`.

Sure, can be address by `squash and merge` workflow, or do you want me to force push a new amended commit msg?

smoya:
> > LGTM, however I think `docs:` commit prefix fits better than `refactor:`.
> 
> Sure, can be address by `squash and merge` workflow, or do you want me to force push a new amended commit msg?

You just need to change it in the PR title. The PR title is used as the commit message when squashing.

char0n:
Done.

fmvilas (dismissed):
Looks good to me too.

@derberg is our CI aware of things like `docs(asyncapi.md):` in the PR title? Or should we just use `docs:` or `chore:`?

derberg:
@char0n no worries, happy to explain.

The only difference is that if you make an editorial change against `next-major`, then the change will wait until the next release window to get released. And if the change is introduced against `master`, and with `docs:` prefix, then we update it also automatically on asyncapi.com

so yeah, `editorial changes` is a key here, because if it is more, then yeah, should not go to `master`

fmvilas:

derberg:

PR #823, opened 2022-07-16
fix: add reference object to server variables field

lorenzsimon:
---
title: Add reference object to server variables field
---
I noticed that some changes from release 2.3.0 and 2.4.0 are not reflected in the specs yet. The option to use a reference object instead of a channel item object was already added to this branch.

My PR documents the option to use a reference object instead of a server variable object. If I understand the release notes correctly, this feature was introduced in 2.4.0.

---

**Related issue(s):**
none

---


fmvilas:
Hey @lorenzsimon. Thanks for taking the time. @char0n already fixed it at #776 and will be included in the next release. I'm closing as a duplicate.

PR #827, opened 2022-08-27
feat: add new channels object

fmvilas:
**Related issue(s):**
#663
#682  
#618


fmvilas:
@GreenRover this PR should unblock your request/reply work.

spec/asyncapi.md
```diff
@@ -529,167 +529,197 @@ defaultContentType: application/json
...
+      userSignedUp:
+        $ref: '#/components/messages/userSignedUp'
+      userCompletedOrder:
+        $ref: '#/components/messages/userCompletedOrder'
+    parameters:
+      userId:
+        $ref: '#/components/parameters/userId'
+    servers:
+      - rabbitmqInProd
+      - rabbitmqInStaging
```
magicmatatjahu, line 652: That change shouldn't be related to this PR, but shouldn't we also allow refs in channel's `servers`? I see that instead of IDs here references are used for channels and messages, so for servers they should too, like:

```yaml
servers:
  - $ref: '#/servers/rabbitmqInProd'
  - $ref: '#/servers/rabbitmqInStaging'
```

I think we even had a discussion on this topic once at some 3.0.0 meeting but I can't remember at which one.

WDYT? cc @smoya

smoya: I think you mean my mention in https://github.com/asyncapi/spec/pull/806#discussion_r900012119 right?

magicmatatjahu: Exactly, the same logic, but I remember that there was a discussion about it at some meeting about the servers themselves.

fmvilas: Yeah, if we end up accepting this PR, we'll have to open an issue to change everything else too. But it also really depends on the output of https://github.com/asyncapi/spec/issues/663.

fmvilas:
/progress 60

New `channels` shape is defined but it's still pending review by code owners.

fmvilas:
@derberg @dalelane @smoya would you mind having a look?

spec/asyncapi.md
```diff
@@ -162,7 +162,7 @@ Field Name | Type | Description
 <a name="A2SInfo"></a>info | [Info Object](#infoObject) | **Required.** Provides metadata about the API. The metadata can be used by the clients if needed.
 <a name="A2SServers"></a>servers | [Servers Object](#serversObject) | Provides connection details of servers.
 <a name="A2SDefaultContentType"></a>defaultContentType | [Default Content Type](#defaultContentTypeString) | Default content type to use when encoding/decoding a message's payload.
-<a name="A2SChannels"></a>channels | [Channels Object](#channelsObject) | **Required** The available channels and messages for the API.
+<a name="A2SChannels"></a>channels | [Channels Object](#channelsObject) | The channels used by this [application](#definitionsApplication).
```
fmvilas, line 165: Notice `channels` is not required anymore.

spec/asyncapi.md
```diff
@@ -529,167 +529,197 @@ defaultContentType: application/json
...
-<a name="channelItemObjectBindings"></a>bindings | [Channel Bindings Object](#channelBindingsObject) \| [Reference Object](#referenceObject) | A map where the keys describe the name of the protocol and the values describe protocol-specific definitions for the channel.
+<a name="channelObjectAddress"></a>address | `string` \| `null` | An optional string representation of this channel's address. The address is typically the "topic name", "routing key", "event type", or "path". When `null` or absent, it MUST be interpreted as unknown. This is useful when the address is generated dynamically at runtime or can't be known upfront. It MAY contain [Channel Address Expressions](#channelAddressExpressions).
+<a name="channelObjectAddressDelimiter"></a>addressDelimiter | `string` | An optional string to use for delimiting the address in multiple logical sections. E.g., the AMQP protocol uses the `.` (dot) character and MQTT uses the `/` (slash) character.
```
fmvilas, line 580: This is new. Instead of assuming the address delimiter is a `/` character, we make it custom. Should we define this in the bindings? I mean not making it customizable but strictly defined in the binding. For instance, the AMQP 0-9-1 binding will have a paragraph saying something like "Implementations MUST use the `.` (dot) character as a delimiter for the addresses.".

What do you think?

spec/asyncapi.md
```diff
@@ -529,167 +529,197 @@ defaultContentType: application/json
...
-<a name="channelItemObjectParameters"></a>parameters | [Parameters Object](#parametersObject) | A map of the parameters included in the channel name. It SHOULD be present only when using channels with expressions (as defined by [RFC 6570 section 2.2](https://tools.ietf.org/html/rfc6570#section-2.2)).
-<a name="channelItemObjectBindings"></a>bindings | [Channel Bindings Object](#channelBindingsObject) \| [Reference Object](#referenceObject) | A map where the keys describe the name of the protocol and the values describe protocol-specific definitions for the channel.
+<a name="channelObjectAddress"></a>address | `string` \| `null` | An optional string representation of this channel's address. The address is typically the "topic name", "routing key", "event type", or "path". When `null` or absent, it MUST be interpreted as unknown. This is useful when the address is generated dynamically at runtime or can't be known upfront. It MAY contain [Channel Address Expressions](#channelAddressExpressions).
```
fmvilas, line 579: Address is no longer a URL template and, instead, it's just a plain string that supports Channel Address Expressions.

spec/asyncapi.md
```diff
@@ -529,167 +529,197 @@ defaultContentType: application/json
...
+<a name="channelObjectAddressDelimiter"></a>addressDelimiter | `string` | An optional string to use for delimiting the address in multiple logical sections. E.g., the AMQP protocol uses the `.` (dot) character and MQTT uses the `/` (slash) character.
+<a name="channelObjectMessages"></a>messages | [Messages Object](#messagesObject) | A map of the messages that will be sent to this channel by any application at any time. **Every message sent to this channel MUST be valid against one, and only one, of the message objects defined in this map.**
```
fmvilas, line 581: No more `oneOf`. The array has the "or" meaning by default.

spec/asyncapi.md
```diff
@@ -2554,6 +2584,6 @@ string | `string` | | |
 byte | `string` | `byte` | base64 encoded characters
 binary | `string` | `binary` | any sequence of octets
 boolean | `boolean` | | |
-date | `string` | `date` | As defined by `full-date` - [RFC3339](https://xml2rfc.ietf.org/public/rfc/html/rfc3339.html#anchor14)
-dateTime | `string` | `date-time` | As defined by `date-time` - [RFC3339](https://xml2rfc.ietf.org/public/rfc/html/rfc3339.html#anchor14)
+date | `string` | `date` | As defined by `full-date` - [RFC3339](https://www.rfc-editor.org/rfc/rfc3339)
+dateTime | `string` | `date-time` | As defined by `date-time` - [RFC3339](https://www.rfc-editor.org/rfc/rfc3339)
```
fmvilas, line 2588: @derberg I fixed these two URLs and the Markdown links Action is still complaining. Any idea?

spec/asyncapi.md
```diff
@@ -529,167 +529,197 @@ defaultContentType: application/json
...
-subscribe:
-  message:
-    oneOf:
-      - $ref: '#/components/messages/signup'
-      - $ref: '#/components/messages/login'
-```
 
 
-Using explicit by-name references to the servers on which the channel is available:
+#### <a name="channelAddressExpressions"></a>Channel Address Expressions
```
fmvilas, line 670: I went all-in defining our own templating rules here. I don't know what I was thinking when I defined the channel address as a URL template in version 2.0.0 😅 Have a look: https://www.rfc-editor.org/rfc/rfc6570#section-2.4.2.

spec/asyncapi.md
```diff
@@ -529,167 +529,197 @@ defaultContentType: application/json
...
+
+Channel addresses MAY contain expressions that can be used to define dynamic values.
+
+Expressions MUST be composed by a name enclosed in curly braces (`{` and `}`). E.g., `{userId}`.
+
+When the channel address contains [delimiters](#channelObjectAddressDelimiter), the expressions MUST be surrounded by either the delimiters, the beginning of the string, or the end of the string. E.g., assuming a `.` (dot) delimiter, `users.{userId}` is correct but `users.id{userId}` is not.
```
fmvilas, line 676: This is covering most of the protocols supporting any form of wildcards: AMQP, MQTT, NATS, etc. It's still missing a Kafka feature that lets you subscribe to a regular expression. I tried to map it here but it's too complex and therefore left it out.

If you got any ideas on how to do it, I'm all ears. Otherwise, I'll open a follow-up issue and this can be included in future versions.

---

If you're curious, I came up with a solution like `users_{userId:/[\w\d]{8}/i}` but quickly got trapped in the [differences between regular expression engines](https://en.wikipedia.org/wiki/Comparison_of_regular_expression_engines) hole.

Kafka protocol doesn't define anything so it's hard for us to make a commitment here. Maybe go for the engine with fewer features, just to make sure everything is supported in every engine? Coming up with a subset ourselves? Should we just forget about it? Hot topic!

spec/asyncapi.md
```diff
@@ -529,167 +529,197 @@ defaultContentType: application/json
...
+<a name="channelObjectAddressDelimiter"></a>addressDelimiter | `string` | An optional string to use for delimiting the address in multiple logical sections. E.g., the AMQP protocol uses the `.` (dot) character and MQTT uses the `/` (slash) character.
+<a name="channelObjectMessages"></a>messages | [Messages Object](#messagesObject) | A map of the messages that will be sent to this channel by any application at any time. **Every message sent to this channel MUST be valid against one, and only one, of the message objects defined in this map.**
+<a name="channelObjectDescription"></a>description | `string` | An optional description of this channel. [CommonMark syntax](https://spec.commonmark.org/) can be used for rich text representation.
+<a name="channelObjectServers"></a>servers | [`string`] | The servers on which this channel is available, specified as an optional unordered list of names (string keys) of [Server Objects](#serverObject) defined in the [Servers Object](#serversObject) (a map). If `servers` is absent or empty then this channel must be available on all servers defined in the [Servers Object](#serversObject).
```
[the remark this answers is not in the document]
fmvilas: As a followup. I want this PR to be as much as possible about new channels object only.

spec/asyncapi.md
fmvilas: Yeah, these helpers were needed because `/` is a delimiter in URLs but that's not the case anymore so whoever is designing the AsyncAPI file needs to make sure their addresses are valid for the protocol in question.

Agree it's not giving us much, gonna remove it. Thanks!

spec/asyncapi.md
fmvilas: LOL, funny. I tried it in [this commit](https://github.com/asyncapi/spec/pull/827/commits/82a335fd61fa376980c2460c5f5cb54021bca183) and didn't work. But the funniest thing is that there's a 404 on https://xml2rfc.ietf.org/public/rfc/html/rfc3339.html#anchor14 but that's exactly the URL I replaced in this PR. It shouldn't be checking for its existence, right?

fmvilas:
@derberg I addressed all the feedback. Let me know what you think now.

spec/asyncapi.md
smoya: It seems the same issue is happening here as well: https://github.com/asyncapi/parser-js/pull/601#issuecomment-1240468666

spec/asyncapi.md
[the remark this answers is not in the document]
fmvilas: Actually, thought it better and gonna address it here. It's still about the channels object.

spec/asyncapi.md
derberg: yeah, looks like there are some issues with action that we are using,
I'll open a PR to `.github` repo to use their new pre release

in the meantime, remove your changes of links as they will be fixed with https://github.com/asyncapi/spec/pull/832

derberg (approved):
LGTM

again, info, once we merge, we release a release candidate 😉

spec/asyncapi.md
```diff
@@ -554,142 +553,147 @@ Field Pattern | Type | Description
...
 ```
 
 
 
 
-#### <a name="channelItemObject"></a>Channel Item Object
+#### <a name="channelObject"></a>Channel Object
 
-Describes the operations available on a single channel.
+Describes a shared communication channel.
```
jonaslagoni, line 568: A shared communication channel with who? Am I the only one who feels like this sentence is incomplete in some way... 🤔

spec/asyncapi.md
```diff
@@ -554,142 +553,147 @@ Field Pattern | Type | Description
...
+<a name="channelObjectMessages"></a>messages | [Messages Object](#messagesObject) | A map of the messages that will be sent to this channel by any application at any time. **Every message sent to this channel MUST be valid against one, and only one, of the [message objects](#messageObject) defined in this map.**
+<a name="channelObjectDescription"></a>description | `string` | An optional description of this channel. [CommonMark syntax](https://spec.commonmark.org/) can be used for rich text representation.
+<a name="channelObjectServers"></a>servers | [[Reference Object](#referenceObject)] | An array `$ref` pointers to the definition of the servers in which this channel is available. If `servers` is absent or empty, this channel MUST be available on all the servers defined in the [Servers Object](#serversObject). Please note the `servers` property value MUST be an array of [Reference Objects](#referenceObject) and, therefore, MUST NOT contain an array of [Server Objects](#serverObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience.
```
jonaslagoni, line 577:
```suggestion
<a name="channelObjectServers"></a>servers | [[Reference Object](#referenceObject)] | An array of `$ref` pointers to the definition of the servers in which this channel is available. If `servers` is absent or empty, this channel MUST be available on all the servers defined in the [Servers Object](#serversObject). Please note the `servers` property value MUST be an array of [Reference Objects](#referenceObject) and, therefore, MUST NOT contain an array of [Server Objects](#serverObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience.
```

jonaslagoni (commented):
Just had a few remarks, otherwise looks good 👍

dalelane (approved):
LGTM - I particularly like moving the topic name out of the spec document key, and into an address field

spec/asyncapi.md
fmvilas: We don't know in advance. We only know it's gonna be shared 🤷

spec/asyncapi.md
fmvilas: Good catch 🦅 👁️

fmvilas:
😱 😄

fmvilas:
@derberg It didn't create any release candidate. Is it expected?

derberg:
no, mainly because the fix I did in release pipeline did not go to `next-major-spec` branch as there are merge conflicts between it and `master`. It is good though, as PR title should be `feat!:` to get 3.0 RC and not 2.5 RC. So by accident we did not release a confusing release 😄

PR #834, opened 2022-09-15
ci: fix release workflow to support release branches

derberg:
when we changed release branches, we forgot to update the workflow that reacts to them, we forgot to change the names of branches in the config

I intentionally did not make a change like `'**-spec'` to have branches named explicitly

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,7 +4,8 @@ on:
   push:
     branches:
       - master
-      - '**-release'
+      - next-spec
+      - next-major-spec
  
 jobs:
   release:


derberg:
@fmvilas @dalelane I need quick approval as we are now not able to produce release candidates for 2.5.0 as we should

.github/workflows/release.yml
```diff
@@ -4,7 +4,8 @@ on:
   push:
     branches:
       - master
-      - '**-release'
+      - next-spec
+      - next-major-spec
```
magicmatatjahu, line 8: In parser we have also `next-major` and in react-component in modelina `next`

smoya (approved):

char0n:
Right, I've just bumped into this issue as well. @fmvilas @dalelane can you approve please?

@derberg would you able to trigger this workflow manually after the change is merged?

derberg:
@char0n manual run is not available, and rerun will probably not pick this change. But maybe we will be able to merge googepubsub binding soon, otherwise, we will just have to do an empty PR to trigger release 🤷🏼

derberg:
@fmvilas @dalelane one of you folks need to approve this one 🙏🏼

fmvilas (approved):

derberg:

PR #839, opened 2022-09-22
ci: fix release workflow to support release branches

char0n:


derberg (approved):

derberg:
I think after the merge, we will actually get a release candidate from the old commit as semantic release always analyzes all commits from the past, if no tag was created 🤞🏼 

/rtm

PR #840, opened 2022-09-22
fix: add missed IBM MQ Operation Binding

magicmatatjahu:
I found that we have forgot to define (probably it's a bug) definition for `IBM MQ Operation Binding`. I know that some bindings don't need to have definition for particular kind of binding, but we should have reserved keywords.

I found it in review of https://github.com/asyncapi/spec/pull/836 PR 

cc @derberg @fmvilas and @dalelane you should be most interested :)

PR #836, opened 2022-09-19
feat: add googlepubsub bindings

whitlockjc:
Adds support for Google Cloud Pub/Sub to the `Channel Binding Object` and `Message Binding Object` lists.

**Related issue(s):**
Bindings PR: https://github.com/asyncapi/bindings/pull/141
JSON Schema PR: https://github.com/asyncapi/spec-json-schemas/pull/253

Issue #830, opened 2022-09-05
Work on 2.5.0 release


char0n:
Release 2.5.0 is scheduled for September 2022

### Detailed info:

* https://github.com/asyncapi/spec/blob/master/RELEASE_PROCESS.md

Assigned release coordinator and person responsible for the 2.5.0 release: @char0n

### Kick-off:

* **TODO**: incorporate release plan described by this issue in one of the September community meetings

### Release branches:

* https://github.com/asyncapi/spec/tree/next-spec
* https://github.com/asyncapi/spec-json-schemas/tree/next-spec
* https://github.com/asyncapi/parser-js/tree/next-spec
* https://github.com/asyncapi/bindings/tree/master

### Release notes:

* draft PR - https://github.com/asyncapi/website/pull/PULLREQUEST

#### Accepted

- [x] https://github.com/asyncapi/spec/issues/775
  - [x] https://github.com/asyncapi/spec/pull/776
- [x] https://github.com/asyncapi/spec/issues/654
  - [x] https://github.com/asyncapi/spec/pull/809
  - [x] https://github.com/asyncapi/spec-json-schemas/pull/232
  - [x] https://github.com/asyncapi/parser-js/pull/565
- [x] Missing IBMMQ Operation Binding
  - [x] https://github.com/asyncapi/spec/pull/840
  - [x] https://github.com/asyncapi/bindings/pull/149
- [x] Google Pub/Sub bindings
  - [x] https://github.com/asyncapi/bindings/pull/141
  - [x] https://github.com/asyncapi/spec/pull/836
  - [x] https://github.com/asyncapi/spec-json-schemas/pull/253
- [x] https://github.com/asyncapi/spec/issues/612 (editorial change)
  - [x] https://github.com/asyncapi/spec/pull/779
- [x] https://github.com/asyncapi/spec/issues/674 (editorial change)
  - [x] https://github.com/asyncapi/spec/pull/782
- [x] https://github.com/asyncapi/spec/pull/818 (editorial change)
- [x] https://github.com/asyncapi/spec/pull/832 (editorial change)
- [x] https://github.com/asyncapi/bindings/pull/115

#### Discarded

- [x] https://github.com/asyncapi/bindings/pull/143
- [x] https://github.com/asyncapi/spec/pull/826
- [x] https://github.com/asyncapi/spec/pull/595
- [x] https://github.com/asyncapi/spec/pull/790

### Progress:
- [x] Update release branches from forks
[...]

PR #818, opened 2022-07-05
docs: align wording with RFC 2119

char0n:
This is an editorial change, so sending against `master` branch.

Refs https://github.com/swagger-api/apidom/issues/1663

More on RFC 2119 -> https://www.rfc-editor.org/rfc/rfc2119

PR #842, opened 2022-09-28
feat: release for version 2.5.0 of the spec

### LINKED ISSUE #612: Channel Item Object `$ref` field resolution rules

### LINKED ISSUE #654: Allow grouping servers

### LINKED ISSUE #674: Clarification on URL and it's resolution rules

### LINKED ISSUE #775: Extending `Server.variables` field type

char0n:
Refs https://github.com/asyncapi/spec/issues/830

Features
---
- Server Variables can be now referenced from inside the Server Object (#776) ([5deef1e](https://github.com/asyncapi/spec/commit/5deef1e9148947689d53db3a1d013d49cd9df7d0)), closes #775 by @char0n 
- Decorating Server Object with tags (#809) ([b0c2f9c](https://github.com/asyncapi/spec/commit/b0c2f9c086d940e8016ec1fa60107fe74cc070db)), closes #654 by @smoya
- Added new Google Cloud Pub/Sub Bindings (#836) ([babc3be](https://github.com/asyncapi/spec/commit/babc3bedb5d85375a305c7268e96257113c1933e)) ([binding definition](https://github.com/asyncapi/bindings/tree/master/googlepubsub)) by @whitlockjc 
- Other enhancements
  - Clarified ChannelItem.$ref resolution rules (#779) ([9a293c7](https://github.com/asyncapi/spec/commit/9a293c77f6bf74a056a8845dabfbe20e73a6f320)), closes #612 by @char0n
  - Clarification around URL resolution rules (#782) ([fc324ef](https://github.com/asyncapi/spec/commit/fc324ef3aa92d313918a00ea9a5eb6c85720a36d)), closes #674 by @char0n
  - Align wording with [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119) (#818) ([5580a9c](https://github.com/asyncapi/spec/commit/5580a9c5fbef55e925e7514b986c95a0e4c85821)) by @char0n

Fixes
---
- Added missing IBM MQ Operation Binding (#840) ([68d85d1](https://github.com/asyncapi/spec/commit/68d85d19e5be12fa6e13d0fc5851993934dbdcde)) ([binding definition](https://github.com/asyncapi/bindings/tree/master/ibmmq#operation-binding-object)), refs [#605](https://github.com/asyncapi/spec/issues/605) by @magicmatatjahu

Resolves https://github.com/asyncapi/spec/issues/775
Resolves https://github.com/asyncapi/spec/issues/465
Resolves https://github.com/asyncapi/spec/issues/654


char0n:
Here are the release notes for the 2.5.0 version of the spec that should be used for the GitHub Release.

```md
Release notes article -> https://www.asyncapi.com/blog/release-notes-2.5.0

Features
---
- Server Variables can be now referenced from inside the Server Object (#776) ([5deef1e](https://github.com/asyncapi/spec/commit/5deef1e9148947689d53db3a1d013d49cd9df7d0)), closes #775 by @char0n 
- Decorating Server Object with tags (#809) ([b0c2f9c](https://github.com/asyncapi/spec/commit/b0c2f9c086d940e8016ec1fa60107fe74cc070db)), closes #654 by @smoya
- Added new Google Cloud Pub/Sub Bindings (#836) ([babc3be](https://github.com/asyncapi/spec/commit/babc3bedb5d85375a305c7268e96257113c1933e)) ([binding definition](https://github.com/asyncapi/bindings/tree/master/googlepubsub)) by @whitlockjc 
- Other enhancements
  - Clarified ChannelItem.$ref resolution rules (#779) ([9a293c7](https://github.com/asyncapi/spec/commit/9a293c77f6bf74a056a8845dabfbe20e73a6f320)), closes #612 by @char0n
  - Clarification around URL resolution rules (#782) ([fc324ef](https://github.com/asyncapi/spec/commit/fc324ef3aa92d313918a00ea9a5eb6c85720a36d)), closes #674 by @char0n
  - Align wording with [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119) (#818) ([5580a9c](https://github.com/asyncapi/spec/commit/5580a9c5fbef55e925e7514b986c95a0e4c85821)) by @char0n

Fixes
---
- Added missing IBM MQ Operation Binding (#840) ([68d85d1](https://github.com/asyncapi/spec/commit/68d85d19e5be12fa6e13d0fc5851993934dbdcde)) ([binding definition](https://github.com/asyncapi/bindings/tree/master/ibmmq#operation-binding-object)), refs [#605](https://github.com/asyncapi/spec/issues/605) by @magicmatatjahu

Full Changelog: https://github.com/asyncapi/spec/compare/v2.4.0...v2.5.0
```

derberg (approved):
LGTM 

@char0n thanks so much man for your support and all the efforts to get shit done 🍻

@fmvilas @dalelane please make final approval for the release

smoya (approved):

fmvilas (approved):
👏 🍺  Great stuff!

dalelane (approved):
🎉

char0n:
@fmvilas @smoya @derberg @delelane this is good to go. According to https://github.com/asyncapi/spec/blob/master/RELEASE_PROCESS.md#step-13---merge-the-release-branches, one of the code owner should merge this PR. Can I ask one of you to merge this PR?

Updated release notes to be used for GitHub release are in: https://github.com/asyncapi/spec/pull/842#issuecomment-1260785616

Thank you!

derberg:
https://github.com/asyncapi/spec/releases/tag/v2.5.0 is out

derberg:

Issue #829, opened 2022-09-01
Unify all referencing mechanisms in v3


fmvilas:
Spin off from this conversation: https://github.com/asyncapi/spec/issues/663#issuecomment-1158094050.

## Summary

We use different referencing mechanisms in the spec:

* By name in components: SecurityRequirement Object
* By name in servers: Channel.servers property
* By using `$ref`: pretty much everywhere

This issue is to discuss if we should unify all of them and how.

fmvilas:
I'm in favor of using `$ref` everywhere possible. More details in https://github.com/asyncapi/spec/issues/663#issuecomment-1158094050.

GreenRover:
I agree.  Having it unique would be a good idea. 
`$ref` looks for me as the best solution as well.

#### Contra:
- `$ref` would allow illegal references
```yaml
operations:
  sendUserSignedUp:
    action: send
    channel:
      $ref: '#/message/xyz' <--- should be /channel/ not /message/
```
But there for we have schema validator to point out mistakes like this.

#### Pro:
- It is unique we not need to be use to handle multiple ways to solve the same problem
- It is supported by most IDE. Your have jump marks and mouse over support
- Be more like brother project OpenAPI

derberg:
let's do `$ref` but consistently everywhere

fmvilas:
/progress 60

PR is in place now: https://github.com/asyncapi/spec/pull/852.

jonaslagoni:
I cannot seem to find the changes in the parser, @fmvilas are you sure this change is complete?

I have the changes in for the spec https://github.com/asyncapi/spec/pull/852 and JSON Schema files https://github.com/asyncapi/spec-json-schemas/pull/316.

fmvilas:
Yes, it's missing. I created an issue here: https://github.com/asyncapi/parser-js/issues/746. Thanks for the pointer!

Issue #828, opened 2022-08-30
Clarify the shape of the SecurityRequirement Object


magicmatatjahu:
Together with @smoya we wonder if we have a good understanding of the use of the `SecurityRequirements Object` item in the `Server Object` and `Operation Object`. The specification gives the shape for `SecurityRequirement Object` as:

![image](https://user-images.githubusercontent.com/20404945/187448417-a3d681c3-1d76-468e-827f-4c1f34968d3d.png)

Then, according to the definition, are we able to use two `SecuritySchemas` in one `SecurityRequirement Object` like:

```yaml
security:
  - apiKeyScheme: []
    oauthScheme: ['...']
  - openIDScheme: []
```

?

It seems to me that we should describe it in a better way and if it is possible to use two schemas for one `SecurityRequirement Object` then we should also give an example using two schemas. If not then it still requires a precise description that only one schema is allowed for one `SecurityRequirement Object`.

cc @derberg @fmvilas @dalelane

fmvilas:
That's a good question. I got this from OpenAPI and [it doesn't clarify it either](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#securityRequirementObject). My gut feeling is that it's possible and it's fine. So the following two examples should be equivalent:

```yaml
security:
  - apiKeyScheme: []
    oauthScheme: ['...']
```

```yaml
security:
  - apiKeyScheme: []
  - oauthScheme: ['...']
```

magicmatatjahu:
@fmvilas Thanks for fast response!

> So the following two examples should be equivalent:

Hmm, I don't think so 😅 As we read description for single `SecurityRequirement Object`:

> When a list of Security Requirement Objects is defined on a [Server object](https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md#serverObject), only one of the Security Requirement Objects in the list needs to be satisfied to authorize the connection.

so when you have two schemas like:

```yaml
- apiKeyScheme: []
  oauthScheme: ['...']
```

You need to satisfy to authorize the connection these two schemas, both `apiKeyScheme` AND `oauthScheme`. In the case:

```yaml
- apiKeyScheme: []
- oauthScheme: ['...']
```

we need to satisfy `apiKeyScheme` OR `oauthScheme`. Of course I can be wrong, but for me array !== object.

fmvilas:
Oh yeah, you're right. So there you go. If it's an object, it's an AND operation, otherwise, it's an OR.

[...]

Issue #778, opened 2022-05-04
[3.0.0] Allow references to be used in any part of the specification


magicmatatjahu:
The current version of AsyncAPI (2.x.x) has strictly defined places where users can use references in a document. We should extend the available places to every part of AsyncAPI.

Benefits:
- no PR and issue like https://github.com/asyncapi/spec/issues/775 (after every release we have at least 1 because we forgot to update places where we can use references)
 - easier support in tooling - currently in our official tooling we don't check where references were used, but I don't see the benefits of such checking
- people wouldn't be confused if a reference is allowed or not

It seems to me that instead of giving the type as the union of the object and reference in each place, we could add a section in the spec explaining how to use references (at the beginning) and they are allowed to be use everywhere.

This is a breaking change, so I'm assigning this issue for 3.0.0.

char0n:
HI @magicmatatjahu,

Very quickly couple of points I can think of:

---

The `Reference Object` comes originally from OpenAPI spec and as far as I understand is was introduced to have full control over how definitions written using the spec look like (structure) and which parts do make sense to be reusable and which not.

---

This proposal will IMHO introduce problematic ambiguity. We'll also have problem recognizing definition author intentions. How do we tell if `$ref` was meant as a arbitrary field name or as a `Reference Object`?

```js
{
  "components": {
    "schemas": {
      "$ref": { }
    }
  }
}
```
---

There will be problem with identifying resolution rules. Every version of the spec can have specific resolution rules. **2.4** uses [JSON Reference](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03), **3.0** can use [RFC3986](https://tools.ietf.org/html/rfc3986#section-5.2).

Having following definition:

```yaml
asyncapi: 
  $ref: 'external-document.yaml#/path/to/version',
info:
  title: Streetlights Kafka API
  version: '1.0.0'
  description: |
    The Smartylighting Streetlights API allows you to remotely manage the city lights.
```

In order to know what are the resolution rules for the `Reference Object`, I have to determine what is the version of the AsyncAPI spec used to write this definition. I cannot resolve as I don't know the spec version and and I cannot get the spec version as I don't know the resolution rules. We got into classical  `chicken and egg situation` which cannot be resolved.

magicmatatjahu:
@char0n Thanks for comment.

[...]

PR #852, opened 2022-10-13
feat: unify all referencing mechanisms

fmvilas:
### Description
This PR unifies all the referencing mechanisms in the spec. So far, only the `security` requirements object had to be changed.

### Related issues
#829

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -75,7 +75,6 @@ It means that the [application](#definitionsApplication) allows [consumers](#def
       - [Reference Object](#referenceObject)
       - [Schema Object](#schemaObject)
       - [Security Scheme Object](#securitySchemeObject)
-      - [Security Requirement Object](#security-requirement-object)
       - [OAuth Flows Object](#oauth-flows-object)  
       - [OAuth Flow Object](#oauth-flow-object)
       - [Server Bindings Object](#serverBindingsObject)
@@ -342,7 +341,7 @@ Field Pattern | Type | Description
 
 ```json
 {
-  "production": {
+  "development": {
     "url": "development.gigantic-server.com",
     "description": "Development server",
     "protocol": "kafka",
@@ -352,7 +351,7 @@ Field Pattern | Type | Description
 ```
 
 ```yaml
-production:
+development:
   url: development.gigantic-server.com
   description: Development server
   protocol: kafka
@@ -373,7 +372,7 @@ Field Name | Type | Description
 <a name="serverObjectProtocolVersion"></a>protocolVersion | `string` | The version of the protocol used for connection. For instance: AMQP `0.9.1`, HTTP `2.0`, Kafka `1.0.0`, etc.
 <a name="serverObjectDescription"></a>description | `string` | An optional string describing the host designated by the URL. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
 <a name="serverObjectVariables"></a>variables | Map[`string`, [Server Variable Object](#serverVariableObject) \| [Reference Object](#referenceObject)]] | A map between a variable name and its value.  The value is used for substitution in the server's URL template.
-<a name="serverObjectSecurity"></a>security | [[Security Requirement Object](#securityRequirementObject)] | A declaration of which security mechanisms can be used with this server. The list of values includes alternative security requirement objects that can be used. Only one of the security requirement objects need to be satisfied to authorize a connection or operation.
+<a name="serverObjectSecurity"></a>security | [[Security Scheme Object](#securitySchemeObject) \| [Reference Object](#referenceObject)] | A declaration of which security schemes can be used with this server. The list of values includes alternative [security scheme objects](#securitySchemeObject) that can be used. Only one of the security scheme objects need to be satisfied to authorize a connection or operation.
 <a name="serverObjectTags"></a>tags | [Tags Object](#tagsObject) | A list of tags for logical grouping and categorization of servers.
 <a name="serverObjectBindings"></a>bindings | [Server Bindings Object](#serverBindingsObject) \| [Reference Object](#referenceObject) | A map where the keys describe the name of the protocol and the values describe protocol-specific definitions for the server.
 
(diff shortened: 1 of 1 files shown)


spec/asyncapi.md
```diff
@@ -373,7 +372,7 @@ Field Name | Type | Description
...
 <a name="serverObjectVariables"></a>variables | Map[`string`, [Server Variable Object](#serverVariableObject) \| [Reference Object](#referenceObject)]] | A map between a variable name and its value.  The value is used for substitution in the server's URL template.
-<a name="serverObjectSecurity"></a>security | [[Security Requirement Object](#securityRequirementObject)] | A declaration of which security mechanisms can be used with this server. The list of values includes alternative security requirement objects that can be used. Only one of the security requirement objects need to be satisfied to authorize a connection or operation.
+<a name="serverObjectSecurity"></a>security | [[Security Scheme Object](#securitySchemeObject) \| [Reference Object](#referenceObject)] | A declaration of which security schemes can be used with this server. The list of values includes alternative [security scheme objects](#securitySchemeObject) that can be used. Only one of the security scheme objects need to be satisfied to authorize a connection or operation.
```
fmvilas, line 375: Instead of having a weird syntax like:

```yaml
security:
  - httpBearerToken: []
```

My proposal is that we do the following:

```yaml
...
  security:
    - $ref: '#/components/securitySchemes/httpBearerToken'
...  
components:
  securitySchemes:
    httpBearerToken:
      type: http
      scheme: bearer
```

The array of the current syntax is just for OAuth2 and OpenIDConnect schemes so I created a new property called `scopes` in the Security Scheme Object. These will be the mandatory scopes needed to connect or perform an operation. It offers less reusability in the `security` property but you can anyway create as many Security Scheme Objects inside components so, yeah, not an issue.

spec/asyncapi.md
```diff
@@ -399,128 +398,82 @@ protocol: kafka
...
-      "description": "Development server",
-      "protocol": "amqp",
-      "protocolVersion": "0.9.1",
-      "tags": [
-        { 
-          "name": "env:development",
-          "description": "This environment is meant for developers to run their own tests"
-        }
-      ]
+  "url": "{username}.gigantic-server.com:{port}/{basePath}",
```
fmvilas, line 405: Here I'm just removing context information in the example. Just to keep it the same as in the rest of the spec.

spec/asyncapi.md
```diff
@@ -2442,106 +2401,32 @@ Field Name | Type | Applies To | Description
 <a name="oauthFlowAuthorizationUrl"></a>authorizationUrl | `string` | `oauth2` (`"implicit"`, `"authorizationCode"`) | **REQUIRED**. The authorization URL to be used for this flow. This MUST be in the form of an absolute URL.
 <a name="oauthFlowTokenUrl"></a>tokenUrl | `string` | `oauth2` (`"password"`, `"clientCredentials"`, `"authorizationCode"`) | **REQUIRED**. The token URL to be used for this flow. This MUST be in the form of an absolute URL.
 <a name="oauthFlowRefreshUrl"></a>refreshUrl | `string` | `oauth2` | The URL to be used for obtaining refresh tokens. This MUST be in the form of an absolute URL.
-<a name="oauthFlowScopes"></a>scopes | Map[`string`, `string`] | `oauth2` | **REQUIRED**. The available scopes for the OAuth2 security scheme. A map between the scope name and a short description for it.
+<a name="oauthFlowScopes"></a>availableScopes | Map[`string`, `string`] | `oauth2` | **REQUIRED**. The available scopes for the OAuth2 security scheme. A map between the scope name and a short description for it.
```
fmvilas, line 2404: Renaming the `scopes` property of the OAuthFlow Object to `availableScopes` to differentiate from the other one added in this PR.

spec/asyncapi.md
jonaslagoni: I like it!

magicmatatjahu: Consider that in the security field we have 3 options:
- defining one security (as in your case)
- defining several security that must be met at once:
  
  ```yaml
  ...
  security: 
    - - $ref: '#/components/securitySchemes/1schema'
      - $ref: '#/components/securitySchemes/2schema'
  ```

- defining several security, for which only one security must be satisfied. 

  ```yaml
  ...
  security: 
    - $ref: '#/components/securitySchemes/1schema'
    - $ref: '#/components/securitySchemes/2schema'
  ```
  
So as you can see, we have problem with second case (several securities have to met at once). We have even issue for that https://github.com/asyncapi/spec/issues/828

magicmatatjahu: > The array of the current syntax is just for OAuth2 and OpenIDConnect schemes so I created a new property called scopes in the Security Scheme Object. These will be the mandatory scopes needed to connect or perform an operation. It offers less reusability in the security property but you can anyway create as many Security Scheme Objects inside components so, yeah, not an issue.

Why not to create security item shape as:

```yaml
security: 
  - scheme:
       $ref: '#/components/securitySchemes/1schema'
    scopes: [...]
```

?

fmvilas: > Why not to create security item shape as:
> 
> ```yaml
> security: 
>   - scheme:
>        $ref: '#/components/securitySchemes/1schema'
>     scopes: [...]
> ```
> 
> ?

Why should we? :) The proposed syntax looks much cleaner IMHO. It only has the problem of meeting multiple security schemes at once but I'm not sure it should be a requirement. Left a comment at https://github.com/asyncapi/spec/issues/828#issuecomment-1280635558.

fmvilas: I understand your concern but, honestly, I don't think most people would have dozens of OAuth schemes with different scopes, just probably a couple or three. In any case, you can still reuse it this way, right?

```yaml
...
  security:
    - $ref: '#/components/securitySchemes/oauth2UserRead'
...  
components:
  securitySchemes:
    oauth2base:
      type: oauth2
      description: 'blah blah'
    oauth2UserRead:
      $ref: '#/components/securitySchemes/oauth2base'
      scopes: ['user-read']
    oauth2UserWrite:
      $ref: '#/components/securitySchemes/oauth2base'
      scopes: ['user-write']
```

Or are we planning not to support merging an object when using `$ref` alongside other properties?

magicmatatjahu: > Or are we planning not to support merging an object when using $ref alongside other properties?

As I remember, we have issue for that (or mention about it in another issue, I don't know), but I'm not involving with it. cc @smoya

jonaslagoni: > As I remember, we have issue for that (or mention about it in another issue, I don't know), but I'm not involving with it. cc @smoya

@magicmatatjahu you raised it as thoughts here: https://github.com/asyncapi/spec/issues/778#issuecomment-1117128502

I don't think it has been discussed anywhere else 🧐 

> Or are we planning not to support merging an object when using $ref alongside other properties?

It's not that we are planning not to support it, but we have not planned or had a discussion about whether that should be a feature I think 🤔

jonaslagoni: >And tooling is already behaving this way, I think.

That is only because the tooling has bugs, it is not intended behavior... 🙂 

Adding this functionality will definitely complicate it from a spec perspective, but yea simplify it in this case 😆

fmvilas: Bug or feature? 😛 To be honest, this is a thing I see many people doing/trying expecting it to work. But yeah, that's not a discussion for this thread sorry for the noise.

spec/asyncapi.md
```diff
@@ -2278,6 +2231,7 @@ Field Name | Type | Applies To | Description
 <a name="securitySchemeObjectBearerFormat"></a>bearerFormat | `string` | `http` (`"bearer"`) | A hint to the client to identify how the bearer token is formatted.  Bearer tokens are usually generated by an authorization server, so this information is primarily for documentation purposes.
 <a name="securitySchemeFlows"></a>flows | [OAuth Flows Object](#oauthFlowsObject) | `oauth2` | **REQUIRED**. An object containing configuration information for the flow types supported.
 <a name="securitySchemeOpenIdConnectUrl"></a>openIdConnectUrl | `string` | `openIdConnect` | **REQUIRED**. OpenId Connect URL to discover OAuth2 configuration values. This MUST be in the form of an absolute URL.
+<a name="securitySchemeScopes"></a>scopes | [`string`] | `oauth2` \| `openIdConnect` | List of the needed scope names. An empty array means no scopes are needed.
```
fmvilas, line 2234: Thinking about adding the `encryption` key to the Security Scheme Object and removing it from the `type` field. Encryption mechanisms are compatible with all the types listed in `type` after all. Maybe the only one to be left out is `plain`? @derberg @dalelane @char0n @smoya @magicmatatjahu @jonaslagoni

dalelane: I'm not sure I understand what this would end up like...

Do you mean we would have, for example, `encryption: scramSha512`? 

If so, then I'm not sure I would like that, because I don't think that fits well - as to me SCRAM is more about describing the security mechanism (the type of challenge-response that the client has to do with the server) than just describing the SHA encryption bit. 

If not, then I've misunderstood :)

fmvilas: Sorry for the poor explanation. No, it would be `encryption: symmetric or `encryption: asymmetric. We now have [`type: symmetricEncryption` and `type: assymetricEncryption`](https://github.com/asyncapi/spec/blob/2a7acb2e2503381648c0ff2caa97588f50edb974/spec/asyncapi.md#fixed-fields-22). They both mean that the message content is encrypted, even if the connection is not. I'm considering moving this to a property because it's not an auth type per se. Actually, you can send an encrypted message over any kind of connection.

I'm actually wondering now if this is being used by anyone 😅 It's not very useful because you only get to know that something is encrypted but it doesn't tell you how it's encrypted so, yeah 🤷  

Anyway, this is not really relevant to this PR but just came to my mind and had to share.

fmvilas: Thinking further, I think I'm gonna remove encryption types from 3.0.0 and see if someone complains 😄 If they do, we can add them back on 3.1.0 haha! If not, it means nobody cares.

fmvilas:
@derberg @dalelane @char0n @magicmatatjahu @jonaslagoni would you mind having a look and approving if you like it? I want to start getting rid of this small PRs.

dalelane (approved):

spec/asyncapi.md
handrews: I strongly encourage folks to file thoughts around potential reference merging in AsyncAPI as issues at the [unified referencing discussion repo](https://github.com/json-schema-org/referencing), where we can consider them alongside, for example, the limited merging of `summary` and `description` that OpenAPI 3.1 does outside of JSON Schema references.

This wouldn't mean AsyncAPI is blocked from taking other action, but unified referencing will only be unified if we know what's going on with each project.

handrews (commented):
This consistency is great to see!

Is `discriminator` being considered as well?  It uses property values as implicit references.  I believe OpenAPI 4 is likely to deprecate it in favor of JSON Schema's forthcoming `propertyDependencies` (which uses `$ref` normally).  So that would be one option (even with AsyncAPI 3.0 staying on draft-07, you could import that keyword as an extension in place of `discriminator`).

BTW, I am not demanding that it be addressed in this PR or this release, I'm just pointing out that it's an implicit reference as currently defined.  Fixing security scheme is definitely higher-value!

derberg (commented):
lgtm, just resolve conflicts and I'm ready to merge

fmvilas:
👍 Done!

derberg (approved):

PR #854, opened 2022-10-19
docs: add info about breaking changes to contributing guide

### LINKED ISSUE #688: What do we define as a breaking change?

Dule-martins:
I extended the CONTRIBUTING.md file with content for Breaking Change vs Non Breaking Change.
This PR Resolves #688


CONTRIBUTING.md
```diff
@@ -78,6 +78,18 @@ move forward.
   The AsyncAPI spec, despite describing technical behavior, is intended to be
   read by people. Use natural tone and include motivation and examples.
 
+## Breaking Change vs Non Breaking Change
+
+The specification is not much different than software. Some changes provided in the spec can cause breaking changes for tools that support it. For example, if one of the properties that was `required` becomes `optional`, it is considered a breaking change because some tools might depend on that property and fail if all suddenly it is not provided.
```
fmvilas, line 83:
```suggestion
Specifications are not much different than software. Some changes provided in the spec can cause breaking changes for tools that support it. For example, if one of the properties that was `required` becomes `optional`, it is considered a breaking change because some tools might depend on that property and fail if all suddenly it is not provided. 
```

CONTRIBUTING.md
```diff
@@ -78,6 +78,18 @@ move forward.
...
+The specification is not much different than software. Some changes provided in the spec can cause breaking changes for tools that support it. For example, if one of the properties that was `required` becomes `optional`, it is considered a breaking change because some tools might depend on that property and fail if all suddenly it is not provided. 
+
+_Non-breaking changes:_
+- Adding a property
+
+_Breaking changes:_
+- Making a property required
+- Making a property optional
+- Removing a property
+- Changing the type of a property in any way
```
fmvilas, line 92: Is it intended to be an exhaustive list or just an example?

fmvilas (changes requested):

CONTRIBUTING.md
Dule-martins: from the suggestion made I think it is a list, I can't be sure of how exhaustive it is nor if it is an example.
Maybe @derberg could explain that.

derberg: it is intended as a result of https://github.com/asyncapi/spec/issues/688

it should be explicit

fmvilas (dismissed):

derberg:
@dalelane @char0n @smoya  as this is a change to contribution guide, I think the rule `to have 3 code owners approval` applies here. 2 more are missing

smoya (dismissed):
LGTM! 🚀🌔

CONTRIBUTING.md
```diff
@@ -78,6 +78,18 @@ move forward.
...
+Specifications are not much different than software. Some changes provided in the spec can cause breaking changes for tools that support it. For example, if one of the properties that was `required` becomes `optional`, it is considered a breaking change because some tools might depend on that property and fail if all suddenly it is not provided. 
+
+_Non-breaking changes:_
+- Adding a property
+
+_Breaking changes:_
+- Making a property required
+- Making a property optional
+- Removing a property
+- Changing the type of a property in any way
```
char0n, line 92: IMHO this sentence is not precise. I'm suggesting following one:

```suggestion
- Changing the type of a property in backward compatible way
```
**Justification:**: There are multiple ways the field type can change. AsyncAPI 2.x.y is using **UNION** structural typing, thus there are two situations that are similar, but in context of backward compatibility, different.

**Case 1 - backward compatible**

Before Type: `Server Variable Object`
After Type: `Server Variable Object | Reference Object`

**Case 2 - backward incompatible**

Before Type: `Server Variable Object | Reference Object`
After Type: `Server Variable Object`

There is also another case which is variant of `Case 1` and backward compatible as long as `Server Variable Object` and `Difference Object` have the same shape (in context of structural typing). This case can be simplified by calling it a **type renaming**.

**Case 1 (variant 1) - backward compatible**

Before Type: `Server Variable Object`
After Type: `Different Object`

char0n (changes requested):
Proposing a suggestion about type backward compatibility.

CONTRIBUTING.md
fmvilas: Then your suggestion should say "backward incompatible" way, right? 🤔

char0n: @fmvilas right ;] Thanks, updated the suggestion.

CONTRIBUTING.md
```diff
@@ -78,6 +78,18 @@ move forward.
   The AsyncAPI spec, despite describing technical behavior, is intended to be
   read by people. Use natural tone and include motivation and examples.
 
+## Breaking Change vs Non Breaking Change
+
+Specifications are not much different than software. Some changes provided in the spec can cause breaking changes for tools that support it. For example, if one of the properties that was `required` becomes `optional`, it is considered a breaking change because some tools might depend on that property and fail if all suddenly it is not provided. 
+
+_Non-breaking changes:_
+- Adding a property
```
dalelane, line 86: should this be adding an optional property?

fmvilas: Yes, definitely. If you add a required property, it's a breaking change.

derberg: @dalelane makes sense, can you make a code change suggestion for @Dule-martins please?

dalelane, line 86:
```suggestion
- Adding a new optional property
```

CONTRIBUTING.md
```diff
@@ -78,6 +78,18 @@ move forward.
...
+
+_Breaking changes:_
```
dalelane, line 88:
```suggestion
_Breaking changes:_
- Adding a new required property
```

CONTRIBUTING.md
Dule-martins: @derberg I just committed all suggestions made here.

smoya (dismissed):

Dule-martins (commented):
Hi, @fmvilas do you think something here is undone? Can you let me what that is so I could work on it?

CONTRIBUTING.md
```diff
@@ -78,6 +78,19 @@ move forward.
   The AsyncAPI spec, despite describing technical behavior, is intended to be
   read by people. Use natural tone and include motivation and examples.
 
+## Breaking Change vs Non Breaking Change
+
+Specifications are not much different than software. Some changes provided in the spec can cause breaking changes for tools that support it. For example, if one of the properties that was `required` becomes `optional`, it is considered a breaking change because some tools might depend on that property and fail if all suddenly it is not provided.
```
dalelane, line 83:
```suggestion
Specifications are not much different than software. Some changes provided in the spec can cause breaking changes for tools that support it. For example, if one of the properties that was `required` becomes `optional`, it is considered a breaking change because some tools might depend on that property and fail if it is not provided. 
```

suggested improvement for readability - I don't think it changes the intended meaning

dalelane (dismissed):
I've left a minor suggested wording improvement, but essentially this looks good to me

derberg:
we need to specify something super important about `breaking change` and understanding of `backwards incompatibility`

discussed scenario: 
- I have AsyncAPI document in version 2.4 
- It has  `security` field under operation (not supported in previous versions)
- I switch my version of AsyncAPI document to 2.3

expected result:
- as adding new property is not a breaking change, it is expected that it is a valid AsyncAPI 2.3 document
- in other words, parser written for AsyncAPI 2.3 will still successfully parse such AsyncAPI document with `security` in operation

current result:
- tools fail because JSON Schema for AsyncAPI says `additionalProperties: false`
- specification says explicitly that `additional data can be added to extend the specification` but as `x-` extension

---

This basically means that `new optional filed` is also a breaking changes because it is not backward compatible

more info on 3.0 call -> https://youtu.be/hN6aE3Ebn08?t=1105 (specific moment pointed)

thoughts?

GreenRover:
Recap of today spec 3.0 meeting:
- backward compatibility is not a goal: 
  if the parser is cape able to handle max spec version 2.7 it will not be able to handle 2.8.
  Of the viewpoint tooling every minor spec increase is breaking
- a possible way to define breaking change is:
  you take a spec and change the version number to a newer one and done have to modify something else. The spec is not breaking.

CONTRIBUTING.md
Dule-martins: Yes

dalelane (approved):

GreenRover (approved):

fmvilas (approved):

derberg (approved):

derberg:
@char0n we are not able to merge without your approval as in the past you requested changes

char0n (approved):
LGTM

char0n:
@derberg approval provided

smoya (approved):

PR #858, opened 2022-10-25
feat: add server and channel traits objects

magicmatatjahu:
---
title: "Add server and channel traits objects"
---

---

**Related issue(s):**
https://github.com/asyncapi/spec/issues/618

---

That PR adds `Server Trait Object` and `Channel Trait Object` to be able to create generic channels and servers (e.g. for registries) and extend them. The idea of `Channel Trait` was conceived while writing the https://github.com/asyncapi/spec/issues/618#issuecomment-1289044244. In the current state of spec v3, we have problems related to the creation of similar channels (with the same data), but differing in the list of available messages. 

Example: We can have application where we have one channel, but operations perform logic only on particular messages, so (in current spec v3) we need to define multiple channels with this same data, but with different `messages` list. When channel has only `address` and `messages` fields copying is easy, but however with situation when channel has defined `tags`, `externalDocs`, `bindings`, `parameters` etc, we can have situation like: 

```yaml
channels:
  ping:
    address: /{root}
    parameters:
      root: ...
    servers: 
      - $ref: #/servers/someServer1
      - $ref: #/servers/someServer2
    bindings:
      kafka:
        ...
    messages:
      ping:
        $ref: '#/components/messages/ping'
        
  pong:
    address: /{root}
    parameters:
      root: ...
    servers: 
      - $ref: #/servers/someServer1
      - $ref: #/servers/someServer2
    bindings:
      kafka:
        ...
    messages:
      pong:
        $ref: '#/components/messages/pong'

  heartbeat:
    address: /{root}
    parameters:
      root: ...
    servers: 
      - $ref: #/servers/someServer1
      - $ref: #/servers/someServer2
    bindings:
      kafka:
        ...
    messages:
      heartbeat:
        $ref: '#/components/messages/heartbeat'
      
  systemStatus:
    address: /{root}
    parameters:
      root: ...
    servers: 
      - $ref: #/servers/someServer1
      - $ref: #/servers/someServer2
    bindings:
      kafka:
        ...
    messages:
      systemStatus:
        $ref: '#/components/messages/systemStatus'
      
  subscriptionStatus:
    address: null
    messages:
      subscriptionStatus:
        $ref: '#/components/messages/subscriptionStatus'
        
  subscribe:
    address: /{root}
    parameters:
      root: ...
    servers: 
      - $ref: #/servers/someServer1
      - $ref: #/servers/someServer2
    bindings:
      kafka:
        ...
    messages:
      subscribe:
        $ref: '#/components/messages/subscribe'
        
  unsubscribe:
    address: /{root}
    parameters:
      root: ...
    servers: 
      - $ref: #/servers/someServer1
      - $ref: #/servers/someServer2
    bindings:
      kafka:
        ...
        
operations:
  pingPong:
    action: send
    channel: 
      $ref: '#/channels/ping'
    reply:
      channel: 
        $ref: '#/channels/pong'
  heartbeat:
    action: receive
    channel: 
      $ref: '#/channels/heartbeat' 
  systemStatus:
    action: receive
    channel: 
      $ref: '#/channels/systemStatus' 
  subscribe:
    action: send
    channel: 
      $ref: '#/channels/subscribe'
    reply:
      channel: 
        $ref: '#/channels/subscriptionStatus'
  unsubscribe:
    action: send
    channel: 
      $ref: '#/channels/unsubscribe'
    reply:
      channel: 
        $ref: '#/channels/subscriptionStatus'   
```

so as we see we have duplicated data like `parameters`, `bindings` etc. Of course we can use `$ref` but it doesn't solve problem at all, because we still needs to write a lot of code. `channels[x].traits` help us with that:

```yaml
channels:
  ping:
    traits:
      - $ref: '#/components/channelTraits/rootChannel'
    messages:
      ping:
        $ref: '#/components/messages/ping'
        
  pong:
    traits:
      - $ref: '#/components/channelTraits/rootChannel'
    messages:
      pong:
        $ref: '#/components/messages/pong'

  heartbeat:
    traits:
      - $ref: '#/components/channelTraits/rootChannel'
    messages:
      heartbeat:
        $ref: '#/components/messages/heartbeat'
      
  systemStatus:
    traits:
      - $ref: '#/components/channelTraits/rootChannel'
    messages:
      systemStatus:
        $ref: '#/components/messages/systemStatus'
      
  subscriptionStatus:
    address: null
    messages:
      subscriptionStatus:
        $ref: '#/components/messages/subscriptionStatus'
        
  subscribe:
    traits:
      - $ref: '#/components/channelTraits/rootChannel'
    messages:
      subscribe:
        $ref: '#/components/messages/subscribe'
        
  unsubscribe:
    traits:
      - $ref: '#/components/channelTraits/rootChannel'
     messages:
      subscribe:
        $ref: '#/components/messages/unsubscribe'
        
operations:
  pingPong:
    action: send
    channel: 
      $ref: '#/channels/ping'
    reply:
      channel: 
        $ref: '#/channels/pong'
  heartbeat:
    action: receive
    channel: 
      $ref: '#/channels/heartbeat' 
  systemStatus:
    action: receive
    channel: 
      $ref: '#/channels/systemStatus' 
  subscribe:
    action: send
    channel: 
      $ref: '#/channels/subscribe'
    reply:
      channel: 
        $ref: '#/channels/subscriptionStatus'
  unsubscribe:
    action: send
    channel: 
      $ref: '#/channels/unsubscribe'
    reply:
      channel: 
        $ref: '#/channels/subscriptionStatus'   
        
components:
  channelTraits:
    rootChannel:
      address: /{root}
      parameters:
        root: ...
      servers: 
        - $ref: #/servers/someServer1
        - $ref: #/servers/someServer2
      bindings:
        kafka:
          ...
```

As we can see we remove a lot of duplicated code.

Similar situation with `Server Trait Object`s - they are not as necessary as `Channel Trait Objects`, but due to the fact that we have traits for messages, operations and channels then servers should have them too.

Another solution to handle above problem with redundant channel data - [feat: add messages field in the Operation Object](https://github.com/asyncapi/spec/pull/870).

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -540,11 +540,11 @@ Describes a trait that MAY be applied to a [Server Object](#serverObject). This 
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="serverObjectDescription"></a>description | `string` | An optional string describing the host designated by the URL. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
-<a name="serverObjectVariables"></a>variables | Map[`string`, [Server Variable Object](#serverVariableObject) \| [Reference Object](#referenceObject)]] | A map between a variable name and its value.  The value is used for substitution in the server's URL template.
-<a name="serverObjectSecurity"></a>security | [[Security Requirement Object](#securityRequirementObject)] | A declaration of which security mechanisms can be used with this server. The list of values includes alternative security requirement objects that can be used. Only one of the security requirement objects need to be satisfied to authorize a connection or operation.
-<a name="serverObjectTags"></a>tags | [Tags Object](#tagsObject) | A list of tags for logical grouping and categorization of servers.
-<a name="serverObjectBindings"></a>bindings | [Server Bindings Object](#serverBindingsObject) \| [Reference Object](#referenceObject) | A map where the keys describe the name of the protocol and the values describe protocol-specific definitions for the server.
+<a name="serverTraitObjectDescription"></a>description | `string` | An optional string describing the host designated by the URL. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
+<a name="serverTraitObjectVariables"></a>variables | Map[`string`, [Server Variable Object](#serverVariableObject) \| [Reference Object](#referenceObject)]] | A map between a variable name and its value.  The value is used for substitution in the server's URL template.
+<a name="serverTraitObjectSecurity"></a>security | [[Security Requirement Object](#securityRequirementObject)] | A declaration of which security mechanisms can be used with this server. The list of values includes alternative security requirement objects that can be used. Only one of the security requirement objects need to be satisfied to authorize a connection or operation.
+<a name="serverTraitObjectTags"></a>tags | [Tags Object](#tagsObject) | A list of tags for logical grouping and categorization of servers.
+<a name="serverTraitObjectBindings"></a>bindings | [Server Bindings Object](#serverBindingsObject) \| [Reference Object](#referenceObject) | A map where the keys describe the name of the protocol and the values describe protocol-specific definitions for the server.
 
 This object MAY be extended with [Specification Extensions](#specificationExtensions).
 
@@ -752,13 +752,13 @@ Describes a trait that MAY be applied to an [Channel Object](#channelObject). Th
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="channelObjectAddress"></a>address | `string` \| `null` | An optional string representation of this channel's address. The address is typically the "topic name", "routing key", "event type", or "path". When `null` or absent, it MUST be interpreted as unknown. This is useful when the address is generated dynamically at runtime or can't be known upfront. It MAY contain [Channel Address Expressions](#channelAddressExpressions).
-<a name="channelObjectDescription"></a>description | `string` | An optional description of this channel. [CommonMark syntax](https://spec.commonmark.org/) can be used for rich text representation.
-<a name="channelObjectServers"></a>servers | [[Reference Object](#referenceObject)] | An array of `$ref` pointers to the definition of the servers in which this channel is available. If `servers` is absent or empty, this channel MUST be available on all the servers defined in the [Servers Object](#serversObject). Please note the `servers` property value MUST be an array of [Reference Objects](#referenceObject) and, therefore, MUST NOT contain an array of [Server Objects](#serverObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience.
-<a name="channelObjectParameters"></a>parameters | [Parameters Object](#parametersObject) | A map of the parameters included in the channel address. It MUST be present only when the address contains [Channel Address Expressions](#channelAddressExpressions).
-<a name="channelObjectBindings"></a>bindings | [Channel Bindings Object](#channelBindingsObject) \| [Reference Object](#referenceObject) | A map where the keys describe the name of the protocol and the values describe protocol-specific definitions for the channel.
-<a name="channelObjectTags"></a>tags | [Tags Object](#tagsObject) | A list of tags for logical grouping of channels.
-<a name="channelObjectExternalDocs"></a>externalDocs | [External Documentation Object](#externalDocumentationObject) \| [Reference Object](#referenceObject) | Additional external documentation for this channel.
+<a name="channelTraitObjectAddress"></a>address | `string` \| `null` | An optional string representation of this channel's address. The address is typically the "topic name", "routing key", "event type", or "path". When `null` or absent, it MUST be interpreted as unknown. This is useful when the address is generated dynamically at runtime or can't be known upfront. It MAY contain [Channel Address Expressions](#channelAddressExpressions).
+<a name="channelTraitObjectDescription"></a>description | `string` | An optional description of this channel. [CommonMark syntax](https://spec.commonmark.org/) can be used for rich text representation.
+<a name="channelTraitObjectServers"></a>servers | [[Reference Object](#referenceObject)] | An array of `$ref` pointers to the definition of the servers in which this channel is available. If `servers` is absent or empty, this channel MUST be available on all the servers defined in the [Servers Object](#serversObject). Please note the `servers` property value MUST be an array of [Reference Objects](#referenceObject) and, therefore, MUST NOT contain an array of [Server Objects](#serverObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience.
+<a name="channelTraitObjectParameters"></a>parameters | [Parameters Object](#parametersObject) | A map of the parameters included in the channel address. It MUST be present only when the address contains [Channel Address Expressions](#channelAddressExpressions).
+<a name="channelTraitObjectBindings"></a>bindings | [Channel Bindings Object](#channelBindingsObject) \| [Reference Object](#referenceObject) | A map where the keys describe the name of the protocol and the values describe protocol-specific definitions for the channel.
+<a name="channelTraitObjectTags"></a>tags | [Tags Object](#tagsObject) | A list of tags for logical grouping of channels.
+<a name="channelTraitObjectExternalDocs"></a>externalDocs | [External Documentation Object](#externalDocumentationObject) \| [Reference Object](#referenceObject) | Additional external documentation for this channel.
 
 This object MAY be extended with [Specification Extensions](#specificationExtensions).
 
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -378,7 +378,7 @@ Field Name | Type | Description
 <a name="serverObjectSecurity"></a>security | [[Security Requirement Object](#securityRequirementObject)] | A declaration of which security mechanisms can be used with this server. The list of values includes alternative security requirement objects that can be used. Only one of the security requirement objects need to be satisfied to authorize a connection or operation.
 <a name="serverObjectTags"></a>tags | [Tags Object](#tagsObject) | A list of tags for logical grouping and categorization of servers.
 <a name="serverObjectBindings"></a>bindings | [Server Bindings Object](#serverBindingsObject) \| [Reference Object](#referenceObject) | A map where the keys describe the name of the protocol and the values describe protocol-specific definitions for the server.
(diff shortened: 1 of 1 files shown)


magicmatatjahu:
I don't know why, but Github doesn't see my last commits with anchors fixes... I hope that it will be showed in next few hours, if not I will create another PR.

spec/asyncapi.md
```diff
@@ -524,6 +532,42 @@ servers:
...
 
+#### <a name="serverTraitObject"></a>Server Trait Object
+
+Describes a trait that MAY be applied to a [Server Object](#serverObject). This object MAY contain any property from the [Server Object](#serverObject), except `url`, `protocol` and `protocolVersion`.
+
+##### Fixed Fields
+
+Field Name | Type | Description
+---|:---:|---
+<a name="serverObjectDescription"></a>description | `string` | An optional string describing the host designated by the URL. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
```
magicmatatjahu, line 543: I know that here should be `serverTraitObjectDescription` anchor, but Github doesn't see my latest changes and `git status` tells me about up-to-date revisions...

spec/asyncapi.md
```diff
@@ -376,6 +378,7 @@ Field Name | Type | Description
 <a name="serverObjectSecurity"></a>security | [[Security Requirement Object](#securityRequirementObject)] | A declaration of which security mechanisms can be used with this server. The list of values includes alternative security requirement objects that can be used. Only one of the security requirement objects need to be satisfied to authorize a connection or operation.
 <a name="serverObjectTags"></a>tags | [Tags Object](#tagsObject) | A list of tags for logical grouping and categorization of servers.
 <a name="serverObjectBindings"></a>bindings | [Server Bindings Object](#serverBindingsObject) \| [Reference Object](#referenceObject) | A map where the keys describe the name of the protocol and the values describe protocol-specific definitions for the server.
+<a name="serverObjectTraits"></a>traits | [[Server Trait Object](#serverTraitObject) &#124; [Reference Object](#referenceObject) ] | A list of traits to apply to the server object. Traits MUST be merged into the server object using the [JSON Merge Patch](https://tools.ietf.org/html/rfc7386) algorithm in the same order they are defined here.
```
fmvilas, line 381:
```suggestion
<a name="serverObjectTraits"></a>traits | [[Server Trait Object](#serverTraitObject) \| [Reference Object](#referenceObject)] | A list of traits to apply to the server object. Traits MUST be merged into the server object using the [JSON Merge Patch](https://tools.ietf.org/html/rfc7386) algorithm in the same order they are defined here.
```

spec/asyncapi.md
```diff
@@ -388,7 +391,10 @@ A single server would be described as:
   "url": "development.gigantic-server.com",
   "description": "Development server",
   "protocol": "kafka",
-  "protocolVersion": "1.0.0"
+  "protocolVersion": "1.0.0",
+  "traits": [
+    { "$ref": "#/components/serverTraits/kafka" }
+  ]
```
fmvilas, line 397: Would it make sense to create another example for traits? I find it confusing here.

spec/asyncapi.md
```diff
@@ -524,6 +532,42 @@ servers:
 ```
 
 
+#### <a name="serverTraitObject"></a>Server Trait Object
+
+Describes a trait that MAY be applied to a [Server Object](#serverObject). This object MAY contain any property from the [Server Object](#serverObject), except `url`, `protocol` and `protocolVersion`.
```
fmvilas, line 537: Why not `protocolVersion`? Also, you're missing the `traits` property. Otherwise, traits could define traits that could define traits and so on.

spec/asyncapi.md
```diff
@@ -524,6 +532,42 @@ servers:
...
+    }
+  ]
+}
+```
+
+```yaml
+tags:
+  - name: "env:production"
+    description: "This environment is the live environment available for final users"
+```
```
fmvilas, line 568: Can you come up with a more complete example? For instance, `security` and `variables` would be super useful in many cases.

spec/asyncapi.md
```diff
@@ -570,6 +614,8 @@ defaultContentType: application/json
 
 An object containing all the [Channel Object](#channelObject) definitions the [Application](#definitionsApplication) MUST use during runtime.
 
+> If you're looking for a place to define channels that MAY or MAY NOT be implemented by the application, consider defining them in [`components/channels`](#componentsChannels).
```
fmvilas, line 617:
```suggestion
> If you're looking for a place to define channels that MAY or MAY NOT be used or implemented by the application, consider defining them in [`components/channels`](#componentsChannels).
```

spec/asyncapi.md
```diff
@@ -618,6 +664,7 @@ Field Name | Type | Description
 <a name="channelObjectBindings"></a>bindings | [Channel Bindings Object](#channelBindingsObject) \| [Reference Object](#referenceObject) | A map where the keys describe the name of the protocol and the values describe protocol-specific definitions for the channel.
 <a name="channelObjectTags"></a>tags | [Tags Object](#tagsObject) | A list of tags for logical grouping of channels.
 <a name="channelObjectExternalDocs"></a>externalDocs | [External Documentation Object](#externalDocumentationObject) \| [Reference Object](#referenceObject) | Additional external documentation for this channel.
+<a name="channelObjectTraits"></a>traits | [[Channel Trait Object](#channelTraitObject) &#124; [Reference Object](#referenceObject) ] | A list of traits to apply to the channel object. Traits MUST be merged into the channel object using the [JSON Merge Patch](https://tools.ietf.org/html/rfc7386) algorithm in the same order they are defined here.
```
fmvilas, line 667:
```suggestion
<a name="channelObjectTraits"></a>traits | [[Channel Trait Object](#channelTraitObject) \| [Reference Object](#referenceObject)] | A list of traits to apply to the channel object. Traits MUST be merged into the channel object using the [JSON Merge Patch](https://tools.ietf.org/html/rfc7386) algorithm in the same order they are defined here.
```

spec/asyncapi.md
```diff
@@ -660,7 +707,10 @@ This object MAY be extended with [Specification Extensions](#specificationExtens
   "externalDocs": {
     "description": "Find more info here",
     "url": "https://example.com"
-  }
+  },
+  "traits": [
+    { "$ref": "#/components/channelTraits/kafka" }
+  ]
```
fmvilas, line 713: Same as with servers. Would be cool to have another example for traits.

spec/asyncapi.md
```diff
@@ -685,12 +735,49 @@ bindings:
...
   description: 'Find more info here'
   url: 'https://example.com'
+traits:
+  - $ref: '#/components/channelTraits/kafka'
+```
+
+
+#### <a name="channelTraitObject"></a>Channel Trait Object
+
+Describes a trait that MAY be applied to an [Channel Object](#channelObject). This object MAY contain any property from the [Channel Object](#channelObject), except the `messages` and `traits` ones.
```
fmvilas, line 749: Why not `messages`? What if I want to define another channel with the same messages but a different address?

fmvilas (changes requested):

spec/asyncapi.md
magicmatatjahu: I made such simple examples of what operations and message traits have :trollface: I will extend it.

spec/asyncapi.md
magicmatatjahu: Yeah, I missed `traits` 🤦🏼‍♂️ Thanks for catch! However, not specifying `protocolVersion` is deliberate, because it has a correlation with `protocol`. You shouldn't define version like `0.1.0` and then in the server object define `protocol: kafka`, because you don't know if you `kafka` (example) has `0.1.0` version.

We could add `protocolVersion` to the `Server Trait Object`, but together with `protocol`. However `protocol` is mandatory field in the server so we will end up with this same situation as we had with `Operation Object` with `action` and `channel` fields.

spec/asyncapi.md
magicmatatjahu: This is the same problem as with `payload` in the message. You can't define payload in message traits, and likewise you shouldn't be able to add messages to the channel trait - look at my example in the PR description. The current main problem in defining channels is that we define messages on channel level (this is not a problem at all) but when we want to add a "filter" on the operation to inform user on which messages given operation operates, we need to define another channel with subset of messages - maybe this will fix that problem in more elegant way https://github.com/asyncapi/spec/pull/870

> Why not messages? What if I want to define another channel with the same messages but a different address?

This is some use case, but when we give the ability to define the messages at the level of the channel trait then someone can do something like this:

```yaml
channels:
  someChannel:
    address: ...
    messages:
      someMessage: ...
    traits:
      - messages:
           someMessage: ...
```

The ability to override information on the message from the level of the channel trait is a bad idea, that's why I want to avoid it.

spec/asyncapi.md
fmvilas: I think that's completely fine. People usually use one protocol only (two when they use http/ws for frontend too). I think it's up to them to take care that `protocolVersion` still makes sense after applying the trait 🤷

magicmatatjahu: Ok, if someone will start to combine with versioning and protocols it will be his/her problem 😄

spec/asyncapi.md
magicmatatjahu: TBH I don't understand. What you wanna add here? Similar examples of using traits we have in message and operation traits with `$ref`.

fmvilas: Yeah, I know, it's just that by looking at this example I simply would not understand what a server trait is if it were my first time with the spec.

spec/asyncapi.md
fmvilas: The parser should validate it's a legal combination anyway.

char0n:
Would it be possible to limit the scope of this PR to `server traits` and deal with `channel traits` in separate PR? Of course, only If there's not direct connection (one requires another), but from reviewing the PR it wasn't apparent there is any connection.

magicmatatjahu:
@char0n

> Would it be possible to limit the scope of this PR to server traits and deal with channel traits in separate PR? Of course, only If there's not direct connection (one requires another), but from reviewing the PR it wasn't apparent there is any connection.

I can, however I don't know if that PR is worth to focus for `3.0.0`. I think that if someone will need that we can reopen that. WDYT?

cc @fmvilas @derberg @dalelane @smoya

magicmatatjahu:
Closing - see the previous comment.

PR #864, opened 2022-10-26
chore: update AsyncAPI version to 3.0.0

jonaslagoni:
This PR changes the AsyncAPI version to 3.0.0.

I can't remember, was it in the spec that we did not want to change this for now @derberg? Because it would be creating breaking changes when we do version 2.6? (if we do)

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -8,7 +8,7 @@ This version is not yet ready to be used. We're currently working on it. If you 
 
 Part of this content has been taken from the great work done by the folks at the [OpenAPI Initiative](https://openapis.org). Mainly because **it's a great work** and we want to keep as much compatibility as possible with the [OpenAPI Specification](https://github.com/OAI/OpenAPI-Specification).
 
-#### Version 2.5.0
+#### Version 3.0.0
 
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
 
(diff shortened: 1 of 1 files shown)


smoya (approved):

dalelane:
good catch!

I'm guessing this is why the website shows us as having two 2.5.0 specs... with the one at the top of the list being a mis-labelled v3 spec. This is going to confuse anyone who doesn't look at the URL.


https://www.asyncapi.com/docs/reference/specification/v3.0.0-next-major-spec.4
image

smoya:
> good catch!
> 
> I'm guessing this is why the website shows us as having two 2.5.0 specs... with the one at the top of the list being a mis-labelled v3 spec. This is going to confuse anyone who doesn't look at the URL.
> 
> 
> https://www.asyncapi.com/docs/reference/specification/v3.0.0-next-major-spec.4
> image
> 

I swear 3.0 (prerelease) was shown in the top of that list few days ago instead. I don't get what happened tbh 😅
In fact, https://www.asyncapi.com/docs/reference/specification/v2.5.0 doesnt work at all (latest neither).

smoya (approved):

fmvilas:
🤔 Why isn't it showing the (Pre-release) label next to the 3.0.0 one? I know it says 2.5.0 but, you know, the one that's supposed to show 3.0.0. I fixed it here last month: https://github.com/asyncapi/website/pull/979.

jonaslagoni:
> 🤔 Why isn't it showing the (Pre-release) label next to the 3.0.0 one? I know it says 2.5.0 but, you know, the one that's supposed to show 3.0.0. I fixed it here last month: [asyncapi/website#979](https://github.com/asyncapi/website/pull/979).

Problem with the website then? 🤔

smoya:
I confirm this is the culprit for https://github.com/asyncapi/website/issues/1057

All the "bug" comes from https://github.com/asyncapi/website/blob/master/scripts/build-post-list.js#L64.

Will try to give more detail in a next comment.

smoya:
> I'm guessing this is why the website shows us as having two 2.5.0 specs... with the one at the top of the list being a mis-labelled v3 spec. This is going to confuse anyone who doesn't look at the URL.

The reason was a cache thing. See https://github.com/asyncapi/website/issues/1057.

derberg:
> I can't remember, was it in the spec that we did not want to change this for now @derberg? Because it would be creating breaking changes when we do version 2.6? (if we do)

yeah @jonaslagoni I think so, but I'm not 100% sure. I prefer to have it change, but I think it was something @fmvilas was writing

fmvilas (approved):

PR #847, opened 2022-10-04
feat: request/response support

GreenRover:
Title: "Request/Response paradigm support in operation"
Related issue(s): https://github.com/asyncapi/spec/issues/94 https://github.com/asyncapi/spec/issues/558
Champion: @GreenRover

_________________

**Explanation of Problem & Solution:**
Very common type of operation is currently missing in specification per discussion in referenced (related) issues. Operation that results in direct or indirect response to the caller.
The most basic example is publishing a message, where the result of publish operation should be known as soon as possible to the caller. Eg. invoking operation on given channel will return the specific message (or oneOf messages) right away to the caller.
The complicated example is correlation of the request and response, where publish/subscribe operation with given message will result in the specific message to be delivered back to the caller via different channel.

Solution has two parts:

1. Allowing referencing single or multiple messages as a result to operation (indifferent to whether the operation is server or client initiated)
2. Allowing hinting single or multiple messages that can be expected by other party to be received as a result to using specific message in any kind of operation

_________________

Please consume before take part to discussion:
- https://www.youtube.com/watch?v=koLWWoBnIMo

PR #870, opened 2022-11-09
feat: add messages field in the Operation Object

magicmatatjahu:
---
title: "add messages field in the Operation Object"
---

---

**Related issue(s):**
#618 
https://github.com/asyncapi/spec/pull/858 - another solution to fix "reusability" of messages in channels

---

Change is suggested by https://github.com/asyncapi/spec/issues/618#issuecomment-1289044244 and https://github.com/asyncapi/spec/issues/618#issuecomment-1289092374

`messages` (Messages Object) - A map of the messages that will be processed by this operation at any time during the runtime of the application. **The map should contain a subset of messages that are defined in the corresponding channel. If the field is omitted, the operation performs an action on every message defined in the corresponding channel.**


spec/asyncapi.md
```diff
@@ -821,6 +821,7 @@ Field Name | Type | Description
...
 <a name="operationObjectChannel"></a>channel | [Reference Object](#referenceObject) | **Required**. A `$ref` pointer to the definition of the channel in which this operation is performed. Please note the `channel` property value MUST be a [Reference Object](#referenceObject) and, therefore, MUST NOT contain a [Channel Object](#channelObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience.
+<a name="operationObjectMessages"></a>messages | [Messages Object](#messagesObject) | A map of the messages that will be processed by this operation at any time during the runtime of the application. **The map should contain a subset of messages that are defined in the corresponding channel. If the field is omitted, the operation performs an action on every message defined in the corresponding channel.**
```
dalelane, line 824: I'm not sure about 
> "messages that will be processed by this operation at any time during the runtime"

But I suck at writing English so I'm struggling to come up with something much better. 

What about: 

"A map of messages that the application should send to, or be able to receive from, the channel."

or 

"A map of messages that the application must send to, or be able to receive from, the channel."

?

fmvilas:
Let's hold this until we're clear with #847.

spec/asyncapi.md
fmvilas: Another suggestion:

```suggestion
<a name="operationObjectMessages"></a>messages | [Messages Object](#messagesObject) | A map of the messages that MAY be used by this operation. **The map MUST contain a subset of the messages that are defined in the corresponding channel. If the field is omitted, the operation MAY use any of the messages defined in the corresponding channel.**
```

magicmatatjahu: Ok, it's ok, pushed, but we should also change the description of `channel.messages` with `MAY` word to have regularity in descriptions. Thanks!

fmvilas: Channel messages definition is ok as it is. In the operation, it's our application the one that's responsible so that's why we gotta be careful with using `will` instead of `MAY` or `MUST`. But in the case of channels, it may be our application or another one, that's why it's fine to use `will` there. It's not our application contract. Messages will be sent there even if our application does nothing with them.

magicmatatjahu: Yeah, you're right. Thanks for clarification!

magicmatatjahu:
Will be supported in https://github.com/asyncapi/spec/pull/847

PR #872, opened 2022-11-14
ci: remove not needed workflow

codingtenshi:
This PR removes not needed golang and nodejs workflows. 

This change will optimize github actions runtime.

Thanks to [this](https://github.com/asyncapi/.github/blob/master/.github/workflows/global-replicator.yml#L65) and [this](https://github.com/asyncapi/.github/blob/master/.github/workflows/global-replicator.yml#L82) lines the file I just removed will not be created here by bot again.


derberg (approved):
Approving and merging without 3 approval as this is simple CI/CD related cleanup of workflows.

PR #873, opened 2022-11-16
chore: fix example for operation security

derberg:
`security` property on operation level is supposed to be an array of security requirement objects and not a single object -> https://www.asyncapi.com/docs/reference/specification/v2.5.0#operationObject

example before fix: https://studio.asyncapi.com/?url=https://raw.githubusercontent.com/asyncapi/spec/master/examples/operation-security.yml

example after fix: https://studio.asyncapi.com/?url=https://raw.githubusercontent.com/derberg/asyncapi/c61b29329b73867f98914c607f03015549169c34/examples/operation-security.yml

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 .vscode
+.github/.DS_Store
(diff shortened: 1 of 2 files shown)


magicmatatjahu (approved):

dalelane (approved):
LGTM

PR #875, opened 2022-11-20
chore: clarify introduction

### LINKED ISSUE #538: operationId usability in code generation
tl;dr we need a way to provide `operationId` for code generation of different applications, both server and clients, or in other words consumers and producers. So there is a need for different `operationId` names depending on the "perspective" or let's simply call it the "purpose of generated code"

Currently, AsyncAPI document should be written from the perspective of the "client", the user that interacts with described API.

This perspective causes for example a confusion around the understanding of [pub/sub semantics](https://github.com/asyncapi/spec/issues/520)

Another challenge with having only one "perspective" is around the `operationId`. 

Let's take the following part of AsyncAPI document:
```yml
/travel/status:
    subscribe:
        summary: You can listen to travel info status.
        operationId: onTravelInfo #client code perspective, generated client reacts "onTravelInfo" incomming message 
        message:
           $ref: '#/components/messages/travelInfo'
```

The client, the user subscribes to `/travel/status`. The generated code gets a function `onTravelInfo`, that reacts on incoming travel info. Sounds good, right?

Challenge is that I want to not only generate a code for the client, but also the server, the server that exposed the API described with AsyncAPI document. What shall I do with `operationId`. I either rename to to something like `subTravelInfo` which is neutral, but not super clear about the purpose and will make my generated code cryptic. I could also just rename it to `sendTravelInfo` but then have a situation that I do not follow the AsyncAPI approach to describe API from client perspective, make things even more confusing.

### LINKED ISSUE #859: [3.0.0] Incorrect introduction to AsyncAPI
### Describe the issue
In the current [version 3.0 of the spec](https://www.asyncapi.com/docs/reference/specification/v3.0.0-next-major-spec.3) the first introduction of what an AsyncAPI file describes:

>The file(s) MUST describe the operations an [application](https://www.asyncapi.com/docs/reference/specification/v3.0.0-next-major-spec.3#definitionsApplication) accepts. For instance, consider the following AsyncAPI definition snippet:
> ```yaml
> user/signedup:
>   subscribe:
>     message:
>       $ref: "#/components/messages/userSignUp"
> ```
> It means that the [application](https://www.asyncapi.com/docs/reference/specification/v3.0.0-next-major-spec.3#definitionsApplication) allows [consumers](https://www.asyncapi.com/docs/reference/specification/v3.0.0-next-major-spec.3#definitionsConsumer) to subscribe to the user/signedup [channel](https://www.asyncapi.com/docs/reference/specification/v3.0.0-next-major-spec.3#definitionsChannel) to receive userSignUp [messages](https://www.asyncapi.com/docs/reference/specification/v3.0.0-next-major-spec.3#definitionsMessage) produced by the application.

Is no longer valid with the switch of operation keywords from https://github.com/asyncapi/spec/pull/806

fmvilas:
---
title: "Clarify introduction"
---

---

**Related issue(s):**
Fixes #859 
Fixes #538

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -38,7 +38,7 @@ It means that the [application](#definitionsApplication) will subscribe to the `
 
 **The AsyncAPI specification does not assume any kind of software topology, architecture or pattern.** Therefore, a server MAY be a message broker, a web server or any other kind of computer program capable of sending and/or receiving data. However, AsyncAPI offers a mechanism called "bindings" that aims to help with more specific information about the protocol.
 
-It's NOT RECOMMENDED to derive a consumer AsyncAPI document from a producer one or viceversa. There are no guarantees that the channel used by an application to receive messages will be the same channel where another application is sending them. Also, certain fields in the document like `summary`, `description`, and the id of the operation might stop making sense. For instance, given the following consumer snippet:
+It's NOT RECOMMENDED to derive a consumer AsyncAPI document from a producer one or vice versa. There are no guarantees that the channel used by an application to receive messages will be the same channel where another application is sending them. Also, certain fields in the document like `summary`, `description`, and the id of the operation might stop making sense. For instance, given the following consumer snippet:
 
 ```yaml
 operations:
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -38,7 +38,7 @@ It means that the [application](#definitionsApplication) will subscribe to the `
 
 **The AsyncAPI specification does not assume any kind of software topology, architecture or pattern.** Therefore, a server MAY be a message broker, a web server or any other kind of computer program capable of sending and/or receiving data. However, AsyncAPI offers a mechanism called "bindings" that aims to help with more specific information about the protocol.
 
-It's NOT RECOMMENDED to derive a consumer AsyncAPI document from a producer one or vice versa. There are no guarantees that the channel used by an application to receive messages will be the same channel where another application is sending them. Also, certain fields in the document like `summary`, `description`, and the id of the operation might stop making sense. For instance, given the following consumer snippet:
+It's NOT RECOMMENDED to derive a [consumer](#definitionsConsumer) AsyncAPI document from a [producer](#definitionsProducer) one or vice versa. There are no guarantees that the channel used by an application to receive messages will be the same channel where another application is sending them. Also, certain fields in the document like `summary`, `description`, and the id of the operation might stop making sense. For instance, given the following consumer snippet:
 
 ```yaml
 operations:
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -54,15 +54,15 @@ We can't automatically assume that an _opposite_ application exists by simply re
 
 ```yaml
 operations:
-  onUserSignedUp: # <-- This doesn't make sense now
-    summary: On user signed up. # <-- This doesn't make sense now
-    description: Event received when a user signed up on the product. # <-- This doesn't make sense now
+  onUserSignedUp: # <-- This doesn't make sense now. Should be something like sendUserSignedUp.
+    summary: On user signed up. # <-- This doesn't make sense now. Should say something like "Sends a user signed up event".
+    description: Event received when a user signed up on the product. # <-- This doesn't make sense now. Should speak about sending an event, not receiving it.
     action: send
     channel:
       $ref: "#/channels/userSignedUp"
 ```
 
-Aside from the issues mentioned above, there may also be infrastructure restrictions that are not represented here. For instance, an application may use a channel for read-only operations (`receive`) so you can't send messages to the same channel.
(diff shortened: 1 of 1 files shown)


spec/asyncapi.md
```diff
@@ -16,24 +16,54 @@ The AsyncAPI Specification is licensed under [The Apache License, Version 2.0](h
...
 ```
 
-It means that the [application](#definitionsApplication) allows [consumers](#definitionsConsumer) to subscribe to the `user/signedup` [channel](#definitionsChannel) to receive userSignUp [messages](#definitionsMessage) produced by the application.
+It means that the [application](#definitionsApplication) will subscribe to the `userSignedUp` [channel](#definitionsChannel) to receive messages.
 
 **The AsyncAPI specification does not assume any kind of software topology, architecture or pattern.** Therefore, a server MAY be a message broker, a web server or any other kind of computer program capable of sending and/or receiving data. However, AsyncAPI offers a mechanism called "bindings" that aims to help with more specific information about the protocol.
 
+It's NOT RECOMMENDED to derive a consumer AsyncAPI document from a producer one or viceversa. There are no guarantees that the channel used by an application to receive messages will be the same channel where another application is sending them. Also, certain fields in the document like `summary`, `description`, and the id of the operation might stop making sense. For instance, given the following consumer snippet:
```
dalelane, line 41:
```suggestion
It's NOT RECOMMENDED to derive a consumer AsyncAPI document from a producer one or vice versa. There are no guarantees that the channel used by an application to receive messages will be the same channel where another application is sending them. Also, certain fields in the document like `summary`, `description`, and the id of the operation might stop making sense. For instance, given the following consumer snippet:
```

dalelane (approved):
other than one tiny typo, this looks helpful to me

spec/asyncapi.md
fmvilas: You guys write it separately? In Spanish, it's a single word 😅

spec/asyncapi.md
```diff
@@ -16,24 +16,54 @@ The AsyncAPI Specification is licensed under [The Apache License, Version 2.0](h
...
 ```
 
-It means that the [application](#definitionsApplication) allows [consumers](#definitionsConsumer) to subscribe to the `user/signedup` [channel](#definitionsChannel) to receive userSignUp [messages](#definitionsMessage) produced by the application.
+It means that the [application](#definitionsApplication) will subscribe to the `userSignedUp` [channel](#definitionsChannel) to receive messages.
 
 **The AsyncAPI specification does not assume any kind of software topology, architecture or pattern.** Therefore, a server MAY be a message broker, a web server or any other kind of computer program capable of sending and/or receiving data. However, AsyncAPI offers a mechanism called "bindings" that aims to help with more specific information about the protocol.
 
+It's NOT RECOMMENDED to derive a consumer AsyncAPI document from a producer one or vice versa. There are no guarantees that the channel used by an application to receive messages will be the same channel where another application is sending them. Also, certain fields in the document like `summary`, `description`, and the id of the operation might stop making sense. For instance, given the following consumer snippet:
```
derberg, line 41: `to derive a consumer AsyncAPI document from a producer one or vice versa` - this is the only sentence I'm not a fan of. 

In previous sentences, there is no concept of `consumer AsyncAPI document` and `producer AsyncAPI document` introduced. I know what you mean but I don't think this sentence explains it.

shouldn't we try to be more descriptive? that if AsyncAPI document describes API behaviour from API perspective, then it is not possible to use the same document to explain what AsyncAPI does from API user perspective?

or maybe we should even base it directly on code generation example?

derberg (changes requested):
left just one comment

spec/asyncapi.md
fmvilas: What about linking to the definitions of consumer and producer?

```suggestion
It's NOT RECOMMENDED to derive a [consumer](#definitionsApplication) AsyncAPI document from a [producer](#definitionsApplication) one or vice versa. There are no guarantees that the channel used by an application to receive messages will be the same channel where another application is sending them. Also, certain fields in the document like `summary`, `description`, and the id of the operation might stop making sense. For instance, given the following consumer snippet:
```

fmvilas:
Ping @derberg @dalelane @char0n

dalelane (approved):

spec/asyncapi.md
fmvilas: What do you think @derberg?

derberg: LGTM

derberg: oh wait, shouldn't links point to `definitionsProducer` and `definitionsConsumer`?

fmvilas: Yup, you're right. Fixed it 👍

fmvilas: @derberg please approve if you think it's good now.

derberg (approved):

fmvilas:
@dalelane would you mind re-approving 😅

dalelane (approved):

Issue #274, opened 2019-09-23
[BUG] Server Object field "url" description is inaccurate


gibson042:
**Describe the bug**
The AsyncAPI specification says about the Server Object `url` field that it contains a URL and "MAY be relative", but then uses as an example value "development.gigantic-server.com". This is not an (absolute) URL because it lacks a colon-terminated scheme such as "https", but if interpreted as a _relative_ reference then the result will be something like "https://example.com/example-service/asyncapi.json/../development.gigantic-server.com" (specifically because "_a relative reference that does not begin with a slash character is termed a **relative-path reference**_" and inherits the authority component and path prefix of its base URI, cf. [RFC 3986 section 4.2](https://tools.ietf.org/html/rfc3986#section-4.2) and [section 5.2.3](https://tools.ietf.org/html/rfc3986#section-5.2.3)).

**To Reproduce**
1. Navigate to https://www.asyncapi.com/docs/specifications/2.0.0/#serverObject or https://www.asyncapi.com/docs/specifications/1.2.0/#serverObject, or their respective git snapshots ([2.0.0](https://github.com/asyncapi/asyncapi/blob/1b20012080f5d21c134a1014b147ef860ee22f8a/versions/2.0.0/asyncapi.md#server-object), [1.2.0](https://github.com/asyncapi/asyncapi/blob/78e98e01a8077ced16a176f700f789e40df1d1fb/versions/1.2.0/asyncapi.md#server-object)).
2. Observe that `url` is a URL and "MAY be relative", but the absence of text allowing it to be a bare host name (or IP address, for that matter) despite the use of an example value that looks like one.

**Expected behavior**
Presumably the intent is to interpret values matching the [RFC 1034 section 3.5 preferred name syntax](https://tools.ietf.org/html/rfc1034#section-3.5) as updated by [RFC 1123 section 2.1](https://tools.ietf.org/html/rfc1123#section-2) (i.e., non-address strings matching the regular expression pattern `^[a-zA-Z0-9]+(-+[a-zA-Z0-9]+)*(\.[a-zA-Z0-9]+(-+[a-zA-Z0-9]+)*)*$`) as the host component of a URL with default scheme and path (e.g., "https://{value}/") and use the relative reference treatment only as a fallback. If so, the description should be updated accordingly.

And if there is also a desire to allow not just host names but also IP addresses, that should be documented as well.

**Sample document**
n/a

**Screenshots**
n/a

**Additional context**
n/a

[...]

PR #888, opened 2022-12-17
feat: replace server url with server host and pathname

fmvilas:
# Fix bug with Server's URL field

### Related issue(s):
#547 
#274

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -355,73 +355,70 @@ Field Pattern | Type | Description
 
 ```json
 {
-  "servers": {
-    "development": {
-      "host": "localhost:5672",
-      "description": "Development AMQP broker.",
-      "protocol": "amqp",
-      "protocolVersion": "0-9-1",
-      "tags": [
-        { 
-          "name": "env:development",
-          "description": "This environment is meant for developers to run their own tests."
-        }
-      ]
-    },
-    "staging": {
-      "host": "rabbitmq-staging.in.mycompany.com:5672",
-      "description": "RabbitMQ broker for the staging environment.",
-      "protocol": "amqp",
-      "protocolVersion": "0-9-1",
-      "tags": [
-        { 
-          "name": "env:staging",
-          "description": "This environment is a replica of the production environment."
-        }
-      ]
-    },
-    "production": {
-      "host": "rabbitmq.in.mycompany.com:5672",
-      "description": "RabbitMQ broker for the production environment.",
-      "protocol": "amqp",
-      "protocolVersion": "0-9-1",
-      "tags": [
-        { 
-          "name": "env:production",
-          "description": "This environment is the live environment available for final users."
-        }
-      ]
-    }
+  "development": {
+    "host": "localhost:5672",
+    "description": "Development AMQP broker.",
+    "protocol": "amqp",
+    "protocolVersion": "0-9-1",
+    "tags": [
+      { 
+        "name": "env:development",
+        "description": "This environment is meant for developers to run their own tests."
+      }
+    ]
+  },
+  "staging": {
+    "host": "rabbitmq-staging.in.mycompany.com:5672",
+    "description": "RabbitMQ broker for the staging environment.",
+    "protocol": "amqp",
+    "protocolVersion": "0-9-1",
+    "tags": [
+      { 
+        "name": "env:staging",
+        "description": "This environment is a replica of the production environment."
+      }
+    ]
+  },
+  "production": {
+    "host": "rabbitmq.in.mycompany.com:5672",
+    "description": "RabbitMQ broker for the production environment.",
+    "protocol": "amqp",
+    "protocolVersion": "0-9-1",
+    "tags": [
+      { 
+        "name": "env:production",
+        "description": "This environment is the live environment available for final users."
+      }
+    ]
   }
 }
 ```
 
 ```yaml
-servers:
-  development:
-    host: localhost:5672
-    description: Development AMQP broker.
-    protocol: amqp
-    protocolVersion: 0-9-1
-    tags:
-      - name: "env:development"
-        description: "This environment is meant for developers to run their own tests."
-  staging:
-    host: rabbitmq-staging.in.mycompany.com:5672
-    description: RabbitMQ broker for the staging environment.
-    protocol: amqp
-    protocolVersion: 0-9-1
-    tags:
-      - name: "env:staging"
-        description: "This environment is a replica of the production environment."
-  production:
-    host: rabbitmq.in.mycompany.com:5672
-    description: RabbitMQ broker for the production environment.
-    protocol: amqp
-    protocolVersion: 0-9-1
-    tags:
-      - name: "env:production"
-        description: "This environment is the live environment available for final users."
+development:
+  host: localhost:5672
+  description: Development AMQP broker.
+  protocol: amqp
+  protocolVersion: 0-9-1
+  tags:
+    - name: "env:development"
+      description: "This environment is meant for developers to run their own tests."
+staging:
+  host: rabbitmq-staging.in.mycompany.com:5672
+  description: RabbitMQ broker for the staging environment.
+  protocol: amqp
+  protocolVersion: 0-9-1
+  tags:
+    - name: "env:staging"
+      description: "This environment is a replica of the production environment."
+production:
+  host: rabbitmq.in.mycompany.com:5672
+  description: RabbitMQ broker for the production environment.
+  protocol: amqp
+  protocolVersion: 0-9-1
+  tags:
+    - name: "env:production"
+      description: "This environment is the live environment available for final users."
 ```
 
 
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -500,34 +500,28 @@ This object MAY be extended with [Specification Extensions](#specificationExtens
 
 ```json
 {
-  "servers": {
-    "production": {
-      "host": "api.gemini.com",
-      "pathname": "/v1/marketdata/{symbol}",
-      "protocol": "wss",
-      "description": "Market data is a public API that streams all the market data on a given symbol.",
-      "title": "Market Data API v1",
-      "variables": {
-        "symbol": {
-          "description": "Symbols are formatted as `CCY1CCY2` where prices are in `CCY2` and quantities are in `CCY1`."
-        }
-      }
+  "host": "api.gemini.com",
+  "pathname": "/v1/marketdata/{symbol}",
+  "protocol": "wss",
+  "description": "Market data is a public API that streams all the market data on a given symbol.",
+  "title": "Market Data API v1",
+  "variables": {
+    "symbol": {
+      "description": "Symbols are formatted as `CCY1CCY2` where prices are in `CCY2` and quantities are in `CCY1`."
     }
   }
 }
 ```
 
 ```yaml
-servers:
-  production:
-    host: 'api.gemini.com'
-    pathname: '/v1/marketdata/{symbol}'
-    protocol: wss
-    description: Market data is a public API that streams all the market data on a given symbol.
-    title: Market Data API v1
-    variables:
-      symbol:
-        description: Symbols are formatted as `CCY1CCY2` where prices are in `CCY2` and quantities are in `CCY1`.
+host: 'api.gemini.com'
+pathname: '/v1/marketdata/{symbol}'
+protocol: wss
+description: Market data is a public API that streams all the market data on a given symbol.
+title: Market Data API v1
+variables:
+  symbol:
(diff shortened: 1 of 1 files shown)


spec/asyncapi.md
```diff
@@ -552,8 +479,39 @@ Field Name | Type | Description
...
+  production:
+    host: '{org}.mycompany.com'
+    path: '/api/v1'
+    description: The production API server.
+    protocol: https
+    variables:
+      org:
+        default: demo
+        description: This value is assigned by the service provider, in this example `mycompany.com`.
+```
```
magicmatatjahu, line 514: You added here `Server Variable Object Example` but that describes the whole Server Object, here only the:

```yaml
        default: demo
        description: This value is assigned by the service provider, in this example `mycompany.com`.
```

should be written.

spec/asyncapi.md
```diff
@@ -460,83 +399,71 @@ The following shows how multiple servers can be described, for example, at the A
...
+
+An object representing a message broker, a server or any other kind of computer program capable of sending and/or receiving data. This object is used to capture details such as URIs, protocols and security configuration. Variable substitution can be used so that some details, for example usernames and passwords, can be injected by code generation tools.
+
+##### Fixed Fields
+
+Field Name | Type | Description
+---|:---:|---
+<a name="serverObjectHost"></a>host | `string` | **REQUIRED**. The server host name. It MAY include the port. This field supports [Server Variables](#serverObjectVariables). Variable substitutions will be made when a variable is named in `{`braces`}`.
+<a name="serverObjectProtocol"></a>protocol | `string` | **REQUIRED**. The protocol this server supports for connection.
+<a name="serverObjectProtocolVersion"></a>protocolVersion | `string` | The version of the protocol used for connection. For instance: AMQP `0.9.1`, HTTP `2.0`, Kafka `1.0.0`, etc.
```
magicmatatjahu, line 438: In one of examples you wrote (this example in the `Server Variable`) the `path` field, did you do it on purpose or did you forget to remove it?

spec/asyncapi.md
```diff
@@ -552,8 +479,39 @@ Field Name | Type | Description
...
 This object MAY be extended with [Specification Extensions](#specificationExtensions).
 
+##### Server Variable Object Example
 
+```json
+{
+  "servers": {
+    "production": {
+      "host": "{org}.mycompany.com",
+      "path": "/api/v1",
```
derberg, line 489: `path` ? where is this coming from?

derberg (commented):
I think title of PR should be more descriptive. Also PR should have short description so people do not have to read linked issues.

`fix: rename server.url to server.host` - something like that, to explain in one word what this change actually does, as later we use it in the change log

spec/asyncapi.md
fmvilas: We were having a "URL" and now we have a host. A host does not have the "path" part so we have to add it separately.

spec/asyncapi.md
fmvilas: Ah! true! Forgot to add path here 🤦 Thanks for the pointer.

spec/asyncapi.md
fmvilas: The other option is to remove the `protocol` field and go with a URL as we do now. Just not sure how this would play with protocols like `kafka-secure` (e.g., `kafka-secure://host:port/path`) and `secure-mqtt` (e.g., `secure-mqtt://host:port/path`), which are not "real" protocols. In such a case, we should probably get rid of the "secure" variants (that would include `https` and `wss`) and allow the user to specify it in the Security Requirement Object.

derberg: I'm wondering if we are actually fixing something, or making things a bit more complex. You add now `path` example that is a typical REST example for me. All the other examples I know (for example for Kafka) include a version in the topic name. In case of websocket it is completely fine to add path to channel name, but also if someone wants, can add it to `url`. The only use case for `path` that makes sense is HTTP, but then, if it is protocol specify, shouldn't it end up in the binding?

so basically my question is, do we really need to change `url` to `host` and `path` or, as majority of people in https://github.com/asyncapi/spec/issues/274 wrote that we basically need to make description and examples better?

fmvilas: There's are two things here:

1. We often show examples where the `url` is "whatever.example.com". That's not a URL and that should fail because it doesn't have the protocol part of the URL (you know, `protocol://`).
2. We have a field called `protocol` to define the protocol we want to use but it's already in the URL (!) ☝️ So you can end up with `url: kafka://broker.mycompany.com` and `protocol: https` in the same server definition, which is incongruent.

So we either remove the need to put a protocol in the `url` field (which makes it a host instead) or we remove the `protocol` field.

> You add now path example that is a typical REST example for me.

I think you're mixing concepts here. A path is not about REST or not even HTTP. A path is a part of a URL. URLs are composed by `protocol://`, `username:password` (optional`), `hostname`, `port` (optional), `/path/to/resource` (optional), `?query=parameters` (optional), and `#fragment` (optional). These have nothing to do with REST or HTTP. I can have an AMQP broker at `amqp://broker.mycompany.com/public` and another one at `amqp://broker.mycompany.com/partners`. Depending on the path (`public` or `partners`), I'll reach one broker or another. This is actually a common pattern, BTW.

So what I'm doing here is making `path` optional (defaulting to `/` as it's assumed by the URL RFC). In the majority of cases, this path will not be needed but in some cases, it will. And yeah, especially in the case of HTTP and WS it's going to be needed but it's also frequent in AMQP and MQTT brokers.

---

So yeah, I took the approach to keep the `protocol` field and convert the `url` field to a `host` (hostname + port) one. Another approach is to remove `protocol` altogether as it's already in the URL field. A matter of preference, I guess.

> so basically my question is, do we really need to change url to host and path or, as majority of people in https://github.com/asyncapi/spec/issues/274 wrote that we basically need to make description and examples better?

We definitely have to improve the examples but this incongruency goes beyond showing examples or improving description.

derberg: Well one thing is "concepts" and one thing is "just a simple association"

When I see
```
"host": "{org}.mycompany.com",
"path": "/api/v1",
"protocol": "https"
```

My brains says "REST". Then my brain also reminds me of AsyncAPI 1.0 and [baseTopic](https://github.com/asyncapi/spec/tree/1.2.0#baseTopicString). And then it tries to map it to concepts 😄 but yeah, enough with my brain 😄 

Can we then improve the example and not put `https` protocol but something that will not cause brains to fart?
For example your `amqp` example. I guess there are different patterns, as I thought people prefer to split `public/private` or `public/partners` by host as then they do different authentication, rate limits and stuff like that. But yeah, it is flexible and people can do it as they wish.

or best would be to have 2 examples, one with path and one without it:
- websocket from kraken provides different hosts (URL 😄 ) https://docs.kraken.com/websockets/#connectionDetails
- websocket from geminig (https://docs.gemini.com/websocket-api/#market-data) where host is `api.gemini.com` and `pathname` is `v1`

derberg: also the origin of my question

> path ? where is this coming from?

was that you use `path` but it is not described in the spec in your PR.

also side note: I think it should be `pathname` as we do not want to allow search (query params) here -> https://url.spec.whatwg.org/#dom-url-pathname

fmvilas: > Can we then improve the example and not put https protocol but something that will not cause brains to fart?

> or best would be to have 2 examples, one with path and one without it:

Absolutely. I'll add two examples.

> was that you use path but it is not described in the spec in your PR.

True. My fault. Adding it.

> also side note: I think it should be pathname as we do not want to allow search (query params) here -> https://url.spec.whatwg.org/#dom-url-pathname

Good catch. Changing it 👍

spec/asyncapi.md
fmvilas: I think in this case it's justified to have the whole Server Object. Otherwise, there would be no context of where this `demo` variable is coming from.

spec/asyncapi.md
```diff
@@ -399,13 +430,14 @@ An object representing a message broker, a server or any other kind of computer
...
+<a name="serverObjectDescription"></a>description | `string` | An optional string describing the server. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
 <a name="serverObjectTitle"></a>title | `string` | A human-friendly title for the server.
 <a name="serverObjectSummary"></a>summary | `string` | A short summary of the server.
-<a name="serverObjectDescription"></a>description | `string` | An optional string describing the host designated by the URL. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
-<a name="serverObjectVariables"></a>variables | Map[`string`, [Server Variable Object](#serverVariableObject) \| [Reference Object](#referenceObject)]] | A map between a variable name and its value.  The value is used for substitution in the server's URL template.
+<a name="serverObjectVariables"></a>variables | Map[`string`, [Server Variable Object](#serverVariableObject) \| [Reference Object](#referenceObject)]] | A map between a variable name and its value.  The value is used for substitution in the server's `host` template.
```
derberg, line 440:
```suggestion
<a name="serverObjectVariables"></a>variables | Map[`string`, [Server Variable Object](#serverVariableObject) \| [Reference Object](#referenceObject)]] | A map between a variable name and its value.  The value is used for substitution in the server's `host` and `pathname` template.
```

derberg (changes requested):
added one comment as I see your idea is that variables should be supported in `pathname`. Sounds good to me but we need an update.

Also, sorry if my previous comment sounded like I was suggesting that Gemini is a good example for `pathname`. I do not think it is, at least I modeled it differently in the official example -> https://github.com/asyncapi/spec/blob/a8c95348745dbeb5158f241f5b04ce7a1bbb6289/examples/websocket-gemini.yml

fmvilas:
> Also, sorry if my previous comment sounded like I was suggesting that Gemini is a good example for pathname. I do not think it is, at least I modeled it differently in the official example -> https://github.com/asyncapi/spec/blob/a8c95348745dbeb5158f241f5b04ce7a1bbb6289/examples/websocket-gemini.yml

I think it's still a good example. The difference is that you're defining the whole API and I'm just defining one endpoint 😄 But I get your point. Do you think we should change it to avoid confusions with your example?

derberg:
@fmvilas yeah, I'm afraid that 2 examples contradict each other. Maybe add amqp samples you mentioned in the other comment, with `public` and `private` to something like that

fmvilas:
@derberg I think examples are much better now. Have a look, please.

derberg (dismissed):
LGTM 🚀

derberg:
I only think if it is really a fix, and not feat. In the end we have new fields, different new behaviour, so it is more an enhancement, new feature than fix

fmvilas:
@dalelane @char0n @smoya would love to get approvals from you too. This is my last 3.0 PR 🎉

spec/asyncapi.md
```diff
@@ -446,117 +478,38 @@ A single server would be described as:
...
-  "tags": [
-    {
-      "name": "env:development"
-    }
-  ],
-  "externalDocs": {
-    "description": "Find more info here",
-    "url": "https://kafka.apache.org/"
-  }
+  "protocolVersion": "9"
```
dalelane, line 484:
```suggestion
  "protocolVersion": "3.2"
```

spec/asyncapi.md
```diff
@@ -446,117 +478,38 @@ A single server would be described as:
...
-        "8884"
-      ],
-      "default": "8883"
-    },
-    "basePath": {
-      "default": "v2"
-    }
-  }
-}
+protocolVersion: '9'
```
dalelane, line 492:
```suggestion
protocolVersion: '3.1'
```

dalelane (dismissed):
the protocol versions in the Kafka examples aren't valid Kafka versions - so I've added a couple of suggested fixes to replace them with Kafka version numbers

but approving as they're only examples and not super critical

char0n (approved):
LGTM. 

The only issue I found is in here: https://github.com/asyncapi/spec/blob/0d2c2355aae1048e4e881361f00801990205f4e2/spec/asyncapi.md?plain=1#L1628

Example is still using `url`  field instead of new one.

fmvilas:
@derberg @dalelane @char0n @smoya would please review and approve again? 🙏

magicmatatjahu:
@fmvilas Do you want to add the `json-schema` changes yourself? I can help you :trollface:

derberg (approved):

fmvilas:
@magicmatatjahu feel free mate! 😄

magicmatatjahu:
JSON Schema PR https://github.com/asyncapi/spec-json-schemas/pull/329
ParserJS PR https://github.com/asyncapi/parser-js/pull/708

char0n (approved):

dalelane (approved):

PR #892, opened 2023-01-17
chore: update `next-spec` with latest `master`

derberg:
This is just a typical update of release branch with latest changes from master.

There was only once merge conflict I had to solve related to `.github/workflows/if-nodejs-release.yml`. This file was removed from `master` and I had to enforce it on `next-spec` branch


fmvilas (approved):
👍 Let's make sure we merge this one with the Merge commit strategy instead of the Squash and Merge one, to avoid future conflicts.

dalelane (approved):

Issue #869, opened 2022-11-02
Are extensions allowed inside bindings?


derberg:
my opinion: Yes, should be

But this is not documented.

#### Current situation
- AsyncAPI spec described in markdown file do not say anything about it
- Bindings specs described in markdown file do not say anything about it
- because we assumed bindings should support extensions, JSON Schema files for bindings allow usage of `x-` extensions

#### Proposal

It should not be something that is agreed upon per binding, but it should be written down in AsyncAPI spec that each binding should support extensions

Thoughts?

char0n:
> AsyncAPI spec described in markdown file do not say anything about it

As a tooling author I assumed that extensions are not supported, given that spec, nor individual binding objects contain the following sentence: `This object MAY be extended with Specification Extensions.`

> because we assumed bindings should support extensions, JSON Schema files for bindings allow usage of x- extensions

I assumed the opposite as I considered the markdown spec as single source of truth and not the schemas. 

> It should not be something that is agreed upon per binding, but it should be written down in AsyncAPI spec that each binding should support extensions

Agreed.

---

Overall IMHO it's a good idea for individual binding objects to support extensions and to specify it explicitly in the markdown spec file.

char0n:
One more note: binding objects would need to be inspected as for example `amqp` Channel Binding Object look like this:

![image](https://user-images.githubusercontent.com/193286/199692517-4db48b92-1247-4a7a-8d0d-523dfcc6166d.png)

Notice the footer: `This object MUST contain only the properties defined above.`

jonaslagoni:
I actually thought it was supported.

Dont see anything wrong in adding support for it, as it gives you the option to play around with different setups before proposing the changes to the binding.

fmvilas:
> @fmvilas before I open a PR, do you remember why it was not added to the spec in the first place? don't want to work on a PR of there was already some discussion about it.

I don't. Most probably it was simply missed.

PR #893, opened 2023-01-17
feat: add info that protocol binding objects are extendable

derberg:
Resolves https://github.com/asyncapi/spec/issues/869

I wasn't sure what is the best way to specify it on binding level in the spec. Open to different ideas 🙏🏼 

@fmvilas @char0n @dalelane @smoya 

 I will also open separate PR to `bindings` repo to make it clear there for every single binding


dalelane (approved):
This seems like a good approach to me

char0n:
Hi @derberg,

The usual position for the sentence "`This object MAY be extended with [Specification Extensions](#specificationExtensions)`" is right before the list of fixed fields. In this PR you've changed this rule and IMHO this might bring future confusion. If I for example,  as a tooling author, don't find this sentence right after the list of object fixed fields, I'll automatically assume object doesn't support the spec extensions. Consistency is the king ;]

image

---

> Protocol-specific binding objects MAY be extended with [Specification Extensions](#specificationExtensions).

Regarding the above sentence - it might conflict with what the protocol specific binding objects actually demand (some demands no spec extensions). What would be the source of truth for this conflict resolution then? I'd assume the bindings overrides what AsyncAPI spec says. 

IMHO the best solution would be to retain consistency and for every protocol specific binding object we should either add the already established `This object MAY be extended with [Specification Extensions](#specificationExtensions)` or not (if not suitable). 

This would make is compatible with AsyncAPI spec and if protocol specific binding supports the spec extension, people will know where to find this information and will know what it means when this information is missing.

char0n:
> @char0n do you mean we do not need this PR and Protocol-specific binding objects MAY be extended with [Specification Extensions](#specificationExtensions). is not needed?

Yep, that's I'm trying to say.

> the goal of this sentence is to enforce a standard on all the bindings, that they need to be extensible, that it is not individual binding maintainer to disable extensions. If we do not have this here in the spec, how can we ensure it in bindings? this rule should be written in spec IMHO, not in some contribution guide in the bindings repo

I understand your motivation, but IMHO this sentence tries to convey how new bindings should be constructed, which should be part of PR review process to make sure bindings are consistent and always include the `This object MAY be extended with [Specification Extensions](#specificationExtensions)` sentence.

I see also an ambiguity in `Protocol-specific binding objects MAY be extended with [Specification Extensions](#specificationExtensions).` - does this mean that protocol bindings which are currently reserved supports the extension or not?

You've already issued https://github.com/asyncapi/bindings/pull/179, that handles extension on individual protocol binding level. Every binding object that can be extended with extension explicitly states that now. This is consistent with AsyncAPI spec itself and it should be enough for tooling authors.

---

Anyway I don't have strong opinion about this. If you leave the sentence there it's just fine. I was just trying to possibly bring a little bit more consistency and clarity, and eliminate ambiguity.

derberg:
Ok so what I did for now is moving info about binding extensibility back to its original location for consistency. The only change you can see is a new sentence about extensibility of specific binding object.

@dalelane @smoya @fmvilas 
can you folks look on the message from @char0n ?

I'm in doubt. Cause yes, adding the info I suggest is weird. But on the other hand, we cannot just count on the PR review process and that having extensions enabled everywhere is just a matter of consistency. I think we need to have it written somewhere, and still think bindings repo contribution guide is not a place for that

fmvilas (approved):
I think it's totally fine to put it here.

dalelane (approved):

derberg:
@smoya can you also have a look please?

PR #917, opened 2023-03-30
feat: new schema reference object to enable nonjson refs in future

derberg:
> PS. Let's pretend I created a branch in a fork, and not upstream, ok? 😅 

---

#### Breaking Change 

It must target 3.0

---

This PR is a result of:
- https://github.com/asyncapi/spec/pull/912#issuecomment-1477974563
- https://youtu.be/O4TQWBEaXy0?t=2213
- trying to make spec future-proof in case we would like to enable referencing schemas from nonJSON structures

What you see here is that `Reference Object` gets some more clarification and there is new `Schema Reference Object` that allows to have specification extensions.

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -1609,7 +1609,7 @@ In the case of a YAML-formatted Schema, the JSON Reference SHALL be applied to t
 ##### Fixed Fields
 Field Name | Type | Description
 ---|:---:|---
-<a name="referenceRef"></a>$ref | `string` | **REQUIRED.** The reference string.
+<a name="schemaReferenceRef"></a>$ref | `string` | **REQUIRED.** The reference string.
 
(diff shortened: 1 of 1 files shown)


dalelane (approved):
LGTM

derberg:
keeping it open, but looks like it is not needed for 3.0 and will be done in 3.1

jemacineiras:
Hi team,
Seeing this PR I am just curious about how AsyncApi V3 will handle Schemas Registry configuration for external schemas definition storage, like Avro, Protobuf or also Json Schema.
I had build some tools to generate code base on AsyncApi definition and some for testing, mostly of them support definition of schemas using the `$Ref` approach but then it came to have the schema definition in a place like Confluent Schema Registry, ApiCurio Schema Registry, Azure Schema Registry and so, I don't find it. 
I was thinking about how it can be solved, and looks a complex problem since base on your infrastructure you can have one schema registry server, or several, base on your environments.
I think this is something it should be kept in mind, since mostly of the people share AsyncApi documents between applications in order to generate code base on (myself have a small tool who create code where you can combine OpenApi and AsyncAp definitions).
If you wish I can open a suggestion base on this PR.. a PR base on a PR.. looks naughty.
Can you give me any light about?

derberg:
I'm closing it for now as it requires fresh look once we start working on it

PR #934, opened 2023-05-11
docs: align text about send receive and publisher and subscriber

derberg:
- it is mainly to align definitions with current actions. So instead of Producer and Consumer we have Sender and Receiver. Later in docs we can have more details with diagrams, that receiver can be, that it can be processor, or consumer and what is the difference. So basically Sender and Receiver are more generic and match the new `action` prop
- application can also be a broker
- operations are not mandatory really

---
some more context https://youtu.be/7urTb207V-8?t=229


spec/asyncapi.md
```diff
@@ -123,22 +123,22 @@ Aside from the issues mentioned above, there may also be infrastructure configur
...
 
 ### <a name="definitionsApplication"></a>Application
-An application is any kind of computer program or a group of them. It MUST be a [producer](#definitionsProducer), a [consumer](#definitionsConsumer) or both. An application MAY be a microservice, IoT device (sensor), mainframe process, etc. An application MAY be written in any number of different programming languages as long as they support the selected [protocol](#definitionsProtocol). An application MUST also use a protocol supported by the [server](#definitionsServer) in order to connect and exchange [messages](#definitionsMessage). 
+An application is any kind of computer program or a group of them. It MUST be a [sender](#definitionsSender), a [receiver](#definitionsReceiver) or both. An application MAY be a microservice, IoT device (sensor), mainframe process, or a message broker etc. An application MAY be written in any number of different programming languages as long as they support the selected [protocol](#definitionsProtocol). An application MUST also use a protocol supported by the [server](#definitionsServer) in order to connect and exchange [messages](#definitionsMessage).
```
fmvilas, line 129:
```suggestion
An application is any kind of computer program or a group of them. It MUST be a [sender](#definitionsSender), a [receiver](#definitionsReceiver), or both. An application MAY be a microservice, IoT device (sensor), mainframe process, message broker, etc. An application MAY be written in any number of different programming languages as long as they support the selected [protocol](#definitionsProtocol). An application MUST also use a protocol supported by the [server](#definitionsServer) in order to connect and exchange [messages](#definitionsMessage). 
```

spec/asyncapi.md
```diff
@@ -123,22 +123,22 @@ Aside from the issues mentioned above, there may also be infrastructure configur
...
 
-### <a name="definitionsProducer"></a>Producer
-A producer is a type of application, connected to a [server](#definitionsServer), that is creating [messages](#definitionsMessage) and addressing them to [channels](#definitionsChannel). A producer MAY be publishing to multiple channels depending on the [server](#definitionsServer), protocol, and use-case pattern.
+### <a name="definitionsSender"></a>Sender
+A sender is a type of application, that is sending [messages](#definitionsMessage) to [channels](#definitionsChannel). A sender MAY be publishing to multiple channels depending on the [server](#definitionsServer), protocol, and use-case pattern.
```
fmvilas, line 132:
```suggestion
A sender is a type of application, that is sending [messages](#definitionsMessage) to [channels](#definitionsChannel). A sender MAY send to multiple channels depending on the [server](#definitionsServer), protocol, and use-case pattern.
```

spec/asyncapi.md
```diff
@@ -123,22 +123,22 @@ Aside from the issues mentioned above, there may also be infrastructure configur
...
 
-### <a name="definitionsConsumer"></a>Consumer
-A consumer is a type of application, connected to a [server](#definitionsServer) via a supported [protocol](#definitionsProtocol), that is consuming [messages](#definitionsMessage) from [channels](#definitionsChannel). A consumer MAY be consuming from multiple channels depending on the [server](#definitionsServer), protocol, and the use-case pattern.
+### <a name="definitionsReceiver"></a>Receiver
+A receiver is a type of application, that is receiving [messages](#definitionsMessage) from [channels](#definitionsChannel). A receiver MAY be receiving from multiple channels depending on the [server](#definitionsServer), protocol, and the use-case pattern. A receiver MAY send received message further without changing it. A receiver MAY act as a consumer and react to the message. A receiver MAY act as a processor that, for example, aggregates multiple messages in one and sends it further.
```
fmvilas, line 135:
```suggestion
A receiver is a type of application that is receiving [messages](#definitionsMessage) from [channels](#definitionsChannel). A receiver MAY receive from multiple channels depending on the [server](#definitionsServer), protocol, and the use-case pattern. A receiver MAY forward a received message further without changing it. A receiver MAY act as a consumer and react to the message. A receiver MAY act as a processor that, for example, aggregates multiple messages in one and forwards them.
```

spec/asyncapi.md
```diff
@@ -1932,7 +1932,7 @@ The following properties are taken from the JSON Schema definition but their def
 
 - description - [CommonMark syntax](https://spec.commonmark.org/) can be used for rich text representation.
 - format - See [Data Type Formats](#dataTypeFormat) for further details. While relying on JSON Schema's defined formats, the AsyncAPI Specification offers a few additional predefined formats.
-- default - The default value represents what would be assumed by the consumer of the input as the value of the schema if one is not provided. Unlike JSON Schema, the value MUST conform to the defined type for the Schema Object defined at the same level. For example, of `type` is `string`, then `default` can be `"foo"` but cannot be `1`.
+- default - The default value represents what would be assumed by the receiver of the input as the value of the schema if one is not provided. Unlike JSON Schema, the value MUST conform to the defined type for the Schema Object defined at the same level. For example, of `type` is `string`, then `default` can be `"foo"` but cannot be `1`.
```
fmvilas, line 1935: I don't think this change applies. This `consumer` has nothing to do with producer/consumer stuff.

```suggestion
- default - The default value represents what would be assumed by the consumer of the input as the value of the schema if one is not provided. Unlike JSON Schema, the value MUST conform to the defined type for the Schema Object defined at the same level. For example, of `type` is `string`, then `default` can be `"foo"` but cannot be `1`.
```

fmvilas (changes requested):

spec/asyncapi.md
derberg: but in the end it related to schema of message, 

`The default value represents what would be assumed by the consumer`

I think consumer here means the consumer of the message described with the schema

dalelane (dismissed):
not as bad as the can of worms I feared when I saw the title about "publisher/subscriber" "send/receive" ;-)

spec/asyncapi.md
fmvilas: But "consumer" of the JSON Schema may just be a code generation tool for the producer. So the "consumer" here would really be the producer, actually 😅

derberg:
@smoya @fmvilas @char0n all good from your side?

spec/asyncapi.md
fmvilas: Love this one:

> Use it to specify that a property has a predefined value if no other value is present

derberg:
one small sentence change

@dalelane wanna have a look?

@fmvilas 🙏🏼

fmvilas (approved):
👍

smoya (approved):
LGTM! 🚀🌔 Thanks for the changes @derberg 🙌 

As a side note, I had to think about why we moved from `Subscribe/Publish` to `Send/Receive`.

Even though I actively participated in the original discussion and I completely agreed with that.

That means **It could** make sense to add a line in the docs clarifying why **we do not** use the words `Subscribe/Publish` anymore. WDYT?  Obviously, all in a separate issue and PR.

derberg:
@smoya good point, issue created https://github.com/asyncapi/website/issues/1873

/rtm

Issue #691, opened 2022-01-19
Work on 3.0 release


jonaslagoni:
Release 3.0.0 is scheduled for (yea nothing has been decided yet🤷). Depends upon how many champion issues.

## Detailed info
- https://github.com/asyncapi/spec/blob/master/RELEASE_PROCESS.md

## Kick-off

- release coordinating and contributing: https://www.youtube.com/watch?v=9lJY5NcLMKY
- Contribution guidelines: https://github.com/asyncapi/spec/blob/master/CONTRIBUTING.md

## Meetings

- https://github.com/asyncapi/community/issues/235#issuecomment-1016483507

## Spec champions 🔥 
These are the issues that are currently being championed by people, aiming for the 3.0 release.

- [ ] #622 @magicmatatjahu 
- [ ] #596 @magicmatatjahu  
- [ ] #628 @jonaslagoni
- [ ] #658 @jonaslagoni
- [ ] https://github.com/asyncapi/spec/issues/661 @smoya 
- [ ] https://github.com/asyncapi/spec/issues/663 @smoya 
- [ ] https://github.com/asyncapi/spec/issues/689 @smoya 
- [ ] https://github.com/asyncapi/spec/issues/654 @smoya  

## I want to champion something!
First of all, thank you! 🙇 

We use the following milestone https://github.com/asyncapi/spec/milestone/18 to track all the issues that needs to have been considered before we can release 3.0. So pick of the issues, and drive it forward! 

Reach out to slack, in the issue, or our regular meetings if you need anything!

If you need anything changed in this progress issue, leave a comment belove with what to update 🙂

magicmatatjahu:
I don't think so that I should handle https://github.com/asyncapi/spec/issues/596 I created issue but I don't know if I can find the time to do this, I would rather take these [schema formats](https://github.com/asyncapi/spec/issues/622) as well as https://github.com/asyncapi/spec/issues/583 if anyone wants to take it, go ahead. @smoya maybe you? I know you know a lot about these changes in the new drafts and what impact they would have.

damaru-inc:
I'd like to see schema versioning considered as a feature in 3.0. Please see this issue: https://github.com/asyncapi/spec/issues/697

magicmatatjahu:
I think that we should create branch for that `3.0.0` release like `next-branch` or `3.0.0-release` and start creating PR proposals. I don't know if that topic was raised in other issue/discussion. What do you think?

jonaslagoni:
[...]

PR #935, opened 2023-05-11
feat: simplify parameter object

### LINKED ISSUE #583: Inconsistencies for schemas in server.variables and channel.parameters

fmvilas:
Following our conversations in the Spec 3.0 meetings, here's the simplification of the Parameter Object.

Fixes #583

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -1014,48 +1014,34 @@ location: $message.header#/replyTo
 
 #### <a name="parametersObject"></a>Parameters Object
 
-Describes a map of parameters included in a channel name.
+Describes a map of parameters included in a channel address.
 
-This map MUST contain all the parameters used in the parent channel name.
+This map MUST contain all the parameters used in the parent channel address.
 
 ##### Patterned Fields
 
 Field Pattern | Type | Description
 ---|:---:|---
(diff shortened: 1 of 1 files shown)


dalelane (approved):
👍

jonaslagoni:
@fmvilas is it true this require no changes to the parser nor spec-json-schema?

fmvilas:
🤔 No, I think it does require changes in these repos. I just didn't do them. Easy change though.

jonaslagoni:
Alright, added it to the progress section of https://github.com/asyncapi/spec/issues/691 to track it.

Issue #949, opened 2023-06-29
Ensure consistency when using either `Application` or `API` terms


smoya:
### Description
Current spec (and also v3) mixes the terms `Application` and `API` (and derivatives) to refer to the same thing.

For example, some of the [Info Object](https://www.asyncapi.com/docs/reference/specification/v2.6.0#infoObject) fields say:

> title: The title of the application.
> version: Provides the version of the application API.
> description: A short description of the application.
> contact: The contact information for the exposed API.

By taking the following statement that can be found in the [Introduction](https://www.asyncapi.com/docs/reference/specification/v2.6.0#introduction):

> The AsyncAPI Specification defines a set of files required to describe the API of an [application](https://www.asyncapi.com/docs/reference/specification/v3.0.0-next-major-spec.12#definitionsApplication).

I vote for unifying these under one term to avoid any possible confusion. I believe we should stick with the `Application API` term.

cc @derberg @jonaslagoni @alequetzalli @dalelane @fmvilas

timmiedinnie:
As stated in my PR:

I agree there should be a change in favor of consistency.

However the suggestion to replace "Application" with "Application API" may not be suitable in most contexts. The resulting phrase, "The Application's Application programming interface," could be confusing or redundant.

In this proposed change, only a few instances of "Application" were replaced with "Application API," and it may not be appropriate in all situations. Furthermore, replacing "API" with "Application API" often creates even more awkward phrasing.

Therefore, I have submitted https://github.com/asyncapi/spec/pull/950 for review purposes only. In my opinion, using the term "API" consistently would be a better approach.

smoya:
To the rest of codeowners @fmvilas @derberg @dalelane @char0n:  do you think this makes sense?

sagnik3788:
hello @smoya  can you assign me any issue so that i can work on docs

smoya:
> Therefore, I have submitted #950 for review purposes only. In my opinion, using the term "API" consistently would be a better approach.

Sorry for not coming back to this earlier, but I noticed I never shared I also agree, after reviewing your PR, that the simple term "API" could fit better.

Would you be happy to work on that and close https://github.com/asyncapi/spec/pull/950 @timmiedinnie ?

smoya:
With your 👍 reaction, I understand you @timmiedinnie are gonna work on the suggested change. 
I closed https://github.com/asyncapi/spec/pull/950 in favour of that new PR you will create.

Thank you so much! 🙌

[...]

PR #950, opened 2023-06-29
docs: improved consistency using terms Application or API #949

timmiedinnie:
Changed several terms Application to Application API

---
 "Consistency  when using terms Application or API "
---

---

The suggestion to replace "Application" with "Application API" may not be suitable in most contexts. The resulting phrase, "The Application's Application programming interface," could be confusing or redundant.

In this proposed change, only a few instances of "Application" were replaced with "Application API," and it may not be appropriate in all situations. Furthermore, replacing "API" with "Application API" often creates even more awkward phrasing.

Therefore, I have submitted this pull request for review purposes only. In my opinion, using the term "API" consistently would be a better approach. Please review this pull request accordingly.

I would advise against accepting the pull request and suggest exploring alternative solutions.

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -17,7 +17,7 @@ The AsyncAPI Specification is a project used to describe and document message-dr
 The AsyncAPI Specification defines a set of files required to describe such an API.
 These files can then be used to create utilities, such as documentation, integration and/or testing tools.
 
-The file(s) MUST describe the operations an [application](#definitionsApplication) accepts. For instance, consider the following AsyncAPI definition snippet:
+The file(s) MUST describe the operations an [Application API](#definitionsApplication) accepts. For instance, consider the following AsyncAPI definition snippet:
 
 ```yaml
 user/signedup:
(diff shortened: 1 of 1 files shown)


smoya:
This is brilliant @timmiedinnie. Perfect for starting a discussion around it 👏 👏  Thanks for creating this PR.

spec/asyncapi.md
```diff
@@ -732,7 +732,7 @@ bindings:
 
 Describes a publish or a subscribe operation. This provides a place to document how and why messages are sent and received.
 
-For example, an operation might describe a chat application use case where a user sends a text message to a group. A publish operation describes messages that are received by the chat application, whereas a subscribe operation describes messages that are sent by the chat application.
+For example, an operation might describe a chat Application API use case where a user sends a text message to a group. A publish operation describes messages that are received by the chat Application API, whereas a subscribe operation describes messages that are sent by the chat Application API.
```
smoya, line 735: These are clear examples where the original term should stay.

smoya:
Closing as agreed in https://github.com/asyncapi/spec/issues/949#issuecomment-1771844148

Issue #948, opened 2023-06-27
[v3] Support multi format schemas in message headers


smoya:
### Description
The current v3 spec mentions that the [Message Object](https://www.asyncapi.com/docs/reference/specification/v3.0.0-next-major-spec.12#messageObject) headers must be of type "object". Literally says: 

> Schema definition of the application headers. Schema MUST be of type "object". It MUST NOT define the protocol headers. If this is a [Schema Object](https://www.asyncapi.com/docs/reference/specification/v3.0.0-next-major-spec.12#schemaObject), then the schemaFormat will be assumed to be "application/vnd.aai.asyncapi+json;version=asyncapi" where the version is equal to the [AsyncAPI Version String](https://www.asyncapi.com/docs/reference/specification/v3.0.0-next-major-spec.12#A2SVersionString).

Since the introduction of the support of [multi-format schemas](https://github.com/asyncapi/spec/pull/910/files), message headers also support different schema formats. 
Then, in my head, reading the sentence `Schema MUST be of type "object".` sounds confusing to me, especially for formats like Avro, where (afaik) there is no such type but `record` iirc.
This constraint is also in place in the AsyncAPI JSON Schema documents describing the message object. See https://github.com/asyncapi/spec-json-schemas/blob/next-major-spec/definitions/3.0.0/messageObject.json#L21-L27

Is that restriction in place with the purpose of limiting and simplifying what a header looks like? Shall we keep it and update it according to different possible shapes depending on the format? Or rather, after the recent changes, should we get rid of it?

Relates to https://github.com/asyncapi/spec/issues/622

cc @GreenRover @dalelane @derberg @jonaslagoni @fmvilas

derberg:
yeah, sounds like `Schema MUST be of type "object"` is just a left over that should be removed. I don't think it is needed.

GreenRover:
I vote for keep headers to be only async-api schema formated. To keep it simple. I see no benefit from defining header for example as avro.

GreenRover:
Ok, my memories slowly come back. I guess Fran brought an reasons that required non ansycapi-schema as header.
But you @smoya only have concerns about the wording?

What i wanted to say is: Schema MUST be of type "object".  
A headers have to be some kind of map. 
A string/number like in the body is not valid.

smoya:
> But you @smoya only have concerns about the wording?

[...]

PR #939, opened 2023-05-24
ci: update release workflow

kaushik-rishi:
---
title: "Update release workflow"
---

---

**Related issue(s):**
Fixes #938 

---

PR #952, opened 2023-07-13
fix: clarify expected type for Message headers

smoya:
Comes from https://github.com/asyncapi/spec/issues/948#issuecomment-1624853505

This PR clarifies the expected type and shape for Message headers.


smoya:
@fmvilas would you mind taking a look to this? 🙏

derberg (approved):

derberg:
please do not merge before https://github.com/asyncapi/spec/pull/939 as release is somehow broken and RC for traits for example was not out

PR #955, opened 2023-07-20
docs: add info that runtime expression is used by reply obj

derberg:
small stuff that we forgot, just noticed today when I was explaining req/rep to one of content contributors


fmvilas (approved):

smoya (approved):
Nice catch! 🙌

Issue #938, opened 2023-05-24
update release workflow in release branch


derberg:
Release https://github.com/asyncapi/spec/blob/next-major-spec/.github/workflows/release.yml needs update

Because of https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

```yml
      - name: Setup Node.js
        uses: actions/setup-node@v1
```

needs to be replaced with
```yml
- name: Setup Node.js
   uses: actions/setup-node@v3
   with:
      node-version: 18
```

also 
```
npm install conventional-changelog-conventionalcommits
```

update to
```
npm install --save-dev conventional-changelog-conventionalcommits@5.0.0
```

and switch:
```
actions/checkout@v2
```

to 
```
actions/checkout@v3
```

kaushik-rishi:
@derberg let's say i made the contribution, how would we generally test these out ? just by triggering the workflow ?

kaushik-rishi:
Also, this PR/commit should be a `chore` and not a `feat` right ?

derberg:
>  just by triggering the workflow ?

only way is to get it to some test repo and test it there 🤷🏼 

> Also, this PR/commit should be a chore and not a feat right ?

`ci` 😉

PR #956, opened 2023-07-25
ci: update release workflow on release branch

kaushik-rishi:
---
title: "Update the release workflow on the release branch (next-major-spec)"
---

---

**Related issue(s):**
Fixes https://github.com/asyncapi/spec/issues/938

---

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,7 +18,7 @@ jobs:
         with:
           node-version: 18
       - name: Add plugin for conventional commits
-        run: npm install --save-dev conventional-changelog-conventionalcommits@5.0.0
+        run: npm install --save-dev conventional-changelog-conventionalcommits@6.1.0
       - name: Release to GitHub
         env:
           GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,4 +26,4 @@ jobs:
           GIT_AUTHOR_EMAIL: info@asyncapi.io
           GIT_COMMITTER_NAME: asyncapi-bot
           GIT_COMMITTER_EMAIL: info@asyncapi.io
-        run: npx semantic-release@17.4.3
+        run: npx semantic-release@21.0.7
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -11,13 +11,14 @@ jobs:
...
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Setup Node.js
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
       - name: Add plugin for conventional commits
-        run: npm install conventional-changelog-conventionalcommits
+        run: npm install --save-dev conventional-changelog-conventionalcommits@5.0.0
```
derberg, line 21:
```suggestion
        run: npm install --save-dev conventional-changelog-conventionalcommits@6.1.0
```

they release so fast 😄

derberg (commented):
I'm not 100% sure all these updates will fix the release looking at the error:

```
[9:11:12 PM] [semantic-release] › ℹ  Start step "analyzeCommits" of plugin "@semantic-release/commit-analyzer"
(node:1734) UnhandledPromiseRejectionWarning: TypeError: config.types.flatMap is not a function
    at getWriterOpts (/home/runner/work/spec/spec/node_modules/conventional-changelog-conventionalcommits/writer-opts.js:76:41)
    at module.exports (/home/runner/work/spec/spec/node_modules/conventional-changelog-conventionalcommits/writer-opts.js:47:22)
(node:1734) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:1734) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

so better if you also `npx semantic-release@17.4.3` -> `npx semantic-release@21.0.7`

I checked and the only breaking changes are npm related (not problem here) and requirement for node 18, which is already done in this PR

so we should be good to go

derberg:
@kaushik-rishi I cannot make code change suggestion for `npx semantic-release@21.0.7` so you need to change it on your own

kaushik-rishi:
Updated @derberg .

derberg (approved):
Thanks!

PR #965, opened 2023-08-22
Gitter example: fix header format

lornajane:
---
title: "Gitter example: fix header format"
---

---

**Related issue(s):**

---

I noticed (when working on tooling for AsyncAPI) that the Gitter example we have doesn't follow the HTTP Message Bindings format - but it almost does, so I fixed it :)

No spec changes, just an update to an example file


examples/gitter-streaming.yml
```diff
@@ -160,5 +160,7 @@ components:
     streamingHeaders:
       http:
         headers:
-          'Transfer-Encoding': 'chunked'
-          Trailer: '\r\n'
+          type: object
+          properties:
+            'Transfer-Encoding': 'chunked'
+            Trailer: '\r\n'
```
fmvilas, line 166: Did you mean this instead? 🤔 

```suggestion
            'Transfer-Encoding':
              const: 'chunked'
            Trailer:
              const: '\r\n'
```

fmvilas (changes requested):

fmvilas (approved):

PR #988, opened 2023-11-10
fix: remove mentions of Operation Item Object

char0n:
---
title: "Fix mentions of non existing Operation Item Object"
---

There is actually no Operation Item Object,
just an Operation Object.

---


smoya (approved):
Good catch! LGTM! 🚀🌔

dalelane (approved):

Issue #980, opened 2023-10-18
Provide definition for the reply


Tenischev:
In current Spec v3 pre-release the `reply` appear first time as `operationObject` field:
https://github.com/asyncapi/spec/blob/abfae495b0fbb88fccd56f8099e466a83f4b76ab/spec/asyncapi.md?plain=1#L843
with the description `The definition of the reply in a request-reply operation`.
But Spec v3 has no definition or describing what is meant under `request-reply operation`.
@jonaslagoni wrote the [blog post](https://eventstack.tech/posts/asyncapi-v3-request-reply) describing what it could be, but the blog post is not the Specification. Blog post could be a spice for a cake, but not the cake.

I suggest to provide definition of `reply` in the [Definitions](https://github.com/asyncapi/spec/blob/next-major-spec/spec/asyncapi.md#definitions) section, maybe as subpart of Message definition.

smoya:
> I suggest to provide definition of `reply` in the [Definitions](https://github.com/asyncapi/spec/blob/next-major-spec/spec/asyncapi.md#definitions) section, maybe as subpart of Message definition.

Wondering if it is not already covered by the description of what an [Operation Reply Object](https://github.com/asyncapi/spec/blob/next-major-spec/spec/asyncapi.md#user-content-operation-reply-object) means.

> Describes the reply part that MAY be applied to an Operation Object. If an operation implements the request/reply pattern, the reply object represents the response message.

Tenischev:
For me description of [Operation Reply Object](https://github.com/asyncapi/spec/blob/next-major-spec/spec/asyncapi.md#user-content-operation-reply-object) and definition of [Message](https://github.com/asyncapi/spec/blob/next-major-spec/spec/asyncapi.md#definitionsMessage) are a bit not match.

Message:
> Since a message is a generic mechanism, it can support multiple interaction patterns such as **event, command, request, or response.**

Reply Object:
>  If an operation implements the **request/reply pattern**, the **reply** object **represents the response** message.

I feel here a gap, like "what exactly is `request`?", consequently "is it `send` or `receive` operation?". (_For me now it's clear, but I feel some degree of uncertain, when read Spec first time_)

If we have a look of [EIP request-reply definition](https://www.enterpriseintegrationpatterns.com/patterns/messaging/RequestReply.html), I would say it a bit more solid

> Send a pair of Request-Reply messages, each on its own channel.
> 
> Request-Reply has two participants:
> 
[...]

PR #992, opened 2023-11-16
docs: provide reply definition

Tenischev:
---
title: "Provide more solid definition of the reply"
---

This PR adds definition of the reply message and fix a drift of naming response-reply in the spec.

---

**Related issue(s):**
Resolves: #980

---


Tenischev:
@derberg as requested 😄

GreenRover:
Change makes sense for me.

smoya (commented):
LGTM @Tenischev . Would you mind fixing the conflicts so I can +1 this?

Issue #991, opened 2023-11-16
Message referencing from operation needs to be more strict


derberg:
## Context

- Messages in operation
![Screenshot 2023-11-16 at 18 32 44](https://github.com/asyncapi/spec/assets/6995927/d2ad5569-b53e-42ff-9b79-67cbfdd9f59f)

- Messages in operation reply
![Screenshot 2023-11-16 at 18 33 29](https://github.com/asyncapi/spec/assets/6995927/ecddf219-0e0a-48ef-87f0-bad25c91f5eb)

## Key of the problem

```
It MUST contain a subset of the messages defined in the [channel referenced in this operation](https://github.com/asyncapi/spec/blob/next-major-spec/spec/asyncapi.md#operationObjectChannel)
```

## Confusion

but does that mean that the reference must point to a message through channel location, or can be also `components`?

for me this should be enforced:
```yaml
asyncapi: '3.0.0'
info:
  title: Account Service
  version: 1.0.0
  description: This service is in charge of processing user signups
channels:
  UserSignedUp: 
    messages:
      UserSignedUp:
        $ref: '#/components/messages/UserSigned'
operations:
  user/signedup:
    action: send
    channel: 
      $ref: '#/channels/UserSignedUp'
    messages:
      - $ref: '#/channels/UserSignedUp/messages/UserSignedUp'
components:
  messages:
    UserSigned:
      payload:
        type: object
        properties:
          displayName:
            type: string
            description: Name of the user
          email:
            type: string
            format: email
            description: Email of the user
```

this part
```
channel: 
      $ref: '#/channels/UserSignedUp'
    messages:
      - $ref: '#/channels/UserSignedUp/messages/UserSignedUp'
 ```

and people should not be allowed to do 
```
channel: 
      $ref: '#/channels/UserSignedUp'
    messages:
      - $ref: '#/components/messages/UserSigned'
```

otherwise we'll get into a trouble of mismatch, unless I'm wrong and `messageId` is not considered to be a part of message

take this v2 example
```
asyncapi: '2.6.0'
info:
  title: Account Service
  version: 1.0.0
  description: This service is in charge of processing user signups
channels:
  user/signedup:
    subscribe:
      message:
        $ref: '#/components/messages/UserSignedUp'
components:
  messages:
    UserSignedUp:
      payload:
        type: object
        properties:
          displayName:
            type: string
            description: Name of the user
          email:
            type: string
            format: email
            description: Email of the user
```

converter will convert it to
```
[...]

PR #996, opened 2023-11-28
fix: clarify pointers for channel.servers and operation.channel fields

smoya:
Fixes https://github.com/asyncapi/spec/issues/991

Please help me writing proper sentences. Feel free to suggest changes.


spec/asyncapi.md
```diff
@@ -634,7 +634,7 @@ Field Name | Type | Description
...
+<a name="channelObjectServers"></a>servers | [[Reference Object](#referenceObject)] | An array of `$ref` pointers to the definition of the servers in which this channel is available. It MUST point to a subset of server definitions located in the [Servers Object](#serversObject) and MUST NOT point to a subset of server definitions located in the [Components Object](#componentsObject). If `servers` is absent or empty, this channel MUST be available on all the servers defined in the [Servers Object](#serversObject). Please note the `servers` property value MUST be an array of [Reference Objects](#referenceObject) and, therefore, MUST NOT contain an array of [Server Objects](#serverObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience.
```
fmvilas, line 637: Why not making it more explicit?

```suggestion
<a name="channelObjectServers"></a>servers | [[Reference Object](#referenceObject)] | An array of `$ref` pointers to the definition of the servers in which this channel is available. This value MUST start with `#/servers/`, meaning only pointing to the [root servers object](#serversObject) is allowed. If `servers` is absent or empty, this channel MUST be available on all the servers defined in the [Servers Object](#serversObject). Please note the `servers` property value MUST be an array of [Reference Objects](#referenceObject) and, therefore, MUST NOT contain an array of [Server Objects](#serverObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience.
```

spec/asyncapi.md
```diff
@@ -830,7 +830,7 @@ Describes a specific operation.
...
-<a name="operationObjectChannel"></a>channel | [Reference Object](#referenceObject) | **Required**. A `$ref` pointer to the definition of the channel in which this operation is performed. Please note the `channel` property value MUST be a [Reference Object](#referenceObject) and, therefore, MUST NOT contain a [Channel Object](#channelObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience.
+<a name="operationObjectChannel"></a>channel | [Reference Object](#referenceObject) | **Required**. A `$ref` pointer to the definition of the channel in which this operation is performed. Please note the `channel` property value MUST be a [Reference Object](#referenceObject) and, therefore, MUST NOT contain a [Channel Object](#channelObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience. It MUST point to a channel definition located in the [Channels Object](#channelsObject) and MUST NOT point to a channel definition located in the [Components Object](#componentsObject).
```
fmvilas, line 833: Same as with servers.

fmvilas (changes requested):

spec/asyncapi.md
derberg, line 637:
```suggestion
<a name="channelObjectServers"></a>servers | [[Reference Object](#referenceObject)] | An array of `$ref` pointers to the definition of the servers in which this channel is available. It MUST point to a subset of server definitions located in the [root Servers Object](#serversObject) and MUST NOT point to a subset of server definitions located in the [Components Object](#componentsObject). If `servers` is absent or empty, this channel MUST be available on all the servers defined in the [Servers Object](#serversObject). Please note the `servers` property value MUST be an array of [Reference Objects](#referenceObject) and, therefore, MUST NOT contain an array of [Server Objects](#serverObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience.
```

might make sense to be explicit with `root` to make it easier to understand without clicking link and understanding it points to a definition of a root element

derberg: you cannot be explicit, you can have channel defined in a separate file and then the $ref to server would not start with `#/servers`

derberg (changes requested):
you also need to update definition of channel reference in Operation Reply Object

and yeah, same with message -> https://github.com/asyncapi/spec/issues/991#issuecomment-1830022480

derberg:
examples fix https://github.com/asyncapi/spec/pull/997

jonaslagoni (commented):
I have some use cases I want to run by you and you let me know which rules are applied.

- Use case 1:  A root channel is defined as a reference, and the operation references that root channel. How do you define the message reference? To the root channel or the channel in the components section? 

```
{
    asyncapi: 3.0.0,
    channels: {
        test: {
            $ref: '#/components/channels/test'
        }
    },
    operations: {
        operationTest: {
            channel: {
                $ref: '#/channels/test'
            },
            messages: [
                {
                    $ref: '#/channels/test/messages/test'
                }
            ]
        }
    },
    components: {
        channels: {
            test: {
                messages: {
                    test: {},
                    test2: {}
                }
            }
        }
    }
}
```

- Use case 2:  You use AsyncAPI to define a menu of channels and operations. Are operations allowed to reference channels in components? 
```
{
    asyncapi: 3.0.0,
    components: {
        channels: {
            test: { ... }
        },
        operations: {
            operationTest: {
                channel: {
                    $ref: '#/components/channels/test'
                }
            }
        }
    }
}
```

spec/asyncapi.md
smoya: Exactly. Technically it should contain the string `#/servers` but not necessary in the beginning of the pointer. In fact that's what the Spectral rule checks https://github.com/asyncapi/parser-js/blob/next-major-spec/src/ruleset/v3/ruleset.ts#L56 but you @fmvilas already know because you reviewed my PR and, in fact, you suggested it iirc.

smoya:
> I have some use cases I want to run by you and you let me know which rules are applied.
> 
> * Use case 1:  A root channel is defined as a reference, and the operation references that root channel. How do you define the message reference? To the root channel or the channel in the components section?

Technically, your example is the right way. 
User friendly? Might sound a bit confusing indeed. In addition, I believe confusion increases due to the usage we are giving to JSON Reference in these cases (channels.servers, operations.channel, operations.messages), where they represent just "an ID" of the referenced object VS the rest of the document which are indeed JSON References that are expected to be dereferenced. In fact, the [current description in v3](https://github.com/asyncapi/spec/blob/next-major-spec/spec/asyncapi.md?plain=1#L833) for `operation.channel` field says:

>  A `$ref` pointer to the definition of the channel in which this operation is performed. Please note the `channel` property value MUST be a [Reference Object](#referenceObject) and, therefore, MUST NOT contain a [Channel Object](#channelObject).

I would not allow, in this case, a ref to the message in components even though it is the same referenced in the linked channel because once everything gets dereferenced, those channels.servers, operations.channel, operations.messages will still be presented as references.
We could, however, improve our Spectral rule to detect these cases and, instead of just throwing the current error, be more user friendly specifying that even though they both point to the same message, they need to point to the root object. Not strong opinion and not very willing to do that but 🤷 .

> * Use case 2:  You use AsyncAPI to define a menu of channels and operations. Are operations allowed to reference channels in components?

Operations inside components can reference to either channel or messages from components or from root.

smoya:
@jonaslagoni I think I managed to clarify a bit the use cases you showed in [your comment](https://github.com/asyncapi/spec/pull/996#pullrequestreview-1754090870). 

Also I improved the wording and added clarifications in operation reply object as well. Please all of you take a look 🙏 @fmvilas @derberg

spec/asyncapi.md
fmvilas: You're right. So then the parser can't implement it this way either, right? @smoya

spec/asyncapi.md
```diff
@@ -634,7 +634,7 @@ Field Name | Type | Description
...
+<a name="channelObjectServers"></a>servers | [[Reference Object](#referenceObject)] | An array of `$ref` pointers to the definition of the servers in which this channel is available. If the channel is located in the [root Channels Object](#channelsObject), it MUST point to a subset of server definitions located in the [root Servers Object](#serversObject), and MUST NOT point to a subset of server definitions located in the [Components Object](#componentsObject). If the channel is located in the [Components Object](#componentsObject), it MAY additionally point to a subset of server definitions located in the [Components Object](#componentsObject). If `servers` is absent or empty, this channel MUST be available on all the servers defined in the [Servers Object](#serversObject). Please note the `servers` property value MUST be an array of [Reference Objects](#referenceObject) and, therefore, MUST NOT contain an array of [Server Objects](#serverObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience.
```
fmvilas, line 637:
```suggestion
<a name="channelObjectServers"></a>servers | [[Reference Object](#referenceObject)] | An array of `$ref` pointers to the definition of the servers in which this channel is available. If the channel is located in the [root Channels Object](#channelsObject), it MUST point to a subset of server definitions located in the [root Servers Object](#serversObject), and MUST NOT point to a subset of server definitions located in the [Components Object](#componentsObject) or anywhere else. If the channel is located in the [Components Object](#componentsObject), it MAY point to a [Server Objects](#serverObject) in any location. If `servers` is absent or empty, this channel MUST be available on all the servers defined in the [Servers Object](#serversObject). Please note the `servers` property value MUST be an array of [Reference Objects](#referenceObject) and, therefore, MUST NOT contain an array of [Server Objects](#serverObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience.
```

spec/asyncapi.md
smoya: > You're right. So then the parser can't implement it this way either, right? @smoya

But the rule only applies to objects (Channels, Operations...) defined in the root. See https://github.com/asyncapi/parser-js/blob/next-major-spec/src/ruleset/v3/ruleset.ts#L51

spec/asyncapi.md
```diff
@@ -830,7 +830,7 @@ Describes a specific operation.
...
+<a name="operationObjectChannel"></a>channel | [Reference Object](#referenceObject) | **Required**. A `$ref` pointer to the definition of the channel in which this operation is performed. If the operation is located in the [root Channels Object](#channelsObject), it MUST point to a channel definition located in the [root Channels Object](#channelsObject), and MUST NOT point to a channel definition located in the [Components Object](#componentsObject). If the operation is located in the [Components Object](#componentsObject), it MAY additionally point to a channel definition located in the [Components Object](#componentsObject). Please note the `channel` property value MUST be a [Reference Object](#referenceObject) and, therefore, MUST NOT contain a [Channel Object](#channelObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience.
```
fmvilas, line 833:
```suggestion
<a name="operationObjectChannel"></a>channel | [Reference Object](#referenceObject) | **Required**. A `$ref` pointer to the definition of the channel in which this operation is performed. If the operation is located in the [root Operations Object](#operationsObject), it MUST point to a channel definition located in the [root Channels Object](#channelsObject), and MUST NOT point to a channel definition located in the [Components Object](#componentsObject) or anywhere else. If the operation is located in the [Components Object](#componentsObject), it MAY point to a [Channel Object](#channelObject) in any location. Please note the `channel` property value MUST be a [Reference Object](#referenceObject) and, therefore, MUST NOT contain a [Channel Object](#channelObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience.
```

spec/asyncapi.md
```diff
@@ -839,7 +839,7 @@ Field Name | Type | Description
...
+<a name="operationObjectMessages"></a>messages | [[Reference Object](#referenceObject)] | A list of `$ref` pointers pointing to the supported [Message Objects](#messageObject) that can be processed by this operation. It MUST contain a subset of the messages defined in the [channel referenced in this operation](#operationObjectChannel), and MUST NOT point to a subset of message definitions located in the [Messages Object](#componentsMessages) in the [Components Object](#componentsObject). **Every message processed by this operation MUST be valid against one, and only one, of the [message objects](#messageObject) referenced in this list.** Please note the `messages` property value MUST be a list of [Reference Objects](#referenceObject) and, therefore, MUST NOT contain [Message Objects](#messageObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience.
```
fmvilas, line 842:
```suggestion
<a name="operationObjectMessages"></a>messages | [[Reference Object](#referenceObject)] | A list of `$ref` pointers pointing to the supported [Message Objects](#messageObject) that can be processed by this operation. It MUST contain a subset of the messages defined in the [channel referenced in this operation](#operationObjectChannel), and MUST NOT point to a subset of message definitions located in the [Messages Object](#componentsMessages) in the [Components Object](#componentsObject) or anywhere else. **Every message processed by this operation MUST be valid against one, and only one, of the [message objects](#messageObject) referenced in this list.** Please note the `messages` property value MUST be a list of [Reference Objects](#referenceObject) and, therefore, MUST NOT contain [Message Objects](#messageObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience.
```

spec/asyncapi.md
```diff
@@ -977,8 +977,8 @@ Describes the reply part that MAY be applied to an Operation Object. If an opera
...
...nnel"></a>channel | [Reference Object](#referenceObject) | A `$ref` pointer to the definition of the channel in which this operation is performed. When [address](#operationReplyAddressObject) is specified, the [`address` property](#channelObjectAddress) of the channel referenced by this property MUST be either `null` or not defined. If the operation reply is located inside a [root Operation Object](#operationObject), it MUST point to a channel definition located in the [root Channels Object](#channelsObject), and MUST NOT point to a channel definition located in the [Components Object](#componentsObject). If the operation reply is located inside an [Operation Object] in the [Components Object](#componentsObject) or in the [Replies Object](#componentsReplies) in the [Components Object](#componentsObject), it MAY additionally point to a channel definition located in the [Components Object](#componentsObject). Please note the `channel` property value MUST be a [Reference Object](#referenceObject) and, therefore, MUST NOT contain a [Channel Object](#channelObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience.
```
fmvilas, line 980:
```suggestion
<a name="operationReplyObjectChannel"></a>channel | [Reference Object](#referenceObject) | A `$ref` pointer to the definition of the channel in which this operation is performed. When [address](#operationReplyAddressObject) is specified, the [`address` property](#channelObjectAddress) of the channel referenced by this property MUST be either `null` or not defined. If the operation reply is located inside a [root Operation Object](#operationObject), it MUST point to a channel definition located in the [root Channels Object](#channelsObject), and MUST NOT point to a channel definition located in the [Components Object](#componentsObject) or anywhere else. If the operation reply is located inside an [Operation Object] in the [Components Object](#componentsObject) or in the [Replies Object](#componentsReplies) in the [Components Object](#componentsObject), it MAY point to a [Channel Object](#channelObject) in any location. Please note the `channel` property value MUST be a [Reference Object](#referenceObject) and, therefore, MUST NOT contain a [Channel Object](#channelObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience.
```

spec/asyncapi.md
```diff
@@ -977,8 +977,8 @@ Describes the reply part that MAY be applied to an Operation Object. If an opera
...
+<a name="operationReplyObjectMessages"></a>messages | [[Reference Object](#referenceObject)] | A list of `$ref` pointers pointing to the supported [Message Objects](#messageObject) that can be processed by this operation as reply. It MUST contain a subset of the messages defined in the [channel referenced in this operation reply](#operationObjectChannel), and MUST NOT point to a subset of message definitions located in the [Components Object](#componentsObject) or to a subset of message definitions defined in a [Channel Object](#channelObject) in the [Components Object](#componentsObject). **Every message processed by this operation MUST be valid against one, and only one, of the [message objects](#messageObject) referenced in this list.** Please note the `messages` property value MUST be a list of [Reference Objects](#referenceObject) and, therefore, MUST NOT contain [Message Objects](#messageObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience.
```
fmvilas, line 981:
```suggestion
<a name="operationReplyObjectMessages"></a>messages | [[Reference Object](#referenceObject)] | A list of `$ref` pointers pointing to the supported [Message Objects](#messageObject) that can be processed by this operation as reply. It MUST contain a subset of the messages defined in the [channel referenced in this operation reply](#operationObjectChannel), and MUST NOT point to a subset of message definitions located in the [Components Object](#componentsObject) or anywhere else. **Every message processed by this operation MUST be valid against one, and only one, of the [message objects](#messageObject) referenced in this list.** Please note the `messages` property value MUST be a list of [Reference Objects](#referenceObject) and, therefore, MUST NOT contain [Message Objects](#messageObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience.
```

fmvilas (changes requested):

spec/asyncapi.md
fmvilas: Hmmm 🤔 I still don't think it's correct. Unless we avoid the parser to resolve these references.

smoya: The Spectral rule applies before resolving the references. Not sure if this "answers" your question.

smoya:
Even though this has been merged now, can you @jonaslagoni confirm your concerns about thos edge cases are now clarified with my changes added in the spec? thank you! 🙏

jonaslagoni:
Gonna give it another read when I am doing the converter update

Issue #993, opened 2023-11-17
Add definition for Operation to spec 3.0


buehlefs:
Since version 3 elevates operations to a top level component (and spec v3 Operations have a different semantic compared to 2.x) they should have a dedicated definition in the [definitions](https://github.com/asyncapi/spec/blob/next-major-spec/spec/asyncapi.md#definitions) section. (Additionally, operations could also be added to the [concepts](https://www.asyncapi.com/docs/concepts))

I do not have a fleshed out proposal yet (I may create a PR if I find the time to do so), so feel free to propose a definition or create a PR.

Suggested starting point:

> An operation describes a specific action an application can take to interact with the async api.
> Operations are performed by sending (or receiving) messages to (or from) channels.

Related Issues:
- #980
- #618 Issue about the changes of operations

derberg:
Hey, yes, you're right, we need it. I think your text is ok.

@smoya @fmvilas @char0n @dalelane thoughts?

PR #998, opened 2023-11-30
docs: add operation definition for spec 3.0

buehlefs:
---
title: "Add operation definition for spec 3.0"
---

This PR adds the operation definition from #993 to the spec document.

The second commit adds links to the new definition in other parts of the spec.
The third rewords the description of message objects to better align with the new definition of operations in the spec and adds links to the relevant definitions.

---

**Related issue(s):** #993 

---

### DIFF
--- a/scripts/converter/package-lock.json
+++ b/scripts/converter/package-lock.json
@@ -6,16 +6,16 @@
     "": {
       "name": "converter",
       "dependencies": {
-        "@asyncapi/converter": "1.3.1",
+        "@asyncapi/converter": "^1.4.6",
         "js-yaml": "^4.1.0"
       }
     },
     "node_modules/@asyncapi/converter": {
-      "version": "1.3.1",
-      "resolved": "https://registry.npmjs.org/@asyncapi/converter/-/converter-1.3.1.tgz",
-      "integrity": "sha512-aLbJTGaaqP6IHc+ggi/feQ8oynP/2iEd4S0JuSyWZFSYSR2uAXsDH9FrAEFGPrhD9inbc+p0fTshnhKLNGdJQw==",
+      "version": "1.4.6",
+      "resolved": "https://registry.npmjs.org/@asyncapi/converter/-/converter-1.4.6.tgz",
+      "integrity": "sha512-D0WVqA2bwZNmcDCm9eT4RIt7CCD3uLg0XMYbujz3o2llnmmvaat6wo7xE9eh9N/KufXr3EAn7xOZojP0yCEkDw==",
       "dependencies": {
-        "@asyncapi/parser": "^2.0.0-next-major.11",
+        "@asyncapi/parser": "^3.0.0",
         "js-yaml": "^3.14.1"
       }
     },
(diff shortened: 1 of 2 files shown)


spec/asyncapi.md
```diff
@@ -135,11 +135,14 @@ A sender is a type of application, that is sending [messages](#definitionsMessag
 ### <a name="definitionsReceiver"></a>Receiver
 A receiver is a type of application that is receiving [messages](#definitionsMessage) from [channels](#definitionsChannel). A receiver MAY receive from multiple channels depending on the [server](#definitionsServer), protocol, and the use-case pattern. A receiver MAY forward a received message further without changing it. A receiver MAY act as a consumer and react to the message. A receiver MAY act as a processor that, for example, aggregates multiple messages in one and forwards them.
 
+### <a name="definitionsOperation"></a>Operation
+An operation describes a specific action an [application](#definitionsApplication) can take to interact with the message-driven API. Operations are performed by sending (or receiving) [messages](#definitionsMessage) to (or from) [channels](#definitionsChannel).
```
fmvilas, line 139:
```suggestion
An operation describes a specific action an [application](#definitionsApplication) can take to interact with the [server](#definitionsServer). Operations are performed by sending (or receiving) [messages](#definitionsMessage) to (or from) [channels](#definitionsChannel).
```

fmvilas (changes requested):
Just a small suggestion.

spec/asyncapi.md
buehlefs: An operation in spec 3 is something like `userLogin`, `orderGoods`, or `sendPaymentInfo`. All of these make sense when viewed as an operation of the message-driven API defined in an asyncapi document but not when viewed as an operation on, e.g., a kafka server that is only a part of the API implementation with the role of relaying and storing messages.
The old 2.x operations were operations like `send` and `receive` that were specifically tied to a channel/server. I am against associating the new spec 3 operation with servers, as this would make this difference harder to spot at a glance.

fmvilas: Would that work for you instead?

```suggestion
An operation describes a specific action an [application](#definitionsApplication) can take to interact with a [channel](#definitionsChannel). Operations are performed by sending (or receiving) [messages](#definitionsMessage) to (or from) [channels](#definitionsChannel).
```

I don't want to even mention the words message-driven and API because they're not defined anywhere, and their meaning is already bloated. I'm actually getting rid of the word API as much as possible because in the message-driven world, people don't perceive it as an API (that's often associated with HTTP).

Also, AsyncAPI doesn't only serve the purpose for message-driven systems. Note that we also support HTTP interactions (even more with v3). Even though, theoretically, HTTP APIs could be broken down to message exchange interactions.

fmvilas: Alright. For some reason I thought we already removed all mentions of message-driven API but I can see we didn't. Good to go then 👍

fmvilas (approved):

buehlefs:
I will probably create a new PR to resolve the merge conflicts...

Issue #1031, opened 2024-02-12
Possible bug with example - adeo-kafka-request-reply-asyncapi


Pakisan:
Hi! 

I'm not sure, but looks like this example is broken - https://github.com/asyncapi/spec/blob/master/examples/adeo-kafka-request-reply-asyncapi.yml

I checked It in Studio - no warnings, but when I tried to parse this example using JAsyncAPI I enfaced with errors

I found two strange things:

1. extensions

```
bindings:
      kafka:
        x-key.subject.name.strategy:
          type: string
          description: >
            We use the RecordNameStrategy to infer the messages schema. Use
            `key.subject.name.strategy=io.confluent.kafka.serializers.subject.RecordNameStrategy`
            in your consumer configuration.
        x-value.subject.name.strategy:
          type: string
          description: >
            We use the RecordNameStrategy to infer the messages schema. Use
            `value.subject.name.strategy=io.confluent.kafka.serializers.subject.RecordNameStrategy`
            in your consumer configuration.
```

This property name `x-key.subject.name.strategy` must not be interpreted as valid because of [our regex](https://www.asyncapi.com/docs/reference/specification/v3.0.0#specificationExtensions) `^x-[\w\d\-\_]+$` or not?

2. Schema structure

```
CorrelationId:
      type: string
      format: uuid
      description: >-
        A unique Correlation ID defined from the `REQUEST_ID` or the
        `MESSAGE_ID` provided in the Costing Request.
      example: 1fa6ef40-8f47-40a8-8cf6-f8607d0066ef
```

I may be wrong, but [draft-07](https://json-schema.org/draft-07/draft-handrews-json-schema-validation-01#rfc.section.8.5) doesn't have such field, only examples or not?

Pakisan:
@dalelane @smoya @char0n @GreenRover

smoya:
Regarding point 1, I see the regex we show in [our documentation](https://www.asyncapi.com/docs/reference/specification/v3.0.0#specificationExtensions) doesn't correspond with the one enforced in our [JSON Schema documents](https://github.com/asyncapi/spec-json-schemas/blob/master/schemas/3.0.0.json#L4700-L4704), which indeed allows for `.` characters. So I completely understand then why Studio is not complaining, but I don't get why ** JAsyncAPI** doesn't, tbh.
So in summary, we should either fix the docs or fix the regex. A decision should be taken.

Regarding point 2, yes you are right: there is no `example` field but `examples`.

[...]

PR #1034, opened 2024-02-14
docs: fix adeo-kafka-request-reply-asyncapi.yml

Pakisan:
Replace `example` with `examples`

Reference: https://github.com/asyncapi/spec/issues/1031


smoya (commented):
LGTM! 🚀🌔

smoya (approved):

PR #1046, opened 2024-03-19
ci: add a script and CI job to validate spec examples

### LINKED ISSUE #957: Add a script and CI job to validate spec examples
This feature request was discussed here: https://github.com/asyncapi/spec/pull/947#discussion_r1238603958

We need to make sure all the examples are complete and valid according to our spec. This should of course be checked for each PR that change the examples.

AnimeshKumar923:
This PR adds a script and CI to validate the AsyncAPI documents in the `/examples` directory.
- The embedded examples in spec are to be validated through it.
- Additionally, if any incoming PR is changing the spec examples, it should check that also,

## Tasks

- [x] Write a script to validate the existing AsyncAPI documents examples under the `/examples` directory.
- [x] Develop a YAML file for the CI. _(WIP)_
- [ ] Write a script to validate the AsyncAPI documents containing embedded examples. --> Will be done through a separate PR ([context](https://github.com/asyncapi/spec/pull/1046#pullrequestreview-1991282392)) (phase-2)

cc: @smoya PTAL

--- 
**Related issue(s):**
#957 (partially resolves)


scripts/validation/validate-examples.js
```diff
@@ -0,0 +1,35 @@
+const { execSync } = require('child_process');
+const glob = require('glob');
+
+// Use glob to find all YAML files in the examples directory
+const files = glob.sync('./examples/**/*.{yml,yaml}');
```
smoya, line 5: I believe there are some files you won't be able to validate since they are not AsyncAPI documents such as the files located in https://github.com/asyncapi/spec/tree/master/examples/social-media/common. Not sure if there are more, you will need to check.

scripts/validation/validate-examples.js
```diff
@@ -0,0 +1,35 @@
...
+files.forEach((file) => {
+  filesCount++;
+  try {
+    console.log(`\nValidating: ${file}`);
+    execSync(`npx asyncapi validate ${file}`, { stdio: 'inherit' });
+  } catch (error) {
+    console.error(`Validation failed for: ${file}\n`);
+    errorFilesCount++;
+    filesWithErrors.push(file);
+    // process.exit(1);
```
smoya, line 20: This should be removed. Instead, the process should exit with code 1 after printing all validation errors in https://github.com/asyncapi/spec/pull/1046/files?diff=unified&w=0#diff-db990a3ee807308932214a7f5f1e6e3c279959f27c1c5449d0dfce6e440edcddR32

scripts/validation/validate-examples.js
```diff
@@ -0,0 +1,35 @@
...
+
+let filesCount = 0;
+let errorFilesCount = 0;
+let filesWithErrors = []; // Array to store files that failed validation
+// Validate each file using AsyncAPI CLI
+files.forEach((file) => {
+  filesCount++;
+  try {
+    console.log(`\nValidating: ${file}`);
+    execSync(`npx asyncapi validate ${file}`, { stdio: 'inherit' });
```
smoya, line 15: This is gonna be pretty slow. Can we run it async somehow and compare performance? 🤔

scripts/validation/validate-examples.js
AnimeshKumar923: Yes, it is like that. 
You can see the output:

image

So, what can be done about it? Exclude these? Or something else?

scripts/validation/validate-examples.js
AnimeshKumar923: resolved with abbda0d

scripts/validation/validate-examples.js
AnimeshKumar923: > This is gonna be pretty slow. Can we run it async somehow and compare performance? 🤔

Okay, let me try doing it asynchronously...

AnimeshKumar923: Applied suggestion through baa7c12

AnimeshKumar923: # Performance Report

## Before
When the script was executed _**synchronously**_:
- [Job link](https://github.com/AnimeshKumar923/asyncapi-spec/actions/runs/8410069994/usage)
- _Screenshot_
image

---
## After
When the script was executed _**asynchronously**_:
- [Job link](https://github.com/AnimeshKumar923/asyncapi-spec/actions/runs/8410126555/usage)
- _Screenshot_
image
---

Conclusion: The execution time here has been **reduced by 1m 10s** with the use of asynchronous way of JavsScript.

Thanks for the suggestion! @smoya :+1:

AnimeshKumar923: ## The execution time for the script itself

### Before
image

### After
image

scripts/validation/validate-examples.js
smoya: Exclude them, yes.

scripts/validation/validate-examples.js
AnimeshKumar923: > I think we can go further and skip using a JS script here and find a simple online terminal command that do the job.
> 
> I believe by using `find` which also allows excluding paths + piping to `asyncapi` command (CLI) through `xargs` would end up being a really simple oneliner, not complicated and simple to read and maintain.

Let me look into that...

scripts/validation/validate-examples.js
AnimeshKumar923: > Exclude them, yes.

Got it!

scripts/validation/validate-examples.js
AnimeshKumar923: @smoya The code has been successfully executed. Please have a look at this [workflow job detail](https://github.com/AnimeshKumar923/asyncapi-spec/actions/runs/8573974272/job/23499821262)

---
### Attaching screenshot for more details:
image

---
### Job run details
image
_[Link for reference](https://github.com/AnimeshKumar923/asyncapi-spec/actions/runs/8573974272/usage)_

.github/workflows/validate-examples.yml
```diff
@@ -0,0 +1,27 @@
...
+        uses: actions/checkout@v3
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: '20'
+      - name: Install AsyncAPI CLI
+        run: npm install -g @asyncapi/cli
+      - name: Validate AsyncAPI documents
+        run: |
+          find examples/ \( -path 'examples/social-media/*' -prune \) -o -type f \( -name "*.yml" -o -name "*.yaml" \) -exec asyncapi validate {} \;
```
smoya, line 27: Nice oneliner 👏 👏 

There is only a couple of issues here:

1. The path to skip should be `'examples/social-media/common`.
2. I don't find `prune` option to be very readable. Instead, you can use the `-not` operator. E.g. `-not -path 'examples/social-media/common'`
3. With `-exec` option you can't parallelise the work, so everything will run sequentially. 
Instead, try piping the find output to `xargs`. 

Applying that, will look like:

```sh
find examples -type f \( -name "*.yml" -o -name "*.yaml" \) -not -path 'examples/social-media/common' | xargs -P 10 -L 1 asyncapi validate
```

smoya: Additionally, please add a comment on top of the command explaining why we are skipping that dir.

AnimeshKumar923: > Nice oneliner 👏 👏

Thank you! :grin: To be honest, I'm still learning bash so somethings may take time to understand :sweat_smile: 

> There is only a couple of issues here:
> 
>     1. The path to skip should be `'examples/social-media/common`.
> 
>     2. I don't find `prune` option to be very readable. Instead, you can use the `-not` operator. E.g. `-not -path 'examples/social-media/common'`
> 
>     3. With `-exec` option you can't parallelise the work, so everything will run sequentially.
>        Instead, try piping the find output to `xargs` and the `-P` arg.
> 
> 
> Applying that, will look like:
> 
> ```shell
> find examples -type f \( -name "*.yml" -o -name "*.yaml" \) -not -path 'examples/social-media/common' | xargs -P 10 -L 1 asyncapi validate
> ```

Thank you for the suggestions, will try to understand and implement the same. :pray:

AnimeshKumar923: > Additionally, please add a comment on top of the command explaining why we are skipping that dir.

Sure, I'll do that.

smoya (commented):
On my side, this has my 👍 (pending to switch the branch from `script-ci-spec-validation` to `master` in the workflow).
Additinoally, if we want to merge this PR before continuing the work for ` Write a script to validate the AsyncAPI documents containing embedded examples.` which I believe we should, please @AnimeshKumar923 remove the sentence `Fixes https://github.com/asyncapi/spec/issues/957` from the description so the issue remains open after merging. 

Can you @derberg do a quick check on the workflow just in case I'm missing something critical? 🙏

.github/workflows/validate-examples.yml
```diff
@@ -0,0 +1,28 @@
+name: Validate AsyncAPI Examples
+
+on:
+  push:
```
derberg, line 4: remember to remove

.github/workflows/validate-examples.yml
```diff
@@ -0,0 +1,28 @@
...
+    branches:
+      - script-ci-spec-validation # added temporarily for testing purpose
+      # - master
+
+  pull_request_target:
```
derberg, line 9: there are no complex things that need GH token so you can use "pull_request" event. This way will be easier to work with workflow as changes in it will affect the PR where change is done, no need to merge first

.github/workflows/validate-examples.yml
```diff
@@ -0,0 +1,28 @@
...
+        uses: actions/checkout@v4
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+      - name: Install AsyncAPI CLI
+        run: npm install -g @asyncapi/cli
+      - name: Validate AsyncAPI documents
+        run: |
+          find examples -type f \( -name "*.yml" -o -name "*.yaml" \) -not -path 'examples/social-media/common/*' | xargs -P 10 -L 1 asyncapi validate
```
derberg, line 27: omg, bash on steroids

luckily we have chat gpt to translate it for me when needed 😁

.github/workflows/validate-examples.yml
```diff
@@ -0,0 +1,28 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+      - name: Install AsyncAPI CLI
+        run: npm install -g @asyncapi/cli
```
derberg, line 24: bo it uses latest CLI - advantage is that all latest features will be used but obvious disadvantage is randomly failing CI cause of some breaking change in validation, or some new validation enabled.

yeah, not so obvious in this case 😕

derberg (commented):
I changed PR prefix cause otherwise we would release new spec version 😁

also, one question, if we use AsyncAPI CLI on each PR - what if we do a PR that adds new features for new release, and updates examples 🤔

.github/workflows/validate-examples.yml
AnimeshKumar923: > omg, bash on steroids
> 
> luckily we have chat gpt to translate it for me when needed 😁

:laughing:

.github/workflows/validate-examples.yml
AnimeshKumar923: > remember to remove

Yes.

.github/workflows/validate-examples.yml
AnimeshKumar923: so should be use a specific version of the CLI? :eyes:

.github/workflows/validate-examples.yml
AnimeshKumar923: implemented through b49f4a1

.github/workflows/validate-examples.yml
smoya: Come on! are self explanatory 😆 Let me save your time

```sh
find examples -type f \( -name "*.yml" -o -name "*.yaml" \) -not -path 'examples/social-media/common/*' 
```

This is completely human readable: find in examples dir "things" of type file (f) with name suffixes `yml` or `yaml` and do not include the path  'examples/social-media/common/*' . The output is a list of paths to those files.

```sh
xargs -P 10 -L 1 asyncapi validate 
```

Here maybe the trickiest part: xargs meaning do "x" with the piped arguments (which is the list of files), the -L 1 means take each new line as one execution, -P 10 means parallellize in 10 runs at the same time, and for each of those arguments (paths), execute `asyncapi validate` command.

.github/workflows/validate-examples.yml
smoya: If I can give my opinion here, I think using latest is a good option. We are always testing on the last version, supporting last spec schemas, etc + if something is wrong with the CLI, we are gonna be one of the first who notice.

aeworxet:
@asyncapi/bounty_team

.github/workflows/validate-examples.yml
smoya: ping @derberg  @AnimeshKumar923

smoya: Ping @derberg 👋

derberg: no strong opinion really, there are pros and cons for using latest and not using latest. We just need to be aware that if we decide to use latest - then on PRs we can have random fails from time to time and we need to support contributors

smoya: Agree. I suggest then we can try with this but opened to change if needed at any time.

smoya (approved):
LGTM!

PR #1059, opened 2024-06-07
ci: script to validate embedded examples in the asyncapi.md file

### LINKED ISSUE #957: Add a script and CI job to validate spec examples

AnimeshKumar923:
**Description**

- This PR adds a script and CI _(will added after the script has been finalized)_ to validate the embedded examples in the [asyncapi.md file](https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md)

---

**Remaining tasks:**
- [x] Script enhancement _(achieved through d78f639 --> finalized through 8c0e78c)_
- [x] Add a CI file integrating the script _(achieved through 46f32a5 --> finalized through aa1564f)_
---
**Related issue(s):**
Resolves #957


AnimeshKumar923:
### Current State of the script 

This is what the script currently does as of 915f808:
- This initial script replaces the section in the `ex-base-doc.json` by extracting the `json` section by taking the comment given in this format `<!-- asyncapi-example-tester:{"test":"Info Object","json_path":"$.info"} -->`.

  - Here `test` is the object which stores the name of the example _(will change the name of object later on for better naming convention)_
  - `json_path` is the object specifying the location where the example should be replaced.

### Future objective:
- [ ] Update asyncapi.md file with more improved comments
- [ ] Integrate the script with CLI validation
- [ ] Add a YAML file for the CI

aeworxet:
@asyncapi/bounty_team

spec/asyncapi.md
```diff
@@ -1867,7 +1867,40 @@ Name | Allowed values | Notes
...
+                "type": "int"
+              }
+            ]
+          }
+        }
+      }
+    }
+  }
+}
+```
```
AnimeshKumar923, line 1900: I've added some missing JSON examples while I was going through the document, PTAL @smoya @fmvilas @dalelane @char0n @GreenRover @derberg

spec/asyncapi.md
```diff
@@ -2676,7 +2709,27 @@ Traits MUST be merged with the target object using the [JSON Merge Patch](https:
...
+    {
+      "tags": [
+        {
+          "name": "user"
+        }
+      ]
+    }
+  ]
+}
+```
```
AnimeshKumar923, line 2730: 2nd JSON example, PTAL @smoya @fmvilas @dalelane @char0n @GreenRover @derberg

spec/asyncapi.md
```diff
@@ -2688,7 +2741,20 @@ traits:
...
+{
+  "name": "UserSignup",
+  "description": "A longer description.",
+  "tags": [
+    {
+      "name": "user"
+    }
+  ]
+}
+```
```
AnimeshKumar923, line 2755: 3rd example, PTAL @smoya @fmvilas @dalelane @char0n @GreenRover @derberg

spec/asyncapi.md
smoya: I would not add any new example in this PR but rather focus on the main issue that is testing current ones. If you want to add new tests later on, you can create a different PR. Otherwise we might be spending time reviewing the examples rather than what it matters in this case.

AnimeshKumar923: > I would not add any new example in this PR but rather focus on the main issue that is testing current ones. If you want to add new tests later on, you can create a different PR. Otherwise we might be spending time reviewing the examples rather than what it matters in this case.

@smoya 
I understand. I added these because the markdown comment was given for the YAML version and since I'm only considering the JSON ones for the validation, I thought these would be exempted from the validation therefore made this decision. 

I will revert it for now and will open a new PR addressing the JSON versions for these.

AnimeshKumar923: @smoya 
resolved through a51107d

spec/asyncapi.md
AnimeshKumar923: resolved through a51107d

spec/asyncapi.md
AnimeshKumar923: resolved through a51107d

Refer: https://github.com/asyncapi/spec/pull/1059#discussion_r1636431659

AnimeshKumar923:
## UPDATE
Hi @smoya :wave: 
- I have almost finished the [script](https://github.com/asyncapi/spec/blob/6f18cbc36767bbce74220947b9cd9fc570c21b62/scripts/validation/embedded-examples-validation.js) and the [CI file](https://github.com/asyncapi/spec/blob/6f18cbc36767bbce74220947b9cd9fc570c21b62/.github/workflows/validate-embedded-examples.yml). 
- Thought of using **parser-js** but chose **asyncapi-cli** due to more verbose error details. Please give your thoughts upon it.
- Ran [this test](https://github.com/AnimeshKumar923/asyncapi-spec/actions/runs/9538814183/job/26288449368) to see the working of the script and CI. It seems to be working fine as of now.
- Ran into some validation problems which I've sorted them out. I wrote them in a [separate document](https://github.com/AnimeshKumar923/test-repo/blob/main/embedded-examples-error-resolution.md).
- For [one of files](https://github.com/asyncapi/spec/blob/6f18cbc36767bbce74220947b9cd9fc570c21b62/scripts/validation/updated-docs/Security-Scheme-OAuth-Flow-Object.json), I'm unsure about what's going wrong so would require your help in debugging that.
  - [Alternate link](https://github.com/AnimeshKumar923/test-repo/blob/main/embedded-examples-error-resolution.md#problems-with-security-scheme-oauth-flow-objectjson) for more verbose error log

scripts/validation/embedded-examples-validation.js
```diff
@@ -0,0 +1,279 @@
...
+      console.error("Failed to parse comment JSON:", match[1], e);
+    }
+  }
+
+  return comments;
+}
+
+// Function to extract JSON examples from markdown content
+function extractExamples(content) {
+  const exampleRegex = /```json\s+([\s\S]*?)\s+```/g;
```
smoya, line 33: Wouldn't make sense to create one single regex that gives you the object inside the comment/annotation plus the example?

spec/asyncapi.md
```diff
@@ -273,6 +273,7 @@ This object MAY be extended with [Specification Extensions](#specificationExtens
 
 ##### Info Object Example
 
+<!-- asyncapi-example-tester:{"name":"Info Object","json_path":"$.info"} -->
```
smoya, line 276: How are we testing yaml version of the examples? You didn't add comment annotations

spec/asyncapi.md
smoya, line 1: Some examples are missing the annotation so you are not testing them. Example: https://github.com/asyncapi/spec/blob/9d83edf28cd6892a88c33a4e3762ab082e73a20b/spec/asyncapi.md?plain=1#L368-L372

scripts/validation/embedded-examples-validation.js
smoya: You can use the following regex that will give you 3 groups per match:

1. The raw JSON Object inside the comment, I.e. `{"name":"foo","json_path":"bar"}`
2. The format of the example (`json` | `yaml`)
3. The content of the example

scripts/validation/embedded-examples-validation.js
```diff
@@ -0,0 +1,279 @@
...
+    } else {
+      if (!current[part]) {
+        current[part] = {}; // Create object if it doesn't exist
+      }
+      current = current[part];
+    }
+  });
+}
+
+// Function to apply updates and save the document
```
smoya, line 101: Why should you save to disk?

scripts/validation/embedded-examples-validation.js
```diff
@@ -0,0 +1,279 @@
...
+  // console.log(`\n${combinedData[num-1].name} = ${currentExample}`);
+
+  // Apply updates and save the document
+  applyUpdatesAndSave([item], baseDocPath, outputPath);
+
+  // Validate the output file using the AsyncAPI parser
+  // const validationPromise = validateParser(outputPath);
+
+  // Use the AsyncAPI CLI to validate the output file
+  const validationPromise = validateCli(outputPath)
```
smoya, line 176: Parser should be enough. Bringing the CLI here is too much in terms of dependencies IMHO

spec/asyncapi.md
```diff
@@ -1371,6 +1388,7 @@ Field Name | Type | Description
 
 This object MAY be extended with [Specification Extensions](#specificationExtensions).
 
+<!-- asyncapi-example-tester:{"name":"Message Trait Object","json_path":"$.components.messageTraits.commonHeaders"} -->
```
smoya, line 1391: This shouldn't be in this line but just before \```json``` tag

scripts/validation/embedded-examples-validation.js
```diff
@@ -0,0 +1,279 @@
...
+  } catch (error) {
+    console.error(`Validation failed for ${filePath}: ${error.message}`);
+  }
+}
+
+// Function to validate a file using AsyncAPI CLI
+async function validateCli(filePath) {
+  return new Promise((resolve, reject) => {
+    // Construct the command to run the AsyncAPI CLI validate command
+    const command = `npx asyncapi validate ${filePath}`;
```
smoya, line 215: You should definitely use the parser directly and not call a cmd command here.

smoya:
> * Ran into some validation problems which I've sorted them out. I wrote them in a [separate document](https://github.com/AnimeshKumar923/test-repo/blob/main/embedded-examples-error-resolution.md).

I don't get the whole content of this file. When you mention lines of a document, to which document do you refer? It seems don't match with the spec file so I don't know which other file could be.

smoya:
> For [one of files](https://github.com/asyncapi/spec/blob/6f18cbc36767bbce74220947b9cd9fc570c21b62/scripts/validation/updated-docs/Security-Scheme-OAuth-Flow-Object.json), I'm unsure about what's going wrong so would require your help in debugging that.
> * [Alternate link](https://github.com/AnimeshKumar923/test-repo/blob/main/embedded-examples-error-resolution.md#problems-with-security-scheme-oauth-flow-objectjson) for more verbose error log

At a glance, seems like a bug in the JSON Schemas, but I'm investigating it

smoya:
> For [one of files](https://github.com/asyncapi/spec/blob/6f18cbc36767bbce74220947b9cd9fc570c21b62/scripts/validation/updated-docs/Security-Scheme-OAuth-Flow-Object.json), I'm unsure about what's going wrong so would require your help in debugging that.
> * [Alternate link](https://github.com/AnimeshKumar923/test-repo/blob/main/embedded-examples-error-resolution.md#problems-with-security-scheme-oauth-flow-objectjson) for more verbose error log

The issue is because `authorizationUrl` field **is **not** part of `clientCredentials` oauth flow, but `tokenUrl`. See:

Google Chrome_3DXZrb19
Source: https://www.asyncapi.com/docs/reference/specification/v3.0.0#oauthFlowObject

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: I'll try that!

spec/asyncapi.md
AnimeshKumar923: I initially started with json examples, so I removed the yaml comment. Will add it for testing also...

spec/asyncapi.md
AnimeshKumar923: I'll check the document for such examples.

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: > Why should you save to disk?

@smoya 
So I thought of this approach:
Saving temporarily --> validation of all examples --> print result --> delete the temp files

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: @smoya 
Okay, I'll add the validation through parser.
Thought of adding the CLI due to more verbose details of the error.

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: Got it.

spec/asyncapi.md
AnimeshKumar923: Got it. Will check it out.

AnimeshKumar923:
> > * Ran into some validation problems which I've sorted them out. I wrote them in a [separate document](https://github.com/AnimeshKumar923/test-repo/blob/main/embedded-examples-error-resolution.md).
> 
> I don't get the whole content of this file. When you mention lines of a document, to which document do you refer? It seems don't match with the spec file so I don't know which other file could be.

Ah, it's my shortcoming. I've updated the document with reference to the files being mention. Please have a look once again.

spec/asyncapi.md
AnimeshKumar923: resolved with ed55789

scripts/validation/embedded-examples-validation.js
smoya: But what's the value of writing to disk, if there is one?

scripts/validation/embedded-examples-validation.js
smoya: > Thought of adding the CLI due to more verbose details of the error.

Think about the cost of calling this through command line, meaning you are spinning up new processes on the machine, waiting for the process output, trusting npx binary is installed, etc, etc VS just using the `@asyncapi/parser` package. Additionally IMHO we don't really need such a verbosity.

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: > But what's the value of writing to disk, if there is one?

@smoya 
Let me explain a bit more.
- The parser validation will be [async in nature](https://github.com/asyncapi/spec/blob/ed557896c4a7d9dae334f4515efc3cbf6118c784/scripts/validation/embedded-examples-validation.js#L190C1-L210C1), so I thought it would be better to collect every example and write that down in a separate folder. 
- Then run the validation for that whole folder to get every example checked.

Do you have perhaps any other suggestions for this?

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: > > Thought of adding the CLI due to more verbose details of the error.
> 
> Think about the cost of calling this through command line, meaning you are spinning up new processes on the machine, waiting for the process output, trusting npx binary is installed, etc, etc VS just using the `@asyncapi/parser` package. Additionally IMHO we don't really need such a verbosity.

Got it! Thanks for the detailed explanation. Very much appreciate it. :grin:

smoya:
> > > * Ran into some validation problems which I've sorted them out. I wrote them in a [separate document](https://github.com/AnimeshKumar923/test-repo/blob/main/embedded-examples-error-resolution.md).
> > 
> > 
> > I don't get the whole content of this file. When you mention lines of a document, to which document do you refer? It seems don't match with the spec file so I don't know which other file could be.
> 
> Ah, it's my shortcoming. I've updated the [document](https://github.com/AnimeshKumar923/test-repo/blob/main/embedded-examples-error-resolution.md) with reference to the files being mentioned. Please have a look once again.

### [Operation-Object.json](https://github.com/asyncapi/spec/blob/ced626f6f6abf80e128216793a1bdc656c36c059/scripts/validation/updated-docs/Operation-Object.json)
#### [1st change](https://github.com/AnimeshKumar923/test-repo/blob/main/embedded-examples-error-resolution.md#1st-change) and [2nd change](https://github.com/AnimeshKumar923/test-repo/blob/main/embedded-examples-error-resolution.md#2nd-change)

Congrats! You found some bugs in the examples 🎉 The fixes are the ones you mentioned, so 👍 

#### [3rd change](https://github.com/AnimeshKumar923/test-repo/blob/main/embedded-examples-error-resolution.md#3rd-change)

In this case, you should not remove lines but fix the example. Operation Messages should point to the messages of the related channel. The `$ref` should be, for example, `#/channels/<channelName>/messages/<messageName>`. More on https://www.asyncapi.com/docs/reference/specification/v3.0.0#operationObject. 

### [Message-Example-Object.jso](https://github.com/asyncapi/spec/blob/ced626f6f6abf80e128216793a1bdc656c36c059/scripts/validation/updated-docs/Message-Example-Object.json) and [Operation-Traits-Object.json](https://github.com/asyncapi/spec/blob/ced626f6f6abf80e128216793a1bdc656c36c059/scripts/validation/updated-docs/Operation-Traits-Object.json)

Please share more info about the error instead of the fix you did so I can understand what the issue is.

scripts/validation/embedded-examples-validation.js
smoya: Why not having everything in memory instead of writing to disk?

spec/asyncapi.md
smoya: As we discussed in the past, you can transform those yaml to json on runtime and then validate as you do for json

AnimeshKumar923: > As we discussed in the past, you can transform those yaml to json on runtime and then validate as you do for json. Just a suggestion.

Yes, I'll exactly do that. That's the approach I had in mind. Thanks for the reminder!

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: > Why not having everything in memory instead of writing to disk?

Let me try that...

AnimeshKumar923:
> > > > * Ran into some validation problems which I've sorted them out. I wrote them in a [separate document](https://github.com/AnimeshKumar923/test-repo/blob/main/embedded-examples-error-resolution.md).
> > > 
> > > 
> > > I don't get the whole content of this file. When you mention lines of a document, to which document do you refer? It seems don't match with the spec file so I don't know which other file could be.
> > 
> > 
> > Ah, it's my shortcoming. I've updated the [document](https://github.com/AnimeshKumar923/test-repo/blob/main/embedded-examples-error-resolution.md) with reference to the files being mentioned. Please have a look once again.
> 
> ### [Operation-Object.json](https://github.com/asyncapi/spec/blob/ced626f6f6abf80e128216793a1bdc656c36c059/scripts/validation/updated-docs/Operation-Object.json)
> #### [1st change](https://github.com/AnimeshKumar923/test-repo/blob/main/embedded-examples-error-resolution.md#1st-change) and [2nd change](https://github.com/AnimeshKumar923/test-repo/blob/main/embedded-examples-error-resolution.md#2nd-change)
> 
> Congrats! You found some bugs in the examples 🎉 The fixes are the ones you mentioned, so 👍
> #### [3rd change](https://github.com/AnimeshKumar923/test-repo/blob/main/embedded-examples-error-resolution.md#3rd-change)


Thank you! :pray: I suspected that was the case... :eyes: 

> In this case, you should not remove lines but fix the example. Operation Messages should point to the messages of the related channel. The `$ref` should be, for example, `#/channels/<channelName>/messages/<messageName>`. More on https://www.asyncapi.com/docs/reference/specification/v3.0.0#operationObject.

Let me check this again. Will report you back.

> ### [Message-Example-Object.jso](https://github.com/asyncapi/spec/blob/ced626f6f6abf80e128216793a1bdc656c36c059/scripts/validation/updated-docs/Message-Example-Object.json) and [Operation-Traits-Object.json](https://github.com/asyncapi/spec/blob/ced626f6f6abf80e128216793a1bdc656c36c059/scripts/validation/updated-docs/Operation-Traits-Object.json)
> 
> Please share more info about the error instead of the fix you did so I can understand what the issue is.

@smoya I've updated the [document](https://github.com/AnimeshKumar923/test-repo/blob/main/embedded-examples-error-resolution.md#problems-in-message-object) to reflect the same. PTAL.

spec/asyncapi.md
AnimeshKumar923: fixed more via 10ad1ed

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: resolved through d7eb944

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: resolved through d7eb944

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: resolve via d7eb944

spec/asyncapi.md
AnimeshKumar923: but it's added [here](https://github.com/asyncapi/spec/blob/d7eb944bf04c0764866430bbbe8b65538dae8f20/spec/asyncapi.md?plain=1#L367C1-L374C1) and it's also validating in the workflow: image 

You can also see the [temp file here](https://github.com/asyncapi/spec/blob/d7eb944bf04c0764866430bbbe8b65538dae8f20/scripts/validation/updated-docs/License-Object.json)

spec/asyncapi.md
AnimeshKumar923: resolved via 53290ea

AnimeshKumar923: added yaml testing through cf62169

AnimeshKumar923:
> [Operation-Object.json](https://github.com/asyncapi/spec/blob/ced626f6f6abf80e128216793a1bdc656c36c059/scripts/validation/updated-docs/Operation-Object.json)
> [1st change](https://github.com/AnimeshKumar923/test-repo/blob/main/embedded-examples-error-resolution.md#1st-change) and [2nd change](https://github.com/AnimeshKumar923/test-repo/blob/main/embedded-examples-error-resolution.md#2nd-change)
>
> Congrats! You found some bugs in the examples 🎉 The fixes are the ones you mentioned, so 👍

@smoya shall I make the changes in this PR? 
Or open an issue and the corresponding PR fixing it?

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: resolved via e2cc95a

smoya:
> > [Operation-Object.json](https://github.com/asyncapi/spec/blob/ced626f6f6abf80e128216793a1bdc656c36c059/scripts/validation/updated-docs/Operation-Object.json)
> > [1st change](https://github.com/AnimeshKumar923/test-repo/blob/main/embedded-examples-error-resolution.md#1st-change) and [2nd change](https://github.com/AnimeshKumar923/test-repo/blob/main/embedded-examples-error-resolution.md#2nd-change)
> > Congrats! You found some bugs in the examples 🎉 The fixes are the ones you mentioned, so 👍
> 
> @smoya shall I make the changes in this PR? Or open an issue and the corresponding PR fixing it?

I would say you can open another PR against your branch `embedded-examples-validation` so you can demonstrate THIS PR will make CI failing but the other contains the fixes. Then you can merge it later.

But not a really big deal, up to you.

AnimeshKumar923:
> > > [Operation-Object.json](https://github.com/asyncapi/spec/blob/ced626f6f6abf80e128216793a1bdc656c36c059/scripts/validation/updated-docs/Operation-Object.json)
> > > [1st change](https://github.com/AnimeshKumar923/test-repo/blob/main/embedded-examples-error-resolution.md#1st-change) and [2nd change](https://github.com/AnimeshKumar923/test-repo/blob/main/embedded-examples-error-resolution.md#2nd-change)
> > > Congrats! You found some bugs in the examples 🎉 The fixes are the ones you mentioned, so 👍
> > 
> > 
> > @smoya shall I make the changes in this PR? Or open an issue and the corresponding PR fixing it?
> 
> I would say you can open another PR against your branch `embedded-examples-validation` so you can demonstrate THIS PR will make CI failing but the other contains the fixes. Then you can merge it later.
> 
> But not a really big deal, up to you.

@smoya 
I'm thinking that a separate issue and a PR approach would be much more neat and manageable for this task. I should keep this PR as issue-oriented as possible 

I can create a new issue by referencing the appropriate comment for it, immediately after this PR is merged :+1: 
Thanks!

scripts/validation/embedded-examples-validation.js
```diff
@@ -0,0 +1,119 @@
...
+    console.error(`\x1b[31mValidation failed for ${name}: ${error.message}\x1b[0m`);
+  }
+}
+
+// Iterate over the combinedData array, apply updates, and validate each document
+const baseDocPath = './base-doc.json';
+const baseDocSecuritySchemePath = './base-doc-security-scheme-object.json';
+
+const baseDoc = JSON.parse(fs.readFileSync(baseDocPath, 'utf8'));
+const baseDocSecurityScheme = JSON.parse(fs.readFileSync(baseDocSecuritySchemePath, 'utf8'));
```
smoya, line 97: Can't be this merged into the general basedoc?

scripts/validation/embedded-examples-validation.js
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// Extract comments and examples from the markdown file
+const combinedData = extractCommentsAndExamples(markdownContent);
+
+// Function to apply JSON Merge Patch updates to the document
+function applyUpdates(updates, baseDoc) {
+  updates.forEach(update => {
+    try {
+      if (update.json_path === "$") {
```
smoya, line 54: `$` is not a valid json pointer right? I guess we would need to just change it by having an empty json_path (actually `json_pointer`) field

smoya: Additionally, you can remove this check and directly run the `else` part, because most probably `jsonpointer.get(baseDoc, '');` will return a pointer to the root. Just try it

AnimeshKumar923: Let me see that...

AnimeshKumar923: @smoya I tried what you said but was getting this error while attempting it:
```bash
Error during validations: TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'Object'
    --- property 'undefined' closes the circle
    at JSON.stringify (<anonymous>)
    at logDocument (/home/animeshkumar923/open-source/asyncapi-spec/scripts/validation/test-v1.js:92:35)
    at /home/animeshkumar923/open-source/asyncapi-spec/scripts/validation/test-v1.js:115:3
    at Array.map (<anonymous>)
    at Object.<anonymous> (/home/animeshkumar923/open-source/asyncapi-spec/scripts/validation/test-v1.js:108:41)
    at Module._compile (node:internal/modules/cjs/loader:1376:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1435:10)
    at Module.load (node:internal/modules/cjs/loader:1207:32)
    at Module._load (node:internal/modules/cjs/loader:1023:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:135:12)
```
I don't know exactly what's going wrong so I'm going with the previous approach. For that `if-else` approach, it adds correctly the values of root doc in the updated doc.

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: the examples under the `security scheme object` as far as I've seen if they contain these objects, they fail the validation:
```json
"flows": {
  "clientCredentials": {
    "tokenUrl": "https://example.com/api/oauth/dialog",
    "availableScopes": {
      "subscribe:auth_revocations": "Scope required for authorization revocation topic"
    }
  }
},
"scopes": [
  "subscribe:auth_revocations"
]
```

So I though to separate the whole `security scheme object` examples and provide them with a separate doc with minimal objects to add upon, and by doing that, they are exactly constructed as expected and are, in the end, valid documents.

smoya: Not ideal at all to maintain two base documents and have logic dedicated for those. We should work on making it work as just one doc. Happy to assist.

scripts/validation/embedded-examples-validation.js
smoya: But my point still is valid: `$` I not a valid json pointer

AnimeshKumar923:
## UPDATE

@smoya 

- So I was testing further on the json merge patch in a [separate branch](https://github.com/AnimeshKumar923/asyncapi-spec/pull/1), to keep this branch untouched for the time being.

- All the files are being validated as expected after the json merge patch integration except for [`Operation Object`](https://github.com/AnimeshKumar923/asyncapi-spec/blob/f6f527a19e87a02cd0b19bbfda80701cbaace753/scripts/validation/embed-logs/Operation%20Object-json-format.json) and [`Security Scheme OAuth Flow Object`](https://github.com/AnimeshKumar923/asyncapi-spec/blob/f6f527a19e87a02cd0b19bbfda80701cbaace753/scripts/validation/embed-logs/Security%20Scheme%20OAuth%20Flow%20Object-json-format.json) 

- The [updated version](https://github.com/AnimeshKumar923/asyncapi-spec/blob/f6f527a19e87a02cd0b19bbfda80701cbaace753/scripts/validation/embedded-with-logs.js) of the script. (naming is different)

---

- The error for **Operation Object** is like this:
```bash
Warning in Operation Object-json-format: "0" property must have required property "type"
Warning in Operation Object-json-format: "0" property must not be valid
Warning in Operation Object-json-format: Property "petstore_auth" is not expected to be here
Warning in Operation Object-json-format: Operation message does not belong to the specified channel.
```
- [This is the specified section](https://github.com/AnimeshKumar923/asyncapi-spec/blob/f6f527a19e87a02cd0b19bbfda80701cbaace753/scripts/validation/embed-logs/Operation%20Object-json-format.json#L171C2-L187C9) it's referring to.

---

- The error for **Security Scheme OAuth Flow Object** is like this:
```bash
Warning in Security Scheme OAuth Flow Object-json-format: "0" property must have required property "in"
Warning in Security Scheme OAuth Flow Object-json-format: "0" property must not be valid
Warning in Security Scheme OAuth Flow Object-json-format: "0" property must have required property "scheme"
Warning in Security Scheme OAuth Flow Object-json-format: "0" property must have required property "name"
Warning in Security Scheme OAuth Flow Object-json-format: "0" property must have required property "openIdConnectUrl"
Warning in Security Scheme OAuth Flow Object-json-format: Property "flows" is not expected to be here
Warning in Security Scheme OAuth Flow Object-json-format: "clientCredentials" property must not be valid
Warning in Security Scheme OAuth Flow Object-json-format: Property "scopes" is not expected to be here
Warning in Security Scheme OAuth Flow Object-json-format: Property "flows" is not expected to be here
Warning in Security Scheme OAuth Flow Object-json-format: "clientCredentials" property must not be valid
Warning in Security Scheme OAuth Flow Object-json-format: Property "scopes" is not expected to be here
Warning in Security Scheme OAuth Flow Object-json-format: "0" property must have required property "in"
Warning in Security Scheme OAuth Flow Object-json-format: "0" property must not be valid
Warning in Security Scheme OAuth Flow Object-json-format: "0" property must have required property "scheme"
Warning in Security Scheme OAuth Flow Object-json-format: "0" property must have required property "name"
Warning in Security Scheme OAuth Flow Object-json-format: "0" property must have required property "openIdConnectUrl"
Warning in Security Scheme OAuth Flow Object-json-format: "oauth" property must have required property "in"
Warning in Security Scheme OAuth Flow Object-json-format: "oauth" property must not be valid
Warning in Security Scheme OAuth Flow Object-json-format: "oauth" property must have required property "scheme"
Warning in Security Scheme OAuth Flow Object-json-format: "oauth" property must have required property "name"
Warning in Security Scheme OAuth Flow Object-json-format: "oauth" property must have required property "openIdConnectUrl"
Warning in Security Scheme OAuth Flow Object-json-format: Property "flows" is not expected to be here
Warning in Security Scheme OAuth Flow Object-json-format: "clientCredentials" property must not be valid
Warning in Security Scheme OAuth Flow Object-json-format: Property "scopes" is not expected to be here

```
- [This is the specified section](https://github.com/AnimeshKumar923/asyncapi-spec/blob/f6f527a19e87a02cd0b19bbfda80701cbaace753/scripts/validation/embed-logs/Security%20Scheme%20OAuth%20Flow%20Object-json-format.json#L595C4-L614C7) it's referring to.

---

- There's also one more thing that has been changed in that script. As far as I've understood _(please correct me if I'm wrong)_, for the arrays present in the document, the nature of merge patch is such that it doesn't applied to arrays, and it ignores them or treat it like an object.
- The following is an excerpt from the [RFC 7396](https://datatracker.ietf.org/doc/html/rfc7396):
```
There are a few things to note about the function. 
If the patch is anything other than an object,
the result will always be to replace the entire target with the entire patch. 
Also, it is not possible patch part of a target that is not an object,
such as to replace some of the values in an array.
```

- So keeping this in mind, what I did is that added a [check for an array](https://github.com/AnimeshKumar923/asyncapi-spec/blob/f6f527a19e87a02cd0b19bbfda80701cbaace753/scripts/validation/embedded-with-logs.js#L66C1-L78C8) in the script and made this modification in the comments for the jsonPointer of the two array objects: [one](https://github.com/AnimeshKumar923/asyncapi-spec/blob/f6f527a19e87a02cd0b19bbfda80701cbaace753/spec/asyncapi.md?plain=1#L954C65-L954C115), [two](https://github.com/AnimeshKumar923/asyncapi-spec/blob/f6f527a19e87a02cd0b19bbfda80701cbaace753/spec/asyncapi.md?plain=1#L1441C111-L1441C121). After that modification, it actually adds the content inside the arrays.
- Specific [commit](https://github.com/AnimeshKumar923/asyncapi-spec/commit/2c3cadf3b5a28ce659820b0822b210458085e67b) where the change was made
- I suspect that the errors in the two files above are also due to the nature of the merge patch. Let me know what you think about this whole scenario.

Thanks! :+1:

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: > But my point still is valid: `$` I not a valid json pointer

Yes, that's correct. I've changed it to point to `if (update.json_path === "")`

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: Okay, let me try it for myself once. I'll let you know the results. :+1:

smoya:
@AnimeshKumar923 just for the record, I shared with you via Slack DM the fixes of such examples (Yeah, your script is making us discovery errors on the examples! 🎉 👏  ).

Regarding the following:

> * So keeping this in mind, what I did is that added a [check for an array](https://github.com/AnimeshKumar923/asyncapi-spec/blob/f6f527a19e87a02cd0b19bbfda80701cbaace753/scripts/validation/embedded-with-logs.js#L66C1-L78C8) in the script and made this modification in the comments for the jsonPointer of the two array objects: [one](https://github.com/AnimeshKumar923/asyncapi-spec/blob/f6f527a19e87a02cd0b19bbfda80701cbaace753/spec/asyncapi.md?plain=1#L954C65-L954C115), [two](https://github.com/AnimeshKumar923/asyncapi-spec/blob/f6f527a19e87a02cd0b19bbfda80701cbaace753/spec/asyncapi.md?plain=1#L1441C111-L1441C121). After that modification, it actually adds the content inside the arrays.

You are pointing to examples with a JSON Pointer such as `#/operations/sendUserSignUp/traits/0`. That JSON Pointer is not pointing to an array but an object. The element 0 of such an array is a trait object, so I don't think you should find any issue right? I would expect such replacement only if you specify something like `#/operations/sendUserSignUp/traits` (without the zero index at the end) but it's not the case.

AnimeshKumar923:
> @AnimeshKumar923 just for the record, I shared with you via Slack DM the fixes of such examples (Yeah, your script is making us discovery errors on the examples! 🎉 👏 ).

Thank you! :bow: 

> Regarding the following:
> 
> > * So keeping this in mind, what I did is that added a [check for an array](https://github.com/AnimeshKumar923/asyncapi-spec/blob/f6f527a19e87a02cd0b19bbfda80701cbaace753/scripts/validation/embedded-with-logs.js#L66C1-L78C8) in the script and made this modification in the comments for the jsonPointer of the two array objects: [one](https://github.com/AnimeshKumar923/asyncapi-spec/blob/f6f527a19e87a02cd0b19bbfda80701cbaace753/spec/asyncapi.md?plain=1#L954C65-L954C115), [two](https://github.com/AnimeshKumar923/asyncapi-spec/blob/f6f527a19e87a02cd0b19bbfda80701cbaace753/spec/asyncapi.md?plain=1#L1441C111-L1441C121). After that modification, it actually adds the content inside the arrays.
> 
> You are pointing to examples with a JSON Pointer such as `#/operations/sendUserSignUp/traits/0`. That JSON Pointer is not pointing to an array but an object. The element 0 of such an array is a trait object, so I don't think you should find any issue right? I would expect such replacement only if you specify something like `#/operations/sendUserSignUp/traits` (without the zero index at the end) but it's not the case.

Yes, that is happening in this case here.

So what I observed was that, if we specify this JSON Pointer --> `#/operations/sendUserSignUp/traits`, it doesn't adds the structure inside the `traits` array. 
And when we use this Pointer instead --> `#/operations/sendUserSignUp/traits/0`, it add the structure inside the array.

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: Hi @smoya :wave: 
- Tried adding more cases to [the security scheme field](https://gist.github.com/AnimeshKumar923/45e1e38934964f8a64e4fe36dd061246#file-securityscheme-json). 
- Passing the validation as expected, but unsure about the structure of the `securityScheme`. Can you please verify it once? :sweat_smile: 
- Full base doc also given in the gist.

smoya:
> And when we use this Pointer instead --> `#/operations/sendUserSignUp/traits/0`, it add the structure inside the array.

And this is not enough? I mean, isn't that what you wanted to do?

AnimeshKumar923:
> > And when we use this Pointer instead --> `#/operations/sendUserSignUp/traits/0`, it add the structure inside the array.
> 
> And this is not enough? I mean, isn't that what you wanted to do?

@smoya 
Yes that's what I wanted. I'm not objecting anything here, I was just stating my observations to you that ''we are following this approach here, and this is what's happening currently" :+1:

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: @smoya gentle reminder

smoya: > Maybe [this](https://github.com/AnimeshKumar923/asyncapi-spec/blob/54fab4bd1f172a81e43ecd6a2058a4160f1f4e06/spec/asyncapi.md?plain=1#L2614C1-L2634C52) is the potential fix to the **OAuth Flow Object**? 👀

Yes, it is. In fact, I suggest you to add a new subtitle here clarifying with kind of flow the example is representing. 
I would add 

```

###### clientCredentials Oauth Flow object example
```

Right after the `##### OAuth Flow Object Examples` at https://github.com/AnimeshKumar923/asyncapi-spec/blob/54fab4bd1f172a81e43ecd6a2058a4160f1f4e06/spec/asyncapi.md?plain=1#L2612

smoya: Any update on [this](https://github.com/asyncapi/spec/pull/1059#discussion_r1648575632)? Having two files is completely not ideal.

scripts/validation/updated-docs/Channel-Object-json-format.json
smoya, line 1: SHouldn't you remove this whole dir now that, IIRC, the validation happens in memory?

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: > Any update on [this](https://github.com/asyncapi/spec/pull/1059#discussion_r1648575632)? Having two files is completely not ideal.

@smoya 
About that, please refer to [this comment](https://github.com/asyncapi/spec/pull/1059#discussion_r1655149132)

scripts/validation/updated-docs/Channel-Object-json-format.json
AnimeshKumar923: > Shouldn't you remove this whole dir now that, IIRC, the validation happens in memory?

Yes sure. I had that to track the changes, but can now I can reference those happening in the separate branch.

AnimeshKumar923: resolved through f31ff2a

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: > > Maybe [this](https://github.com/AnimeshKumar923/asyncapi-spec/blob/54fab4bd1f172a81e43ecd6a2058a4160f1f4e06/spec/asyncapi.md?plain=1#L2614C1-L2634C52) is the potential fix to the **OAuth Flow Object**? 👀
> 
> Yes, it is. In fact, I suggest you to add a new subtitle here clarifying with kind of flow the example is representing. I would add
> 
> ```
> 
> ###### clientCredentials Oauth Flow object example
> ```
> 
> Right after the `##### OAuth Flow Object Examples` at https://github.com/AnimeshKumar923/asyncapi-spec/blob/54fab4bd1f172a81e43ecd6a2058a4160f1f4e06/spec/asyncapi.md?plain=1#L2612

Got it. Will be adding this in [this PR](https://github.com/AnimeshKumar923/asyncapi-spec/pull/2) which is open against the `embedded-examples-validation` branch.

AnimeshKumar923: > > > Maybe [this](https://github.com/AnimeshKumar923/asyncapi-spec/blob/54fab4bd1f172a81e43ecd6a2058a4160f1f4e06/spec/asyncapi.md?plain=1#L2614C1-L2634C52) is the potential fix to the **OAuth Flow Object**? 👀
> > 
> > 
> > Yes, it is. In fact, I suggest you to add a new subtitle here clarifying with kind of flow the example is representing. I would add
> > ```
> > 
> > ###### clientCredentials Oauth Flow object example
> > ```
> > Right after the `##### OAuth Flow Object Examples` at https://github.com/AnimeshKumar923/asyncapi-spec/blob/54fab4bd1f172a81e43ecd6a2058a4160f1f4e06/spec/asyncapi.md?plain=1#L2612
> 
> Got it. Will be adding this in [this PR](https://github.com/AnimeshKumar923/asyncapi-spec/pull/2) which is open against the `embedded-examples-validation` branch.

Implemented it [here](https://github.com/AnimeshKumar923/asyncapi-spec/blob/b0d36e7d35c023c8a67ed298e0320067fd5409c9/spec/asyncapi.md?plain=1#L2612C1-L2632C4)

spec/asyncapi.md
AnimeshKumar923: marking it as resolved as validation is passing as of 4ad2444

scripts/validation/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "validation",
+  "version": "1.0.0",
+  "description": "for validation of embedded examples in asyncapi.md file",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "validate:examples": "node embedded-examples-validation.js"
+  },
+  "author": "Animesh Kumar",
+  "license": "ISC",
```
smoya, line 10: Any particular reason why ISC instead of Apache 2.0 ?

scripts/validation/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "devDependencies": {
+    "@asyncapi/parser": "^3.1.0",
+    "js-yaml": "^4.1.0",
+    "json-merge-patch": "^1.0.2",
+    "jsonpath-plus": "^9.0.0",
```
smoya, line 15: You don't need this dependency anymore I believe.

scripts/validation/embed-logs/Message Object-json-format.json
```diff
@@ -0,0 +1,730 @@
...
+            "user": {
+              "$ref": "#/components/schemas/userCreate"
+            },
+            "signup": {
+              "$ref": "#/components/schemas/signup"
+            }
+          },
+          "schemaFormat": "application/vnd.apache.avro+json;version=1.9.0",
+          "schema": {
+            "$ref": "path/to/user-create.avsc#/UserCreate"
```
AnimeshKumar923, line 481: The error message for this line:
```bash
Warning in Message Object-json-format: ENOENT: no such file or directory, open 'path/to/user-create.avsc'
```
Same error for the yaml version as well.
@smoya potential fix?

scripts/validation/embed-logs/Components Object-json-format.json
```diff
@@ -0,0 +1,787 @@
...
+            {
+              "name": "id",
+              "type": "int"
+            },
+            {
+              "name": "name",
+              "type": "string"
+            }
+          ],
+          "$ref": "path/to/user-create.avsc#/UserCreate"
```
AnimeshKumar923, line 264: Error for this file:
```bash
Warning in Components Object-json-format: "AvroExample" property must match "then" schema
Warning in Components Object-json-format: "schema" property must match exactly one schema in oneOf
Warning in Components Object-json-format: ENOENT: no such file or directory, open 'path/to/user-create.avsc'
```

@smoya Potential fix?

scripts/validation/package.json
AnimeshKumar923: Yes, will remove those now!

scripts/validation/package.json
AnimeshKumar923: Don't have much idea about licensing, will change it to the recommended one...

scripts/validation/package.json
AnimeshKumar923: only the jsonpath isn't required

scripts/validation/embed-logs/Message Object-json-format.json
smoya: The error is because the file `path/to/user-create.avsc` is expected to exist.
You have two options here:

1. To include an avro file (`user-create.avsc`) in a known path like `.github/workflows/files` or similar path (not having a good suggestion here tbh)
2. Not so ideal, but to exclude the example from testing

scripts/validation/embed-logs/Components Object-json-format.json
smoya: https://github.com/asyncapi/spec/pull/1059#discussion_r1665911946

scripts/validation/package.json
AnimeshKumar923: achieved thorugh e3d6a78

scripts/validation/package.json
AnimeshKumar923: resolved through e3d6a78

scripts/validation/embed-logs/Message Object-json-format.json
AnimeshKumar923: I think including it would be much better. Let me work on it... :eyes:

scripts/validation/embed-logs/Components Object-json-format.json
AnimeshKumar923: Got it. Working on it...

scripts/validation/base-doc-combined.json
```diff
@@ -0,0 +1,711 @@
...
+            "availableScopes": {
+              "subscribe:auth_revocations": "Scope required for authorization revocation topic"
+            }
+          }
+        },
+        "scopes": [
+          "subscribe:auth_revocations"
+        ]
+      }
+    },
```
AnimeshKumar923, line 633: @smoya Please check this format of the document for the `securitySchemas` field if it's valid or not.

scripts/validation/user-create.avsc
```diff
@@ -0,0 +1,15 @@
+{
```
AnimeshKumar923, line 1: @smoya 

Added this file and changed the references in the spec. Commit details: f814d826

It's validating perfectly now with no errors whatsoever.

spec/asyncapi.md
```diff
@@ -1326,12 +1383,13 @@ Example using Avro to define the payload:
   "payload": {
     "schemaFormat": "application/vnd.apache.avro+json;version=1.9.0",
     "schema": {
-      "$ref": "path/to/user-create.avsc#/UserCreate"
+      "$ref": "user-create.avsc"
```
AnimeshKumar923, line 1386: @smoya PTAL at these references. Is this approach okay?

spec/asyncapi.md
```diff
@@ -1344,7 +1402,7 @@ tags:
 payload:
   schemaFormat: 'application/vnd.apache.avro+yaml;version=1.9.0'
   schema:
-    $ref: 'path/to/user-create.avsc/#UserCreate'
+    $ref: 'user-create.avsc'
```
AnimeshKumar923, line 1405: @smoya PTAL at these references. Is this approach okay?

spec/asyncapi.md
```diff
@@ -1595,7 +1664,7 @@ my.org.User
       "AvroExample": {
         "schemaFormat": "application/vnd.apache.avro+json;version=1.9.0",
         "schema": {
-          "$ref": "path/to/user-create.avsc#/UserCreate"
+          "$ref": "user-create.avsc"
```
AnimeshKumar923, line 1667: @smoya PTAL at these references.

spec/asyncapi.md
```diff
@@ -1719,7 +1789,7 @@ components:
     AvroExample:
       schemaFormat: application/vnd.apache.avro+json;version=1.9.0
       schema:
-        $ref: 'path/to/user-create.avsc/#UserCreate'
+        $ref: 'user-create.avsc'
```
AnimeshKumar923, line 1792: @smoya PTAL at these references.

scripts/validation/embed-logs/Message Object-json-format.json
AnimeshKumar923: Validating perfectly after the changes made via f814d82

AnimeshKumar923:
# Current State of PR

Summary of PR as of 5e1b0be.

## Contents required for the validation:

- [Spec document](https://github.com/asyncapi/spec/blob/f814d8266bd880fa042e120d34b9f45924c1fad2/spec/asyncapi.md) that contains the embedded examples.

- The main script [`validate-embedded-examples.js`](https://github.com/asyncapi/spec/blob/f814d8266bd880fa042e120d34b9f45924c1fad2/scripts/validation/embedded-examples-validation.js) which does the validation.

- The [base document](https://github.com/asyncapi/spec/blob/f814d8266bd880fa042e120d34b9f45924c1fad2/scripts/validation/base-doc-combined.json) that is utilized by the script, along with the extracted example, to construct a fully valid AsyncAPI document.

- An [external Avro Scheme](https://github.com/asyncapi/spec/blob/f814d8266bd880fa042e120d34b9f45924c1fad2/scripts/validation/user-create.avsc) file that is referenced in multiple locations in the spec.

- [Workflow file](https://github.com/asyncapi/spec/blob/f814d8266bd880fa042e120d34b9f45924c1fad2/.github/workflows/validate-embedded-examples.yml) which is triggered on incoming PR and checks spec examples.

---

## Flow of the validation

- Workflow is triggered on every incoming PR.
- The script is then executed: _(a brief flow)_
  - The spec file is loaded in the memory
  - The markdown comment and example are extracted. Each component of the comment and the corresponding example is stored in its separate property of the object. The object is stored in an array.
  - All the metadata is then collected and stored in the `combinedData` array.
  - Then for each example present in the `combinedData` array, a full AsyncAPI document is constructed and validated using the parser-js.
  - If all goes well, it gives the message `All examples validated successfully!`. Otherwise, we get the error(s) with the file name and the error output.

---

## Remaining tasks

- [ ] Remove non-essential files
- [ ] A test run for the workflow at the current branch
- [ ] Final review from @smoya

smoya:
> * [ ]  Remove non-essential files
> * [ ]  A test run for the workflow at the current branch
> * [ ]  Final review from @smoya

👍 

And for the record, can you confirm you will work, after merging this (non-related with the Bounty issue) on moving all this knowledge into proper documentation about how validation happens?

AnimeshKumar923:
> > * [ ]   Remove non-essential files
> > * [ ]   A test run for the workflow at the current branch
> > * [ ]   Final review from @smoya
> 
> 👍
> 
> And for the record, can you confirm you will work, after merging this (non-related with the Bounty issue) on moving all this knowledge into proper documentation about how validation happens?

Yes, gladly! I have added that task.

scripts/validation/embed-logs/Components Object-json-format.json
AnimeshKumar923: Same issue as https://github.com/asyncapi/spec/pull/1059#discussion_r1666331157

Marking resolved

AnimeshKumar923:
## Workflow Test Runs

- Workflow fails: https://github.com/AnimeshKumar923/asyncapi-spec/actions/runs/9806341756/job/27077848952

- Workflow passes: https://github.com/AnimeshKumar923/asyncapi-spec/actions/runs/9806389937/job/27078005627

scripts/validation/base-doc-combined.json
smoya: LGTM!

spec/asyncapi.md
```diff
@@ -333,8 +335,7 @@ Field Name | Type | Description
 
 This object MAY be extended with [Specification Extensions](#specificationExtensions).
 
-##### Contact Object Example
```
smoya, line 336: I don't recall why we removed this one 🤔

spec/asyncapi.md
```diff
@@ -1326,12 +1383,13 @@ Example using Avro to define the payload:
   "payload": {
     "schemaFormat": "application/vnd.apache.avro+json;version=1.9.0",
     "schema": {
-      "$ref": "path/to/user-create.avsc#/UserCreate"
+      "$ref": "./user-create.avsc"
```
smoya, line 1386: Please make consistent the usage of quotes vs double quotes.

spec/asyncapi.md
AnimeshKumar923: Oh, that was by mistake I guess. Will revert it.

spec/asyncapi.md
AnimeshKumar923: Can you please elaborate on the locations? As far as I've seen, the strings in YAML are single quoted, and JSON examples are double quoted.

AnimeshKumar923: I've made these changes which I found inconsistent: c788faf

spec/asyncapi.md
AnimeshKumar923: done through b24f71d

scripts/validation/embedded-examples-validation.js
```diff
@@ -0,0 +1,128 @@
...
+  }
+}
+
+// Iterate over the combinedData array, apply updates, and validate each document
+const baseDocPath = './base-doc-combined.json';
+
+const baseDoc = JSON.parse(fs.readFileSync(baseDocPath, 'utf8'));
+
+const validationPromises = combinedData.map(async (item) => {
+  const baseDocument = JSON.parse(JSON.stringify(baseDoc));
```
smoya, line 111: Why do you first stringify to later do JSON.parse? You already have a json object 🤔

scripts/validation/embedded-examples-validation.js
```diff
@@ -0,0 +1,128 @@
...
+  try {
+    const diagnostics = await parser.validate(document);
+
+    if (diagnostics.length > 0) {
+      diagnostics.forEach(diagnostic => {
+        if (diagnostic.level === 'error') {
+          console.error(`\x1b[31mError in ${name}: ${diagnostic.message}\x1b[0m`);
+          process.exit(1);
+        } else {
+          console.log(`\x1b[31mError in ${name}: ${diagnostic.message}\x1b[0m`);
```
smoya, line 92: This code does the same as if in the `if` 🤔

smoya: I would suggest you store the message in a variable so you don't duplicate that much code

AnimeshKumar923: > I would suggest you store the message in a variable so you don't duplicate that much code

Understood.

> This code does the same as if in the if 🤔 but changes the console log level.

Will keep the console log only

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: Ah, yes, my shortcoming. I'll fix it now.

spec/asyncapi.md
AnimeshKumar923: Addressed more issues I could find via b60d353

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: resolved via 5c5d91a

scripts/validation/embedded-examples-validation.js
smoya: It's fine that you use `console.log` or `console.error` depending on the error. 
I meant that you can just write the message in a constant and reuse it later.

Something like 

```ts
const msg = `\x1b[31mError in ${name}: ${diagnostic.message}\x1b[0m`;
if (diagnostic.level === 'error') {
  console.error(msg);
  process.exit(1);
} 

console.log(msg);
```

AnimeshKumar923: implemented through 5c5d91a
PTAL @smoya

AnimeshKumar923: > It's fine that you use `console.log` or `console.error` depending on the error. I meant that you can just write the message in a constant and reuse it later.
> 
> Something like
> 
> ```ts
> const msg = `\x1b[31mError in ${name}: ${diagnostic.message}\x1b[0m`;
> if (diagnostic.level === 'error') {
>   console.error(msg);
>   process.exit(1);
> } 
> 
> console.log(msg);
> ```

Understood. Applied it 34e00e0

scripts/validation/embedded-examples-validation.js
```diff
@@ -0,0 +1,128 @@
...
+          process.exit(1);
```
smoya, line 94: This means that warnings will also make the CI give error. Please remove this `process.exit(1)`

scripts/validation/embedded-examples-validation.js
smoya: You reverted the part where this message was extracted into a constant?

scripts/validation/embedded-examples-validation.js
smoya: I don't see it resolved 🤔

AnimeshKumar923: Let me check once

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: Please check the latest changes

scripts/validation/embedded-examples-validation.js
```diff
@@ -0,0 +1,129 @@
...
+const baseDocPath = './base-doc-combined.json';
+
+const baseDoc = JSON.parse(fs.readFileSync(baseDocPath, 'utf8'));
+
+const validationPromises = combinedData.map(async (item) => {
+  const baseDocument = JSON.parse(JSON.stringify(baseDoc));
+
+  const updatedDocument = applyUpdates([item], baseDocument);
+
+  const documentString = updatedDocument;
```
smoya, line 116: unnecessary reasignment

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: https://github.com/asyncapi/spec/blob/34e00e066d499554a7133066c6f4fd1352cf36cc/scripts/validation/embedded-examples-validation.js#L86C3-L95C10

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: okay.

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: PTAL 81b40a5

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: PTAL 81b40a5

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: PTAL 81b40a5

scripts/validation/embedded-examples-validation.js
AnimeshKumar923: PTAL 81b40a5

scripts/validation/embedded-examples-validation.js
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// Extract comments and examples from the markdown file
+const combinedData = extractCommentsAndExamples(markdownContent);
+
+// Function to apply JSON Merge Patch updates to the document
+function applyUpdates(updates, baseDoc) {
+  updates.forEach(update => {
+    try {
+      const jsonPointerPath = update.json_pointer;
```
smoya, line 54: Nitpick: remove this line. IMHO it is better to know the example comes from the object in the comment example

```suggestion
```

scripts/validation/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "validation",
+  "version": "1.0.0",
+  "description": "for validation of embedded examples in asyncapi.md file",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "validate:examples": "node embedded-examples-validation.js"
+  },
+  "author": "Animesh Kumar",
+  "license": "Apache 2.0",
```
smoya, line 10:
```suggestion
  "license": "Apache-2.0",
```

AnimeshKumar923:
https://github.com/asyncapi/spec/pull/1059#issuecomment-2210485497

PR #1082, opened 2024-12-31
docs: update Message Example headers and payload

Pakisan:
According to JSON Schema headers and payload properties may be of any type

See:
- https://github.com/asyncapi/jasyncapi/pull/208
- https://github.com/asyncapi/spec-json-schemas/issues/560
- https://github.com/asyncapi/website/issues/3530

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -1453,8 +1453,8 @@ Message Example Object represents an example of a [Message Object](#messageObjec
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="messageExampleObjectHeaders"></a>headers | `Map[string, any]` | The value of this field MUST validate against the [Message Object's headers](#messageObjectHeaders) field.
-<a name="messageExampleObjectPayload"></a>payload | `Map[string, any]` | The value of this field MUST validate against the [Message Object's payload](#messageObjectPayload) field.
+<a name="messageExampleObjectHeaders"></a>headers | `any` | The value of this field MUST validate against the [Message Object's headers](#messageObjectHeaders) field.
+<a name="messageExampleObjectPayload"></a>payload | `any` | The value of this field MUST validate against the [Message Object's payload](#messageObjectPayload) field.
 <a name="messageExampleObjectName"></a>name | `string` | A machine-friendly name.
 <a name="messageExampleObjectSummary"></a>summary | `string` |  A short summary of what the example is about.
 


fmvilas (approved):
LGTM 👍

spec/asyncapi.md
```diff
@@ -1453,8 +1453,8 @@ Message Example Object represents an example of a [Message Object](#messageObjec
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="messageExampleObjectHeaders"></a>headers | `Map[string, any]` | The value of this field MUST validate against the [Message Object's headers](#messageObjectHeaders) field.
-<a name="messageExampleObjectPayload"></a>payload | `Map[string, any]` | The value of this field MUST validate against the [Message Object's payload](#messageObjectPayload) field.
+<a name="messageExampleObjectHeaders"></a>headers | `any` | The value of this field MUST validate against the [Message Object's headers](#messageObjectHeaders) field.
```
fmvilas, line 1456: I think the change is justified for `payload` as it can be an array, for instance. However, I'm not sure in which cases `headers` could be anything different than a map.

fmvilas (changes requested):
Changed my mind after thinking a bit more.

spec/asyncapi.md
Pakisan: @fmvilas Yep, I thought same way, but after this I re-read definitions and it may by of any type

In `Message` we are defining schema for `payload` and `headers`. This two properties are JSON Schema or Multi format Schema

So they can be of any defined and valid JSON Schema type - string, number, object, etc. Or Avro type

This is main reason to define `headers` and `payload` in `MessageExample` as `any` type and validate them trough it's schemas in `Message` object

derberg: ### headers

IMHO headers should stay as it is. It was always like this, in previous versions too, key/value as this is who headers look like.

> So they can be of any defined and valid JSON Schema type - string, number, object, etc. Or Avro type

spec is more specific in case of headers if you compare with payload

`Schema MUST be a map of key-value pairs`

### payload

In case of payload. This is what it was before v3 which makes sense as payload is `any`

```
<a name="messageExampleObjectHeaders"></a>headers | `Map[string, any]` | The value of this field MUST validate against the [Message Object's headers](#messageObjectHeaders) field.

<a name="messageExampleObjectPayload"></a>payload | `any` | The value of this field MUST validate against the [Message Object's payload](#messageObjectPayload) field.
```

in v3 we introduced Multi Format Schema Object, so payload can also be `payload.schema` and `payload.schemaFormat` and definition of payload was changed from `any` to `Multi Format Schema Object | Schema Object | Reference Object` but yeah, because of `Schema Object` it is still basically `any`.

which means message payload example should also be `any`.

I went through https://github.com/asyncapi/spec/pull/910 but I see that we were 100% focused there on the `message.payload` definition and `Multi Format Schema Object` that the change in Message Example Object Payload type was overlooked.

Pakisan: @fmvilas @derberg I have returned `Map[string, any]` to headers

fmvilas (approved):
Looks good to me 👍

spec/asyncapi.md
fmvilas: Thanks, @Pakisan! Approved ✅

derberg (approved):

derberg:
@dalelane @smoya @char0n @GreenRover can any of you have a look

dalelane (approved):
This looks fine to me - thanks

GreenRover (approved):

PR #1096, opened 2025-04-01
chore: add sponsors `apideck` and `mercedes-benz` and remove `redhat`

derberg:
Adding `apideck` thanks to @Amzani support

Adding `mercedes-benz` thanks to their exceptional culture - they sponsor because they know they need to invest in sustainability of standards they want to use long term: https://opensource.mercedes-benz.com/manifesto/. They actually pay more than bronze, $300 a month. I had a chat with them. For now, they only generate docs.

Removing `redhat`. Sponsorship was done in past by their marketing team. They changed marketing strategy. Unfortunately it's not clear how to reach proper people there that would support AsyncAPI for the same reason as Mercedes-Benz


fmvilas (approved):
👏  Great job, @derberg.

PR #1101, opened 2025-05-16
docs: fixes asyncapi/parser-js#1093 clarification of Operation->messages

ctataryn:
---
title: "Clarify the behaviour of operation->messages when the messages section is excluded from the operation."
---

Clarifies the point that excluding the `messages` field from the `operation` section will have the effect of including all messages from the channel.

Added the `<p>**Note**:...</p>` part to Operation Object -> messages in the spec doc.

**Related issue(s):**
- https://github.com/asyncapi/parser-js/issues/1093

---


spec/asyncapi.md
```diff
@@ -846,7 +846,7 @@ Field Name | Type | Description
...
+<a name="operationObjectMessages"></a>messages | [[Reference Object](#referenceObject)] | A list of `$ref` pointers pointing to the supported [Message Objects](#messageObject) that can be processed by this operation. It MUST contain a subset of the messages defined in the [channel referenced in this operation](#operationObjectChannel), and MUST NOT point to a subset of message definitions located in the [Messages Object](#componentsMessages) in the [Components Object](#componentsObject) or anywhere else. **Every message processed by this operation MUST be valid against one, and only one, of the [message objects](#messageObject) referenced in this list.** Please note the `messages` property value MUST be a list of [Reference Objects](#referenceObject) and, therefore, MUST NOT contain [Message Objects](#messageObject). However, it is RECOMMENDED that parsers (or other software) dereference this property for a better development experience. <p>**Note**: excluding this property from the Operation implies that all messages from the channel will be included. Explicitly set the `messages` property to `[]` if this operation should contain no messages.</p>
```
fmvilas, line 849: Out of curiosity, when is it useful to have an operation without messages?

derberg: yeah I think we need different wording

@fmvilas this PR got opened because of slack convo https://asyncapi.slack.com/archives/C0230UAM6R3/p1747347552431919. I don't think @ctataryn means operation should contain no message.

@fmvilas you remember when we talked about v3 the `messages` in `operation` and `operation reply` - we always called them a filtering features. That you can have one channel that specifies all messages, but then on each operation level you specify what messages it relates too. And we also talked about situation that if in the operation you have no messages specified explicitly, it means this operation relates to all messages specified in the channel

fmvilas: Thanks for the context, that helps 🙏 

I'm still curious to understand why `[]` should mean "this operation has no messages". Because we could easily say that `[]` is the equivalent of not setting the property at all, i.e., it means all the messages.

jonaslagoni: Just out of pure coincidence I stumpled upon this issue. From my point of view I use it for HTTP requests without body such as GET, I find the easiest to define with empty array. 
image

Have not figured out if there is an easier way 😄

fmvilas: That's a great example, @jonaslagoni. Thanks!

fmvilas (approved):
LGTM 👍

dalelane (approved):

derberg (approved):

PR #1109, opened 2025-06-05
feat: Add ROS 2 bindings to the official specification

amparo-siemens:
---
title: "Add ROS 2 bindings and protocol to the official specification" by SIEMENS AG
---

---

**Related issue(s):**
[ROS 2 binding PR](https://github.com/asyncapi/bindings/pull/270)

---

The ROS 2 binding was finalized in the Bindings repository.
This PR adds the new `ros2` binding in the Server Object, Channel Object, Operation Object and Message Object.
Along with adding `ros2` to the protocol list.

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -144,7 +144,7 @@ A channel is an addressable component, made available by the [server](#definitio
 
 ### <a name="definitionsProtocol"></a>Protocol
 
-A protocol is the mechanism (wireline protocol or API) by which [messages](#definitionsMessage) are exchanged between the application and the [channel](#definitionsChannel). Example protocols include, but are not limited to, AMQP, HTTP, JMS, Kafka, Anypoint MQ, MQTT, Solace, STOMP, Mercure, WebSocket, Google Pub/Sub, Pulsar.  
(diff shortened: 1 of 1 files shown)


fmvilas (changes requested):
I think there are many more places where you should add it. A quick search for the "bindings" word gives me a bunch of different results. IIRC, you're still missing operationTraitsBindings and componentsServerBindings, componentsChannelBindings, etc. Probably somewhere else I'm not remembering right now.

fmvilas (approved):
You're completely, right. My bad, sorry. Looks good as it is 👍

dalelane (commented):
Why "ros2" (as opposed to just "ros") ?
When there is a future updated version of the middleware, would this require a new "ros3" binding? 

(I'm unfamiliar with ROS, so apologies if the answer to this is obvious with context that I'm missing!)

char0n:
> > Why "ros2" (as opposed to just "ros") ? When there is a future updated version of the middleware, would this require a new "ros3" binding?
> > (I'm unfamiliar with ROS, so apologies if the answer to this is obvious with context that I'm missing!)
> 
> We think it definitely make sense to call it "ros2" since there are specific parameters related to ROS 2 that will not work with ROS. Evenmore, ROS is end of life so we believe that ROS 2 should be the name.

@dalelane it make sense. 

We only provide only single `bindingVersion` field on all bindings - which is THE version number of the binding specification for a particular protocol, NOT the version of the protocol itself. So if there are params in ROS2 that won't work in ROS, I don't see other options given the current state how bindings are currently modeled.

PR #1110, opened 2025-06-24
chore: remove sponsors that did not refresh sponsorship

derberg:
- Postman will not continue sponsoring AsyncAPI as platinum sponsor. This was expected after layoff of their AsyncAPI open source team last year.
- HiveMQ will not continue sponsoring AsyncAPI as silver sponsor. We are still in touch with them. They are our fans, so we will for sure reach out next year if they want to renew.
- HDI Global will not continue sponsoring AsyncAPI as silver sponsor. This was expected to be a one time donation, with [case study](https://www.asyncapi.com/casestudies/hdiglobal) and marketing that will show us End User Companies also donate. They are not going very far, they still promote AsyncAPI through their ambassador
- Solace moved from Platinum to Gold


fmvilas (approved):
LGTM 👍

PR #1135, opened 2025-11-15
feat: Add AND logic support for multiple security schemes (#1129)

### LINKED ISSUE #1129: Support AND logic for multiple security schemes in AsyncAPI 3.x
## Problem
Currently, AsyncAPI 3.x only allows expressing OR semantics between multiple security schemes using the security array:

```
security:
  - $ref: '#/components/securitySchemes/saslScram'
  - $ref: '#/components/securitySchemes/mtls'
```

This means either SASL/SCRAM or mTLS can authorize a connection.
However, in many real-world scenarios, a server or operation may require multiple authentication schemes simultaneously (logical AND), for example:
SASL/SCRAM and a valid client certificate (mTLS) for a Kafka broker

Currently, there is no spec-compliant way to enforce both requirements in AsyncAPI 3.x without using a vendor extension (x- fields) or describing it only in prose. This limitation prevents full machine-readable documentation and automated validation of multi-factor authentication requirements.

## Current workarounds
1. Use only one $ref and describe the second requirement in description
Pros: Spec-compliant
Cons: Not machine-enforceable, tooling cannot validate
2. Vendor extension
Pros: Documents AND requirement
Cons: Non-standard, tooling may ignore it
Neither approach is ideal for consistent, spec-compliant multi-scheme security modeling.

## Proposal
Introduce a native, spec-supported syntax to express AND logic between multiple security schemes in a single security requirement object. Possible approaches:
1. Array of $refs array.
```
security:
  - - $ref: '#/components/securitySchemes/saslScram'
    - $ref: '#/components/securitySchemes/mtls'
```
2. Named keys with logical AND semantics (similar to v2):
```
security:
  - saslScram: []
    mtls: []
```
Requirements for tooling: Studio, CLI validator, and generators should enforce that all schemes in a single requirement object are required.

GaneshPatil7517:
This PR implements support for AND logic between multiple security schemes in AsyncAPI 3.x, addressing issue #1129.

Problem
Currently, AsyncAPI 3.x only allows expressing OR semantics when multiple security schemes are specified in the security array. This means "any one of these schemes satisfies the requirement." However, there's no way to express that multiple schemes must be satisfied simultaneously (AND logic), which is a common real-world scenario.

For example, you cannot currently express: "Require BOTH OAuth2 AND API Key" or "Require user authentication AND client certificate."

Solution
This PR introduces support for nested array syntax to enable AND logic while maintaining backward compatibility:

Single-level array (existing): OR logic - any one scheme is sufficient
Nested arrays (new): Outer array = OR, Inner array = AND - all schemes in an inner array must be satisfied
Syntax Examples
OR logic (existing - unchanged):
security:
  - oauth2: [ ]
  - apiKey: [ ]
  
Meaning: OAuth2 AND API Key

AND logic (new):
security:
  - - oauth2: [ ]
    - apiKey: [ ]

Combined OR and AND:
security:
  - - oauth2: [ ]
    - apiKey: [ ]
  - basicAuth: [ ]

Meaning: (OAuth2 AND API Key) OR Basic Auth

Changes Made
✅ Updated Server Object security field documentation to support array of arrays syntax
✅ Updated Operation Object security field with AND/OR logic support
✅ Updated Operation Trait Object security field with AND/OR logic support
✅ Added comprehensive example: operation-security-and-logic-asyncapi.yml
✅ Added documentation in operation-security-asyncapi.yml explaining OR vs AND logic
✅ Maintains 100% backward compatibility with existing single-level arrays
Backward Compatibility
This change is fully backward compatible. Existing AsyncAPI documents using single-level arrays will continue to work exactly as before with OR semantics. The nested array syntax is purely additive.

Testing
Added two example files demonstrating the feature:

operation-security-and-logic-asyncapi.yml - Shows AND logic usage
Updated operation-security-asyncapi.yml - Documents OR vs AND comparison
Fixes
Fixes #1129
@rahmathm1


derberg:
@GaneshPatil7517 we are good to go forward here. In order to move to stage 2 you need to fulfill some requirements: https://github.com/asyncapi/spec/blob/master/CONTRIBUTING.md#stage-2-draft

Also please clearly state who's championing this proposal, you or @rahmathm1

PR #1186, opened 2026-03-01
refactor(converter): make file conversion functions async

muralimadhava96-ui:
refactor(converter): make file conversion functions async

Description
This PR refactors the convertExample and convertExampleDir functions to use fs/promises instead of synchronous file system operations.
Changes
->Replace synchronous fs methods with async equivalents from fs/promises
->Convert functions to use async/await
->Improve error handling for file read and parse failures
->Add validation to prevent processing non-YAML objects
Why
->Prevent blocking the Node.js event loop
->Improve scalability when handling multiple file conversions
->Improve code readability and maintainability
->Align with modern Node.js async best practices
Impact
This change does not modify external behavior or output.
It is strictly an internal refactor.

### DIFF
--- a/scripts/converter/index.js
+++ b/scripts/converter/index.js
@@ -1,57 +1,96 @@
 const path = require('path');
-const fs = require('fs');
+const fs = require('fs/promises');
 const { convert } = require('@asyncapi/converter');
 const jsYaml = require('js-yaml');
+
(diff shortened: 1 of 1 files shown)


derberg:
Thanks but this change is not needed

PR #1188, opened 2026-03-19
fix: clarify null value handling in traits merge mechanism

sahaiash:
## Summary

- Fixes #1178
- Adds explicit clarification that `null` values in traits or target objects MUST be preserved as-is
- `null` MUST NOT be interpreted as a deletion directive (unlike standard JSON Merge Patch semantics)
- Added a **Null Value Example** section under Traits Merge Mechanism with two clear examples

## Changes

- `spec/asyncapi.md` — updated the Traits Merge Mechanism section (line ~2707) with a clarifying sentence and null value examples

## Test plan

- [ ] Spec markdown lints cleanly
- [ ] Embedded examples validate via `npm run validate:examples`
- [ ] New null value examples are clear and consistent with the rule stated

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/spec/asyncapi.md
+++ b/spec/asyncapi.md
@@ -2704,7 +2704,7 @@ Runtime expressions preserve the type of the referenced value.
 
 ### <a name="traitsMergeMechanism"></a>Traits Merge Mechanism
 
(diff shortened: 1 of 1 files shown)


derberg:
@sahaiash thanks but atm I have to close the PR. We first need to discuss what the best approach is before we jump into PR

